PR #32, opened 2014-12-31
some whitespace changes

devoncarew:
- some whitespace nits
- rename `Fnction` to `ModelFunction` - it wasn't clear that the mispelling was deliberate (to avoid a namespace collision)
- move `unittest` to a dev_dependency

@keertip


keertip:
lgtm

PR #34, opened 2015-01-02
fix an npe

devoncarew:
Fix an NPE - it's possible for the referenced `element` to not resolve, and be null. @keertip


keertip:
lgtm

PR #35, opened 2015-01-02
Devoncarew remove space

devoncarew:
Fix an issue where method parameters that are declared without types are displayed with leading spaces. So, go from this:

    sanitizeStacktrace( st, bool shorten): String

to this:

    sanitizeStacktrace(st, bool shorten): String

Also, some small ws changes


keertip:
lgtm

PR #41, opened 2015-01-06
Coverage

sethladd:
The simple test was to ensure code coverage was working.


sethladd:
also @devoncarew if he's curious

test/model_test.dart
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) 2014, the Dart project authors.  Please see the AUTHORS file
```
devoncarew: `2015` :) I got caught by this myself today.

devoncarew:
Awesome! The coverage should go up quite a bit just with a few high level integration tests.

PR #44, opened 2015-01-06
access bootstrap files from web

keertip:
@sethladd 

Access the bootstrap files from the web, instead of copying it from the package. This gets us around to running dartdoc using pub run.

### DIFF
--- a/lib/src/css.dart
+++ b/lib/src/css.dart
@@ -12,6 +12,15 @@ class CSS {
   // The bootstrap css file
   final String cssFilePath = '/packages/bootstrap/css/bootstrap.css';
 
+  final String cssHeader =
+      'https://maxcdn.bootstrapcdn.com/bootstrap/3.3.1/css/bootstrap.min.css';
+
+  final String theme =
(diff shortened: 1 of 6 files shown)


lib/src/css.dart
```diff
@@ -12,6 +12,15 @@ class CSS {
   // The bootstrap css file
   final String cssFilePath = '/packages/bootstrap/css/bootstrap.css';
 
+  final String cssHeader =
+      'https://maxcdn.bootstrapcdn.com/bootstrap/3.3.1/css/bootstrap.min.css';
+
+  final String theme =
+      'https://maxcdn.bootstrapcdn.com/bootstrap/3.3.1/css/bootstrap-theme.min.css';
+
+  final String jsScript =
```
devoncarew: Keerti, fyi, I don't think any of the bits of bootstrap that dartdoc is using requires the JS file, just the main css file and the theme one.

keertip:
Thanks Devon, removed the JS script tag.

sethladd:
Works for now. Thanks!

Issue #42, opened 2015-01-06
dartdoc doesn't "just run"


sethladd:
To repro:

Clone dartdoc
git clone git@github.com:Daegalus/dart-uuid.git
cd dart-uuid
dart ../dartdoc/bin/dartdoc.dart

Error:

```
sethladd@sethladd-macbookair ~/Code/dart-uuid (master)$ dart ../dartdoc/bin/dartdoc.dart 
parsing /Users/sethladd/Code/dart-uuid/lib/aes.dart...
Unhandled exception:
AnalysisException: <rethrow>
Caused by Could not resolve dart:core
#0      AnalysisTask._safelyPerform (package:analyzer/src/generated/engine.dart:6762:7)
#1      AnalysisTask.perform (package:analyzer/src/generated/engine.dart:6727:21)
#2      AnalysisContextImpl._cacheDartResolutionData (package:analyzer/src/generated/engine.dart:2414:86)
#3      AnalysisContextImpl._getDartResolutionData (package:analyzer/src/generated/engine.dart:3046:41)
#4      AnalysisContextImpl._getDartResolutionData2 (package:analyzer/src/generated/engine.dart:3077:36)
#5      AnalysisContextImpl.computeLibraryElement (package:analyzer/src/generated/engine.dart:1192:81)
#6      DartDoc.parseLibraries.<anonymous closure> (package:dartdoc/dartdoc.dart:90:63)
#7      List.forEach (dart:core-patch/growable_array.dart:254)
#8      DartDoc.parseLibraries (package:dartdoc/dartdoc.dart:86:18)
#9      DartDoc.generateDocs (package:dartdoc/dartdoc.dart:42:31)
#10     main (file:///Users/sethladd/Code/dartdoc/bin/dartdoc.dart:26:62)
#11     _startIsolate (dart:isolate-patch/isolate_patch.dart:237)
#12     _startMainIsolate.<anonymous closure> (dart:isolate-patch/isolate_patch.dart:192)
#13     _RawReceivePortImpl._handleMessage (dart:isolate-patch/isolate_patch.dart:130)


#0      AnalysisContextImpl.recordResolveDartLibraryTaskResults (package:analyzer/src/generated/engine.dart:2115:11)
#1      AnalysisContextImpl_AnalysisTaskResultRecorder.visitResolveDartLibraryTask (package:analyzer/src/generated/engine.dart:5304:133)
[...]

PR #45, opened 2015-01-06
add option for sdk location

keertip:
@sethladd

### DIFF
--- a/bin/dartdoc.dart
+++ b/bin/dartdoc.dart
@@ -20,10 +20,10 @@ void main(List<String> arguments) {
   }
   List<String> excludeLibraries =
       results['exclude'] == null ? [] : results['exclude'].split(',');
-
+  String sdkLocation = results['dart-sdk'];
   String url = results['url'];
   var currentDir = Directory.current;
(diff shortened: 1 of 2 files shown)


keertip:
fix for #42

bin/dartdoc.dart
```diff
@@ -20,24 +20,30 @@ void main(List<String> arguments) {
...
   var currentDir = Directory.current;
-  new DartDoc(currentDir, excludeLibraries, url).generateDocs();
+  new DartDoc(currentDir, sdkLocation, excludeLibraries, url).generateDocs();
 }
 
 /// Print help if we are passed the help option or invalid arguments.
 void _printUsageAndExit(ArgParser parser) {
   print(parser.usage);
-  print('Usage: dartdoc [OPTIONS]');
+  print('Usage: dartdoc --dartsdk=sdkLocation [OPTIONS]');
```
sethladd: is there no way to auto-determine this? or, try a few strategies first before requiring this info from the developer?

keertip: Agreed, looking into it. Would prefer not to have the sdk location as a requirement. Added this so that dartdoc can be used for now.

PR #46, opened 2015-01-06
Devoncarew locate sdk

devoncarew:
Use grinder's function to locate the dart sdk. It tries to find it via the --dart-sdk cli arg, by a DART_SDK env var, relative to the dart executable, and via the unix `which` command.

@keertip


lib/dartdoc.dart
```diff
@@ -62,11 +65,19 @@ class DartDoc {
...
 
   List<LibraryElement> parseLibraries(List<String> files) {
-    DartSdk sdk = new DirectoryBasedDartSdk(new JavaFile(_getSdkDir().path));
+    Directory sdkDir = grinder.getSdkDir(_cliArgs);
+    if (sdkDir == null) {
+      print(
+          "Warning: unable to locate the Dart SDK. Please use the --dart-sdk "
+          "command line option or set the DART_SDK environment variable.");
+    }
+    DartSdk sdk = new DirectoryBasedDartSdk(new JavaFile(sdkDir.path));
```
sethladd: what is `new JavaFile` ?

bin/dartdoc.dart
```diff
@@ -23,7 +23,8 @@ void main(List<String> arguments) {
 
   String url = results['url'];
   var currentDir = Directory.current;
-  new DartDoc(currentDir, excludeLibraries, url).generateDocs();
+  DartDoc dartdoc = new DartDoc(currentDir, excludeLibraries, url, arguments);
+  dartdoc.generateDocs();
```
sethladd: method cascades?

lib/dartdoc.dart
devoncarew: It's a weird `analytics` engine-ism.

bin/dartdoc.dart
devoncarew: ok

lib/dartdoc.dart
sethladd: yuck. we should change that in the analysis engine.

sethladd:
lgtm

lib/dartdoc.dart
devoncarew: Set your geologic clock.

keertip:
lgtm! Thanks!

devoncarew:
Hmm, looks like I was working off a slightly older version of the source :( Have merged to master.

Issue #43, opened 2015-01-06
Unable to tell what version dartdoc is


sethladd:
```
sethladd@sethladd-macbookair ~/Code/dartdoc (coverage)$ dart bin/dartdoc.dart --version
Unhandled exception:
FormatException: Could not find an option named "version".
#0      Parser.validate (package:args/src/parser.dart:251:21)
#1      Parser.parseLongOption (package:args/src/parser.dart:240:15)
#2      Parser.parse (package:args/src/parser.dart:77:26)
#3      ArgParser.parse (package:args/src/arg_parser.dart:113:62)
#4      main (file:///Users/sethladd/Code/dartdoc/bin/dartdoc.dart:16:29)
#5      _startIsolate (dart:isolate-patch/isolate_patch.dart:237)
#6      _startMainIsolate.<anonymous closure> (dart:isolate-patch/isolate_patch.dart:192)
#7      _RawReceivePortImpl._handleMessage (dart:isolate-patch/isolate_patch.dart:130)
```

ricowind:
I guess this is a dub of:
https://code.google.com/p/dart/issues/detail?id=16718

sethladd:
Thanks @ricowind, same type of issue.

PR #47, opened 2015-01-06
add version info 

keertip:
fix for #43 

@sethladd

### DIFF
--- a/lib/dartdoc.dart
+++ b/lib/dartdoc.dart
@@ -67,6 +67,7 @@ class DartDoc {
   }
 
   List<LibraryElement> parseLibraries(List<String> files) {
+    print(_getSdkDir().path);
     DartSdk sdk = new DirectoryBasedDartSdk(new JavaFile(_getSdkDir().path));
 
     ContentCache contentCache = new ContentCache();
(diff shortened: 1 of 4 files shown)


lib/dartdoc.dart
```diff
@@ -22,6 +22,11 @@ import 'src/model_utils.dart';
 
 const String DEFAULT_OUTPUT_DIRECTORY = 'docs';
 
+const String NAME = 'dartdoc';
+
+// Update when puspec version changes
+const String VERSION = '0.0.1';
```
sethladd: should we have a grinder task to bump this and our pubspec.yaml ?

Also, how about a test that opens pubspec.yaml and compares to this?

keertip: Have a test that compares this with pubspec

test/dartdoc_test.dart
```diff
@@ -0,0 +1,24 @@
...
+
+import '../lib/dartdoc.dart';
+import '../lib/src/package_utils.dart';
+
+
+tests() {
+  group('dartdoc test', () {
+
+    test('version info', () {
+      String version = getPackageVersion(Directory.current.parent.path);
```
devoncarew: The tests generally run from the project root - I think you want this to be `Directory.current.path`.

Issue #50, opened 2015-01-07
null object does not have a getter 'name'


sethladd:
To repro:

git clone git@github.com:google/quiver-dart.git
cd quiver-dart
dart ../dartdoc/bin/dartdoc.dart

output:

```
sethladd@sethladd-macbookair ~/Code/quiver-dart (master)$ dart ../dartdoc/bin/dartdoc.dart 
parsing /Users/sethladd/Code/quiver-dart/lib/async.dart...
parsing /Users/sethladd/Code/quiver-dart/lib/cache.dart...
parsing /Users/sethladd/Code/quiver-dart/lib/check.dart...
parsing /Users/sethladd/Code/quiver-dart/lib/collection.dart...
parsing /Users/sethladd/Code/quiver-dart/lib/core.dart...
parsing /Users/sethladd/Code/quiver-dart/lib/io.dart...
parsing /Users/sethladd/Code/quiver-dart/lib/iterables.dart...
parsing /Users/sethladd/Code/quiver-dart/lib/mirrors.dart...
parsing /Users/sethladd/Code/quiver-dart/lib/pattern.dart...
parsing /Users/sethladd/Code/quiver-dart/lib/streams.dart...
parsing /Users/sethladd/Code/quiver-dart/lib/strings.dart...
parsing /Users/sethladd/Code/quiver-dart/lib/time.dart...
parsing /Users/sethladd/Code/quiver-dart/lib/testing/async.dart...
parsing /Users/sethladd/Code/quiver-dart/lib/testing/runtime.dart...
parsing /Users/sethladd/Code/quiver-dart/lib/testing/time.dart...
generating docs/quiver_package.html
generating docs/quiver.async.html
Unhandled exception:
The null object does not have a getter 'name'.

NoSuchMethodError: method not found: 'name'
Receiver: null
Arguments: []
#0      Object.noSuchMethod (dart:core-patch/object_patch.dart:45)
#1      HtmlGeneratorHelper.createLinkedTypeName (package:dartdoc/src/generator.dart:536:30)
#2      HtmlGeneratorHelper.createLinkedTypeName (package:dartdoc/src/generator.dart:549:41)
#3      HtmlGeneratorHelper.createLinkedReturnTypeName (package:dartdoc/src/generator.dart:565:34)
#4      Typedef.createLinkedSummary (package:dartdoc/src/model.dart:405:51)
#5      HtmlGenerator.createToc (package:dartdoc/src/generator.dart:198:65)
[...]

PR #53, opened 2015-01-07
add type parameter

keertip:
fix for #50 

@sethladd


lib/src/model.dart
```diff
@@ -43,6 +43,9 @@ abstract class ModelElement {
     if (e is TopLevelVariableElement) {
       return new Variable(e, library);
     }
+    if (e is TypeParameterElement) {
```
sethladd: unit test?

sethladd:
lgtm thanks!

Issue #57, opened 2015-01-07
index page, blue button link goes to 404


sethladd:
![package_quiver](https://cloud.githubusercontent.com/assets/5479/5655429/f094e432-9682-11e4-9896-ed3e6deda43d.png)

sethladd:
The blue button goes to file:///Users/sethladd/Code/quiver-dart/docs/quiver  ... maybe it should go to / or /index.html ?

sethladd:
Fixed.

PR #60, opened 2015-01-08
generate the index file from a template

sethladd:

PR #61, opened 2015-01-08
fix link on main page

keertip:
fix for #57 

@sethladd

### DIFF
--- a/lib/src/generator.dart
+++ b/lib/src/generator.dart
@@ -70,7 +70,7 @@ body {
     html.startTag('ul', attributes: 'class="nav nav-pills nav-stacked"');
     html.startTag('li', attributes: 'class="active"', newLine: false);
     html.write(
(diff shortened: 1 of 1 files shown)


lib/src/generator.dart
```diff
@@ -70,7 +70,7 @@ body {
     html.startTag('ul', attributes: 'class="nav nav-pills nav-stacked"');
     html.startTag('li', attributes: 'class="active"', newLine: false);
     html.write(
-        '<a href="${packageName}">' '<i class="chevron-nav icon-white icon-chevron-right"></i> ' '${packageName}-${packageVersion}</a>');
+        '<a href="inde.html">' '<i class="chevron-nav icon-white icon-chevron-right"></i> ' '${packageName}-${packageVersion}</a>');
```
sethladd: do you mean index.html ?

keertip: Yes. But this fix is no longer needed, once the template cl  #60 lands.

sethladd:
Yeah, let's close this. I'll fix in #60

PR #63, opened 2015-01-08
run dartdoc as a test

keertip:
@devoncarew


tool/grind.dart
```diff
@@ -0,0 +1,49 @@
...
+
+import 'dart:io';
+
+import 'package:dartdoc/dartdoc.dart';
+import 'package:grinder/grinder.dart';
+
+final Directory DOC_DIR = new Directory(DEFAULT_OUTPUT_DIRECTORY);
+
+void main([List<String> args]) {
+  task('init', init);
```
devoncarew: You can use grinder's `defaultInit` here:

    task('init', defaultInit);

tool/travis.sh
```diff
@@ -40,4 +40,11 @@ if [ "$REPO_TOKEN" ]; then
...
-fi
\ No newline at end of file
+fi
+
+# Globally install grinder.
+pub global activate grinder
+export PATH=~/.pub-cache/bin:$PATH
+
+# Run dartdoc.
+grind test
```
devoncarew: I generally use `pub global run grind` here, instead of just `grind`, as it's a bit more robust.

tool/grind.dart
```diff
@@ -0,0 +1,49 @@
...
+      context.log('running dartdoc');
+      runDartScript(context, '../bin/dartdoc.dart');
+
+      File indexHtml = joinFile(DOC_DIR, ['index.html']);
+      if(!indexHtml.existsSync()) context.fail('docs not generated');
+      File docFile = joinFile(DOC_DIR, ['dartdoc.html']);
+      if(!docFile.existsSync()) context.fail('docs not generated');
+
+  } catch (e) {
+    try { DOC_DIR.deleteSync(recursive: true); }
```
devoncarew: I wouldn't be super worried about cleaning up the `docs` dir - just letting an exception that happens pass through to the build script - it'll fail the build (appropriately).

tool/grind.dart
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Run dartdoc and check that the docs are generated.
+ */
+void testDartdoc(GrinderContext context) {
+  if (DOC_DIR.existsSync()) DOC_DIR.deleteSync(recursive: true);
+
+  try {
+      context.log('running dartdoc');
+      runDartScript(context, '../bin/dartdoc.dart');
```
devoncarew: The cwd will be the project root, so `'bin/dartdoc.dart'` should work here.

devoncarew:
Very cool! One fix to the file path, then lgtm.

tool/travis.sh
```diff
@@ -28,6 +32,9 @@ dartanalyzer --fatal-warnings \
   bin/dartdoc.dart \
   lib/dartdoc.dart \
   test/all.dart
+  
+# Run dartdoc.
+pub global run grind test
```
devoncarew: here, `grinder` instead of grind

PR #110, opened 2015-01-15
generate docs for dart sdk

keertip:
@sethladd

### DIFF
--- a/lib/dartdoc.dart
+++ b/lib/dartdoc.dart
@@ -77,7 +77,6 @@ class DartDoc {
 
   List<LibraryElement> _parseLibraries(List<String> files) {
     DartSdk sdk = new DirectoryBasedDartSdk(new JavaFile(_sdkDir.path));
-
     ContentCache contentCache = new ContentCache();
     List<UriResolver> resolvers = [new DartUriResolver(sdk), new FileUriResolver()];
     JavaFile packagesDir = new JavaFile.relative(new JavaFile(_rootDir.path), 'packages');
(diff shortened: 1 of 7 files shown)


lib/dartdoc.dart
```diff
@@ -73,14 +82,13 @@ class DartDoc {
...
-
+  List<LibraryElement> _parseLibraries(List<String> files) {
     DartSdk sdk = new DirectoryBasedDartSdk(new JavaFile(_sdkDir.path));
-
     ContentCache contentCache = new ContentCache();
-    List<UriResolver> resolvers = [new DartUriResolver(sdk), new FileUriResolver()];
-    JavaFile packagesDir = new JavaFile.relative(new JavaFile(_rootDir.path), 'packages');
+    List<UriResolver> resolvers = [new DartUriResolver(sdk),
+                                   new FileUriResolver()];
+    JavaFile packagesDir =
```
sethladd: This always confuses me. When can we rename JavaFile? :)

test/model_test.dart
```diff
@@ -67,6 +69,20 @@ tests() {
...
     });
+
+    var p2 = new Package([e], Directory.current.path, '1.9.0-dev.3.0', true);
+
+    test('sdk name', () {
+      expect(p2.name, 'Dart API Reference');
+    });
+
+    test('sdk version', () {
+      expect(p2.version, '1.9.0-dev.3.0');
```
sethladd: will this change over time? will that make it hard to keep updated?

sethladd:
Woohoo!

lib/dartdoc.dart
keertip: Will talk to the analyzer folks.

test/model_test.dart
keertip: No, this is just a test and does not need to be updated.

lib/dartdoc.dart
sethladd: https://code.google.com/p/dart/issues/detail?id=22075 :)

PR #113, opened 2015-01-15
fix links in index.html for sdk docs

keertip:
@sethladd

### DIFF
--- a/lib/src/html_generator.dart
+++ b/lib/src/html_generator.dart
@@ -448,7 +448,7 @@ body {
 }
 
 String _getFileNameFor(Library library) {
-  return '${library.name.replaceAll('.', '_')}.html';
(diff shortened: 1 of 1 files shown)


lib/src/html_generator.dart
```diff
@@ -448,7 +448,7 @@ body {
 }
 
 String _getFileNameFor(Library library) {
-  return '${library.name.replaceAll('.', '_')}.html';
+  return '${library.name.replaceAll('.', '_').replaceAll(':', '_')}.html';
```
sethladd: can we add a comment explaining why we are doing this? :)

keertip: done.

sethladd:
Is it easy to write a test for this?

keertip:
Not right now, since it is a private method. Will refactor and add tests in a separate cl.

Issue #66, opened 2015-01-09
Display version of package or SDK somewhere on page, and a "generated at" date


sethladd:
If it's a package, display its version. If it's the SDK, display its version. Makes it easy to know how to match docs to a version of the source.

Probably also display a "generated at" date.

keertip:
#134

PR #134, opened 2015-01-27
add sdk info and date to index.html

keertip:
Address #66 

@sethladd


sethladd:
Thanks! How does this work with a package version?

keertip:
Package version can be got with package.version. This is version info for the sdk used to analyze the package.

PR #148, opened 2015-02-03
fix tests

keertip:
@sethladd

### DIFF
--- a/lib/src/model.dart
+++ b/lib/src/model.dart
@@ -491,7 +491,7 @@ class Class extends ModelElement {
     return c.map((e) {
       var cSource =
           (source != null) ? source.substring(e.node.offset, e.node.end) : null;
-      return new Constructor(e, library, cSource);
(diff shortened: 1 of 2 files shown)


sethladd:
thanks!

sethladd:
fwiw I'm not sure we need to pass Package to all of these. We already pass Library, which has a getter to Package. I wouldn't be surprised if you refactor this away :)

keertip:
Yes, was going to make the package part of the library during the refactor.

On Mon, Feb 2, 2015 at 4:47 PM, Seth Ladd <notifications@github.com> wrote:

> fwiw I'm not sure we need to pass Package to all of these. We already pass
> Library, which has a getter to Package. I wouldn't be surprised if you
> refactor this away :)
>
> —
> <https://github.com/dart-lang/dartdoc/pull/148#issuecomment-72570585>.
>



-- 
Keerti

PR #212, opened 2015-02-17
test against dev

keertip:
@sethladd


sethladd:
can we test against dev and stable?

keertip:
Done!

sethladd:
Matrix builds... cool!

PR #222, opened 2015-02-20
Dogfood the new unittest

kevmoo:
With a few other tweaks

### DIFF
--- a/tool/travis.sh
+++ b/tool/travis.sh
@@ -23,7 +23,7 @@ dart ../../bin/dartdoc.dart
 cd ../..
 
 # Run the tests.
-dart test/all.dart
+pub run unittest:unittest
 
 # Gather and send coverage data.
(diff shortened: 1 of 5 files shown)


tool/travis.sh
```diff
@@ -23,7 +23,7 @@ dart ../../bin/dartdoc.dart
 cd ../..
 
 # Run the tests.
-dart test/all.dart
+pub run unittest:unittest
```
sethladd: Do we have to run this via pub run? Is there an alternative way to run it?

kevmoo: `pub run` is the way. You want to run the binary. There is no good way to get at a package binary – except `pub run`

kevmoo:
You don't want to use this. It was mostly an experiment.

the new unittest needs Dart 1.9 – so until you want to bump your min SDK, I'll call this closed.

PR #223, opened 2015-02-20
getting the complete annotation

keertip:
first pass at getting the entire annotation

### DIFF
--- a/lib/src/model.dart
+++ b/lib/src/model.dart
@@ -179,7 +179,7 @@ abstract class ModelElement {
      if (annotations.isNotEmpty) {
        return annotations.map((f) {
          var s =
-         f.toSource();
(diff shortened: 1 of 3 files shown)


lib/src/model.dart
```diff
@@ -172,17 +172,17 @@ abstract class ModelElement {
...
-          return f.element.name;
-        }
-      }).toList(growable:false);
+    var node = element.node;
+    if (node is AnnotatedNode) {
+      List<Annotation> annotations = node.metadata;
+     if (annotations.isNotEmpty) {
+       return annotations.map((f) {
+         var s =
+         f.toSource().substring(1);
```
sethladd: ah ha!

sethladd:
So good, thanks!

PR #225, opened 2015-02-21
resolve references in annotations

keertip:


sethladd:
Ah, nice! Thanks @keertip

PR #226, opened 2015-02-21
Cleanup template_test

kevmoo:
This allows each test to be run independently, which is not the case

...will enable use in future unittest


sethladd:
lgtm thanks!

Issue #237, opened 2015-02-23
For methods or functions in the core SDK marked as async, state that they return a Future


sethladd:
After talking to Gilad, the decision was made to not explicitly document that a function/method was marked as async. Instead, document that it returns a Future.

Let's test that we do the right thing for the SDK and package docs.

sethladd:
Hi @keertip when you have a moment, can you look at what it will take to ask the Analyzer if a method was marked as async and then have our ModelFunction and Method objects state they return a Future?

Thanks!!

sethladd:
Is this possible with the analyzer API?

We'll have to do something similar with async*

keertip:
Yes, we can do this.

sethladd:
![screen shot 2015-03-01 at 1 08 46 pm](https://cloud.githubusercontent.com/assets/5479/6432815/1f5fd474-c014-11e4-8930-8c6e4aa153b9.png)

keertip:
Looks like the sdk libraries have to move to using async. Test when the change has been made.

sethladd:
We do this.

![screen shot 2015-07-17 at 2 58 55 pm](https://cloud.githubusercontent.com/assets/5479/8758033/5f7e7bf4-2c94-11e5-8307-110c43454140.png)

PR #239, opened 2015-02-27
return future for async methods

keertip:
fix for #237

### DIFF
--- a/test/model_test.dart
+++ b/test/model_test.dart
@@ -152,7 +152,7 @@ void main() {
     });
 
     test('get methods', () {
-      expect(Dog.instanceMethods, hasLength(1));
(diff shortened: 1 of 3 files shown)


test/model_test.dart
```diff
@@ -359,6 +356,10 @@ void main() {
     test('linkedName', () {
       expect(p1.type.linkedName, 'int');
     });
+
+    test('async return type', (){
+      expect(asyncM.linkedReturnType, 'Future');
```
sethladd: will this work for the SDK? We'll want Future linked then, yes?

keertip: Yes, we would want it linked then. Will add a note.

PR #257, opened 2015-03-09
look for docs in right place

keertip:


sethladd:
Hm, we should probably get rid of `new`. That was there only to coexist with the previous HTML generator.

Good catch, thought! :)

keertip:
Yes, thought so too. Will do it in a another cl.

On Mon, Mar 9, 2015 at 9:58 AM, Seth Ladd <notifications@github.com> wrote:

> Hm, we should probably get rid of new. That was there only to coexist
> with the previous HTML generator.
>
> Good catch, thought! :)
>
> —
> <https://github.com/dart-lang/dartdoc/pull/257#issuecomment-77894977>.
>



-- 
Keerti

Issue #262, opened 2015-03-13
page for library has empty sections


keertip:
The library page has sections that are empty

![screen shot 2015-03-12 at 10 37 30 pm](https://cloud.githubusercontent.com/assets/2192312/6633789/35b1c5f2-c90b-11e4-9270-486dade46f8a.png)

PR #263, opened 2015-03-13
remove empty sections from library page

keertip:
#262

### DIFF
--- a/lib/src/model.dart
+++ b/lib/src/model.dart
@@ -410,6 +410,7 @@ class Library extends ModelElement {
   Package package;
   List<Class> _classes;
   List<Class> _enums;
+  List<ModelFunction> _functions;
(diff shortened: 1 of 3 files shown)


lib/src/model.dart
```diff
@@ -442,49 +445,68 @@ class Library extends ModelElement {
...
     return _getVariables().where((v) => !v.isConst).toList(growable: false);
   }
 
+  bool get hasConstants => _getVariables().any((v) => v.isConst);
+
   List<Variable> getConstants() {
     return _getVariables().where((v) => v.isConst).toList(growable: false);
   }
 
+  bool get hasEnums => !getEnums().isEmpty;
```
sethladd: fwiw I like `getEnums().isNotEmpty` better :)

PR #273, opened 2015-03-15
fixing up handling of generics

sethladd:


sethladd:
Probably doesn't cover all cases, but it's better.

PR #291, opened 2015-03-23
Oneliner: Skip Over Code Blocks

kaendfinger:
This PR makes sure that the oneliner function skips over code-blocks. This is to ensure that if a code block is above the initial documentation comment (which it should pretty much never be), that the documentation will still look clean by not displaying code blocks are HTML Escaped (thus having &lt; etc sequences in the oneliner).


sethladd:
Can you please file a bug that details the problem, and then link it to this PR? Thanks!

kaendfinger:
See #295

PR #288, opened 2015-03-22
Bug Fixes

kaendfinger:
See the commits for more details, but this fixes quite a few bugs.

PR #292, opened 2015-03-23
Implement SDK Intro

kaendfinger:
This implements a fully working SDK intro.

This requires that PR #288 is merged as well to work fully correctly.

### DIFF
--- a/lib/src/package_utils.dart
+++ b/lib/src/package_utils.dart
@@ -33,3 +33,50 @@ String getPackageDescription(String directoryName) {
 
 String getPackageVersion(String directoryName) =>
     _getPubspec(directoryName)['version'];
+
+const String SDK_INTRO = """
+Welcome to the Dart API reference documentation,
+covering the official Dart API libraries.
+Some of the most fundamental Dart libraries include:
+
+* [dart:core]:
+  Core functionality such as strings, numbers, collections, errors,
+  dates, and URIs.
+* [dart:html]:
+  DOM manipulation for web apps.
+* [dart:io]:
+  I/O for command-line apps.
+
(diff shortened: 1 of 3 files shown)


lib/src/package_utils.dart
```diff
@@ -33,3 +33,50 @@ String getPackageDescription(String directoryName) {
 
 String getPackageVersion(String directoryName) =>
     _getPubspec(directoryName)['version'];
+
+const String SDK_INTRO = """
```
sethladd: Can we please pull this in via a command-line arg? It doesn't feel right to have SDK-specific verbage in a generic dartdoc tool.

sethladd:
Not ready for merge.

lib/src/package_utils.dart
kaendfinger: As in, add an argument like --intro=PATH_TO_FILE?
What really should happen is that the SDK intro file should be included in the SDK itself. That would help quite a bit.

keertip: Yes, that is how it should be. :+1:

kaendfinger: Making a Code Review to copy the SDK Intro into the SDK

kaendfinger: @keertip @sethladd https://codereview.chromium.org/1015013006/
Once this is merged, I'll do a check to see if the SDK contains one, and if so, we will use it, otherwise, we wont.

sethladd:
Going to close this, as we don't want to embed SDK specific text into this tool. Thanks, though!

kaendfinger:
No Problem

PR #293, opened 2015-03-23
Pre-Compile Templates for Better Performance

kaendfinger:
Pre-Compiling Templates gives a decent speed boost.

### DIFF
--- a/lib/src/html_generator.dart
+++ b/lib/src/html_generator.dart
@@ -206,7 +206,7 @@ class HtmlGenerator extends Generator {
     return f;
   }
 
-  void _writeFile(String filename, template, Map data) {
+  void _writeFile(String filename, MustacheTemplate template, Map data) {
     File f = _createOutputFile(filename);
     String content = template(data, assumeNullNonExistingProperty: false, errorOnMissingProperty: true);
     f.writeAsStringSync(content);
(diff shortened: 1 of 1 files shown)


kaendfinger:
@keertip

lib/src/html_generator.dart
```diff
@@ -30,21 +30,21 @@ class HtmlGenerator extends Generator {
   final String generatedOn;
 
   // TODO: fix this to be a single map, just like _partialTemplates
-  static final String indexTemplate = _loadTemplate('templates/index.html');
-  static final String libraryTemplate =
+  static final Function indexTemplate = _loadTemplate('templates/index.html');
```
sethladd: Can we use a typedef here? Function is so... non-descriptive :)

lib/src/html_generator.dart
```diff
@@ -204,21 +204,18 @@ class HtmlGenerator extends Generator {
     return f;
   }
 
-  void _writeFile(String filename, String template, Map data) {
+  void _writeFile(String filename, template, Map data) {
```
sethladd: Use type annotation for template

kaendfinger: Unable to. We pass both strings and functions to it.

kaendfinger:
@sethladd this saves about 4 seconds when generating the SDK docs on my MacBook Pro.

kaendfinger:
@sethladd ping

sethladd:
Thanks for working on the performance fix. I'm very uneasy about a function that returns dynamic. Can you do a deeper refactoring so that each function returns exactly one type ?

kaendfinger:
@sethladd Sure

kaendfinger:
I'm going to actually redo this PR so you don't have to fix the conflicts.

PR #294, opened 2015-03-23
Multiple Bug Fixes

kaendfinger:
- Fix Formatting of Optional Parameters
- Include Interfaces in Inherited Methods and Properties

### DIFF
--- a/templates/class.html
+++ b/templates/class.html
@@ -224,7 +224,7 @@
     </section>
     {{/class.hasStaticMethods}}
 
-    {{#class.hasInstanceMethods}}
+    {{#class.hasMethods}}
     <section id="instance-methods-summary" class="summary">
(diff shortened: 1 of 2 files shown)


templates/class.html
```diff
@@ -224,7 +224,7 @@ <h2 class="text-display-1">Static Methods</h2>
     </section>
     {{/class.hasStaticMethods}}
 
-    {{#class.hasInstanceMethods}}
+    {{#class.hasMethods}}
```
sethladd: Hm,  I think this is correct. Good catch!

sethladd:
Lovely. Thanks!

sethladd:
Can you file bugs for these? I'd like to keep solid records of what we fixed with our PRs.

PR #296, opened 2015-03-23
Lots of Tweaks

kaendfinger:
- Don't show hidden interfaces in implements section.
- Don't show hidden superclasses in extends section.
- Remove hrefs for unresolved annotations. (It was taking you to a page called 'null')
- Fix urls for references to properties inside method documentation.
- Add a link on method/function pages which will take you to the class/library that defines it.


kaendfinger:
cc @sethladd @keertip

keertip:
Could you create/reference issues, and add code to fake_package so that all the fixes are captured.

kaendfinger:
@keertip I'll try my best :)

kaendfinger:
@keertip Added issues. What kinds of tests do I need? AFAIK, we don't have generation tests, which is what 99% of this relates to.

keertip:
@kaendfinger, thanks for adding code to the fake package!

keertip:
lgtm!

kaendfinger:
Thanks!

PR #306, opened 2015-03-24
Pre-Compile Templates

kaendfinger:
This pre-compiles templates to make generation faster. It seems to make it a few seconds quicker.

cc @sethladd


sethladd:
Nice! lgtm (thank you for removing the dynamic annotations :)

sethladd:
cc @keertip this might help generation speed

kaendfinger:
Of Course :)

PR #309, opened 2015-03-24
Add inline code syntax.

### LINKED ISSUE #303: Legacy [:code:] references no longer recognized
The documentation for `bool.toString()` references e.g. [:true:] and [:"true":].  It is supported by the old dartdoc:

https://api.dartlang.org/apidocs/channels/stable/dartdoc-viewer/dart:core.bool#id_toString

But in the new dartdoc

https://ci.directcode.org/guestAuth/repository/download/Dart_SdkDocs/.lastSuccessful/dart_core/bool/toString.html

the escape sequences are not recognized:

Returns [:"true":] if the receiver is [:true:], or [:"false":] if the receiver is [:false:].

Probably those docs should just be updated to reference e.g. `` `true` `` and `` `"true"` ``, but I wanted to confirm that here first before filing a bug on dartbug.com.

kaendfinger:
This fixes #303.

cc @sethladd

### DIFF
--- a/lib/src/html_generator.dart
+++ b/lib/src/html_generator.dart
@@ -15,6 +15,7 @@ import 'package:markdown/markdown.dart' as md;
 import 'package:path/path.dart' as path;
 
 import 'model.dart';
+import 'html_utils.dart';
 import '../generator.dart';
 
 class HtmlGenerator extends Generator {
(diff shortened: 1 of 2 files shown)


sethladd:
thanks!

PR #310, opened 2015-03-24
Implement Property Pages

kaendfinger:
This implements properties pages.

- Generates pages for static properties on classes
- Generates pages for instance properties on classes
- Generates pages for constant properties on classes.
- Generates pages for constant properties on libraries.
- Generates pages for top level properties on libraries.

This also links everything on the index pages.

cc @sethladd


kaendfinger:
@sethladd This should work pretty well.

sethladd:
You rock. I haven't tried it, but we can dive in after the merge.

PR #312, opened 2015-04-02
add a command line flag to specify description file for package

keertip:
@sethladd


bin/dartdoc.dart
```diff
@@ -68,5 +69,7 @@ ArgParser _createArgsParser() {
   parser.addFlag('sdk-docs', help: 'generate docs for the Dart SDK.'
       ' '
       'Use "--dart-sdk" option to specify path to sdk');
+  parser.addOption('readme',
+      help: 'path to the package description file, including file name');
```
sethladd: "package or SDK" ?

I think "including file name" is probably redundant.

lib/src/package_utils.dart
```diff
@@ -20,7 +20,11 @@ Map _getPubspec(String directoryName) {
   return loadYaml(contents);
 }
 
-String getPackageDescription(String directoryName) {
+String getPackageDescription(bool isSdk, String readmeLoc, String directoryName) {
+  if (isSdk && readmeLoc !=  null) {
+    var readmeFile = new File(readmeLoc);
+    if (readmeFile.existsSync()) return readmeFile.readAsStringSync();
```
sethladd: if it doesn't, what happens? when do we tell the user the filename that they provided is invalid?

keertip: Missed that. Will add check.

bin/dartdoc.dart
keertip: changed to SDK

test/fake_package/test_readme.md
```diff
@@ -0,0 +1 @@
+Welcome to the Dart API reference documentation.
```
sethladd: Maybe include a bit of markdown syntax here so we can verify it's getting converted correctly?

sethladd:
lgtm with one small suggestion

test/fake_package/test_readme.md
keertip: done

PR #321, opened 2015-04-03
remove duplicated func

guillermooo:
This one doesn't seem to be used, whereas 

https://github.com/dart-lang/dartdoc/blob/master/lib/src/io_utils.dart#L112

is used somewhere else.

It may make more sense to keep this one because of the more exact name, though.


keertip:
Thanks!

PR #328, opened 2015-04-07
add an option to insert html in footer

keertip:


lib/dartdoc.dart
```diff
@@ -29,8 +29,8 @@ const String NAME = 'dartdoc';
 const String VERSION = '0.0.1';
 
 /// Initialize and setup the generators
-List<Generator> initGenerators(String url) {
-  return [new HtmlGenerator(url)];
+List<Generator> initGenerators(String url, String footer) {
```
sethladd: should footer be a named optional param?

templates/_footer.html
```diff
@@ -23,6 +23,8 @@
         prettyPrint();
     });
 </script>
+
+<!-- Footer Placeholder -->
```
sethladd: maybe a note that says "do not remove"

lib/dartdoc.dart
keertip: done.

templates/_footer.html
keertip: done.

Issue #324, opened 2015-04-03
Put type on right-hand side for properties, constants, typedefs, etc


sethladd:
cc @keertip @lukechurch

sethladd:
![screen shot 2015-04-03 at 10 44 53 am](https://cloud.githubusercontent.com/assets/5479/6986055/75b56912-d9ee-11e4-869a-1b01170bc84e.png)


Are we sure we want this? Doesn't the arrow imply "returns" ? Does a property "return" a type? Or is a property _a_ type of....

sethladd:
Dart Editor uses arrows for methods/functions, but not properties.

![dart_editor2](https://cloud.githubusercontent.com/assets/5479/6986242/20180e68-d9f0-11e4-8dbc-8a0af92e1e6f.png)

![dart_editor](https://cloud.githubusercontent.com/assets/5479/6986244/24cb17ac-d9f0-11e4-8ea6-5bc72d5aa317.png)

keertip:
@lukechurch 

Do we need the arrow?  Reads fine without it.

sethladd:
Removed the arrow.

![screen shot 2015-04-07 at 1 05 18 pm](https://cloud.githubusercontent.com/assets/5479/7032313/b7de0d2c-dd26-11e4-9918-674e56013f3b.png)

PR #330, opened 2015-04-07
Right hand side types

sethladd:


sethladd:
Covers issue https://github.com/dart-lang/dartdoc/issues/324

PR #338, opened 2015-04-08
use travis containers

kevmoo:
Faster to boot up

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: dart
+sudo: false
 dart: 
   - dev
   - stable


kevmoo:
FYI @devoncarew – this can help travis turn around a lot faster. They use docker containers so they can avoid booting an entire VM for each run.

devoncarew:
Cool; I remember turning this on for a project, but ended up turning it off (can't recall the reason why). I'll give this another shot; I have been bit by long latency waiting for builds to start up.

devoncarew:
lgtm; I changed the two dartpad projects over

sethladd:
Thanks!

PR #341, opened 2015-04-09
don't explicitly show exported libraries

keertip:


lib/src/model.dart
```diff
@@ -579,6 +622,8 @@ class Library extends ModelElement {
         growable: false);
   }
 
+  List<Class> getAllClasses() => _allClasses;
```
sethladd: why not getter?

keertip: done

PR #347, opened 2015-04-11
Typdefs and bug fixes

sethladd:


kevmoo:
Check you formatting.

https://travis-ci.org/dart-lang/dartdoc/jobs/58030533#L142

:smile:

sethladd:
Hah, that's awesome. You are sneaky.

Is there any automated way to run the formatter, and just commit whatever
formatting it made?

kevmoo:
ideally you'd make sure each commit is formatted – instead of adding another 'format' commit after your work

You could/should wire up each git repo for which you care with this: https://plus.sandbox.google.com/+KevinMoore314/posts/RMeFKTzEQoB

Then you can't commit unless it's formatted

sethladd:
Is there a workflow that's better than "always remember to run format before I send up a PR?"  That is, can I automatically include it in a workflow/set of scripts?

kevmoo:
Not sure how to hack that w/ git tools.

We need this in intelliJ - format on save.

keertip:
We now have format on save in Dart Editor and Eclipse plugins.

PR #351, opened 2015-04-12
Moar annotations

sethladd:


sethladd:
I ran the formatter this time :)

PR #363, opened 2015-04-13
fix gen of docs for sdk

keertip:


sethladd:
Should we generate the SDK docs in the CI system so we catch regressions like this?

lib/src/model.dart
```diff
@@ -1258,7 +1258,17 @@ class Parameter extends ModelElement {
 
   @override
   String get _href {
-    throw new UnsupportedError('parameters do not have hrefs');
+    var p = _parameter.enclosingElement;
```
sethladd: Why does a parameter need _href ? Isn't _href only used when we're generating a page for it?

sethladd:
We also need to add an HTML ID to all parameters on dedicated page.

lib/src/model.dart
keertip: Also used to resolve references in comments - parameters are linked in comments for methods/functions.

PR #371, opened 2015-04-13
fix linking for constants

keertip:


lib/src/model.dart
```diff
@@ -1028,7 +1028,8 @@ class Field extends ModelElement {
     if (_field.node == null) return null;
     var v = _field.node.toSource();
     if (v == null) return null;
-    return v.substring(v.indexOf('=') + 1, v.length).trim();
+    var string = v.substring(v.indexOf('= ') + 2, v.length);
```
sethladd: Not all constants have a space between the `=`. I think we need to roll this line back.

test/fake_package/lib/example.dart
```diff
@@ -11,6 +11,7 @@ get y => 2;
 
 const String COLOR = 'red';
 const String COMPLEX_COLOR = 'red' + '-' + 'green' + '-' + 'blue';
+const ConstantCat MY_CAT = const ConstantCat('tabby');
```
sethladd: Please also add a case for `const String COLOR='green'` and other cases of random spacing.

sethladd:
Disclaimer: just assumptions. testing different scenarios now.

lib/src/model.dart
keertip: Will do, and also add a test case.

sethladd:
I added a few more scenarios to the test. Everything looks good!

PR #375, opened 2015-04-13
Use HTTPS for Google Fonts in template

kaendfinger:


sethladd:
Definitely. Good catch. Thank you!

kaendfinger:
Of Course :)

Issue #377, opened 2015-04-14
Generating SDK docs creates two directories: dart:async and dart_async


sethladd:
This is interesting:

```
~/Code/dartdoc/docs[sdk-docs-task*]$ ls -l
total 64
drwxr-xr-x    8 sethladd  staff    272 Apr 13 21:11 css
drwxr-xr-x   18 sethladd  staff    612 Apr 13 21:09 dart:async
drwxr-xr-x    3 sethladd  staff    102 Apr 13 21:10 dart:core
drwxr-xr-x   19 sethladd  staff    646 Apr 13 21:11 dart:html
drwxr-xr-x    3 sethladd  staff    102 Apr 13 21:11 dart:web_audio
drwxr-xr-x    6 sethladd  staff    204 Apr 13 21:11 dart:web_sql
drwxr-xr-x   45 sethladd  staff   1530 Apr 13 21:09 dart_async
drwxr-xr-x   55 sethladd  staff   1870 Apr 13 21:10 dart_collection
drwxr-xr-x   66 sethladd  staff   2244 Apr 13 21:10 dart_convert
drwxr-xr-x  117 sethladd  staff   3978 Apr 13 21:10 dart_core
drwxr-xr-x  815 sethladd  staff  27710 Apr 13 21:11 dart_html
drwxr-xr-x   26 sethladd  staff    884 Apr 13 21:11 dart_indexed_db

```

Notice that both `dart:async` and `dart_async` are created.

sethladd:
Just confirmed this behavior on my work machine.

sethladd:
Thank you!

PR #381, opened 2015-04-14
fix issue with library name

keertip:
fix for #377


sethladd:
Thanks!

sethladd:
Should we add a test so it doesn't regress?

keertip:
Yes, next task is to add sdk doc gen to the build.

PR #388, opened 2015-04-14
add option to specify input dir

keertip:


bin/dartdoc.dart
```diff
@@ -44,6 +44,16 @@ void main(List<String> arguments) {
     exit(1);
   }
 
+  var inputDir;
```
sethladd: let's use annotations here. I've learned my lesson when we omit them :)

bin/dartdoc.dart
```diff
@@ -44,6 +44,16 @@ void main(List<String> arguments) {
...
+  var input_dir = args['input-dir'];
```
sethladd: what's the difference between this one and inputDir? Are there more descriptive variable names?

bin/dartdoc.dart
```diff
@@ -76,22 +88,24 @@ void _printUsageAndExit(ArgParser parser) {
...
   parser.addFlag('version',
       help: 'Display the version for $NAME.', negatable: false);
+  parser.addOption('dart-sdk',
+      help: "Location of the Dart SDK. Use if SDK isn't automatically located.");
   parser.addFlag('sdk-docs', help: 'Generate ONLY the docs for the Dart SDK.');
   parser.addOption('sdk-readme',
       help: 'Path to the SDK description file. Use if generating Dart SDK docs.');
-  parser.addOption('footer',
-      help: 'Path to an HTML, inserted into the footer of every page.');
+  parser.addOption('input-dir',
```
sethladd: if we're going to use `input-dir`, let's also use `output-dir`. I think `dir` is extraneous, so my vote is for `input`. Thoughts?

bin/dartdoc.dart
```diff
@@ -76,22 +88,24 @@ void _printUsageAndExit(ArgParser parser) {
...
+      help: 'Path to source directory, defaults to current directory');
   parser.addOption('output',
       help: 'Path to output directory.', defaultsTo: 'docs');
+  parser.addOption('footer',
+      help: 'HTML text, inserted into the footer of every page.');
```
sethladd: Oh, not a path to a file? This might be hard to pass into the command-line, as our uses will want to pass in multi-line footers

keertip: has to be changed to path to file. Will do that when adding the option for header.

bin/dartdoc.dart
keertip: sure, will change

bin/dartdoc.dart
sethladd: Oh, maybe I was confused. What do I actually pass in on the command line? A path or the actual contents?

bin/dartdoc.dart
keertip: done!

bin/dartdoc.dart
keertip: cleaned up

bin/dartdoc.dart
keertip: Sorry, I was not clear. Right now you have to pass the HTML text in on command line. This is not ideal, and will be changed to pass in path to a file with the contents.

bin/dartdoc.dart
```diff
@@ -76,22 +88,24 @@ void _printUsageAndExit(ArgParser parser) {
...
       help: 'Display the version for $NAME.', negatable: false);
+  parser.addOption('dart-sdk',
+      help: "Location of the Dart SDK. Use if SDK isn't automatically located.");
   parser.addFlag('sdk-docs', help: 'Generate ONLY the docs for the Dart SDK.');
   parser.addOption('sdk-readme',
       help: 'Path to the SDK description file. Use if generating Dart SDK docs.');
-  parser.addOption('footer',
-      help: 'Path to an HTML, inserted into the footer of every page.');
+  parser.addOption('input',
+      help: 'Path to source directory, defaults to current directory');
```
sethladd: should we use the `defaultsTo` option?

bin/dartdoc.dart
sethladd: Gotcha. Thanks! :)

bin/dartdoc.dart
keertip: yes, thanks!

sethladd:
lgtm!

PR #395, opened 2015-04-14
faster

kevmoo:
On my machine runtime went from ~8s to ~4s when run on itself.


keertip:
Awesome! Thanks!

sethladd:
:+1:

PR #396, opened 2015-04-15
add option for inserting into header

keertip:


bin/dartdoc.dart
```diff
@@ -53,7 +53,18 @@ void main(List<String> arguments) {
   List<String> excludeLibraries =
       args['exclude'] == null ? [] : args['exclude'].split(',');
   String url = args['hosted-url'];
-  String footer = _resolveTildePath(args['footer']);
+  String footerFilePath = _resolveTildePath(args['footer']);
```
sethladd: Should we write a single function that resolves the Tilde and prints the warning if the file doesn't exist? Looks like two copies of basically the same logic.

test/html_generator_test.dart
```diff
@@ -60,7 +60,7 @@ void main() {
     test('header and footer', () {
       String content = templates.indexTemplate({},
           assumeNullNonExistingProperty: true, errorOnMissingProperty: false);
-      expect(content.contains('<p>User inserted message</p>'), true);
+      expect(content.contains('<p>User inserted message</p>'), false);
```
sethladd: for next time: isFalse

or:

expect(content, contains(...))

(the error messages are nicer this way)

bin/dartdoc.dart
keertip: Yes, we can.

test/html_generator_test.dart
keertip: Thanks, will do.

PR #404, opened 2015-04-17
try to make linkedParams less hacky

polux:
Gets rid of `<!-- end param -->` which looks a bit hacky. I've checked visually that the output looks the same, and the tests pass, but I doubt the tests check for this.


sethladd:
Thank you!!

PR #406, opened 2015-04-17
get enum values

keertip:


sethladd:
Cool! But does it scale? :) Let's try an enum with lots and lots of values. Probably more of an issue with the UI so feel free to open an issue for me to track for "enums with lots and lots of values"

PR #428, opened 2015-04-22
Fixes for generating docs for box2d

keertip:


lib/dartdoc.dart
```diff
@@ -100,6 +100,7 @@ class DartDoc {
         new SourceFactory(/*contentCache,*/ resolvers);
 
     var options = new AnalysisOptionsImpl()..analyzeFunctionBodies = false;
+    options.cacheSize = 512;
```
kevmoo: you might could use cascades for the cacheSIze, too. For consistency. :smile:

keertip: will change it! Thanks for the tip.

PR #429, opened 2015-04-22
fix showing optional params in signature

keertip:


lib/dartdoc.dart
```diff
@@ -99,8 +99,9 @@ class DartDoc {
     SourceFactory sourceFactory =
         new SourceFactory(/*contentCache,*/ resolvers);
 
-    var options = new AnalysisOptionsImpl()..analyzeFunctionBodies = false;
-    options.cacheSize = 512;
+    var options = new AnalysisOptionsImpl()
+      ..analyzeFunctionBodies = false
+      ..cacheSize = 512;
```
sethladd: nice.

lib/src/model.dart
```diff
@@ -357,7 +357,7 @@ abstract class ModelElement {
 
     String renderParams(Iterable<Parameter> params,
         [String open = '', String close = '']) {
```
sethladd: would love if these were named params. makes the call site much easier to read.

sethladd:
Thanks!

lib/src/model.dart
keertip: done.

PR #442, opened 2015-05-03
Tweaks

kevmoo:
A bunch of random cleanup

CC @devoncarew @sethladd @keertip

### DIFF
--- a/lib/src/html_generator.dart
+++ b/lib/src/html_generator.dart
@@ -582,7 +582,7 @@ String resolveDocReferences(String text, MustacheContext nestedContext) {
   var resolvedText;
   var element = _getElement();
   if (element != null) {
-    resolvedText = element.resolveReferences(text);
+    resolvedText = element.resolveReferences(docs: text);
     return resolvedText;
   }
   return text;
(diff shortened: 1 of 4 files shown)


lib/src/html_generator.dart
```diff
@@ -41,10 +41,10 @@ class Templates {
   TemplateRenderer topLevelPropertyTemplate;
   TemplateRenderer typeDefTemplate;
 
-  final Map<String, String> _partialTemplates = {};
+  final Map<String, String> _partialTemplates = <String, String>{};
```
sethladd: why do we need the generics here? how does it help?  (do we ever do an `is` check on this object?)

sethladd:
thanks! one comment to resolve. generally, we try not to add generics to the collections object itself unless we need to do an `is` check.

lib/src/html_generator.dart
kevmoo: Generics are reified. This allows you to get runtime exceptions in checked mode if the wrong kind of item is added.

kevmoo:
I'm a big believer that having the type of the instance is *more* important than on the field.

PR #458, opened 2015-05-07
do not show dynamic for generics

keertip:


lib/src/model.dart
```diff
@@ -1424,15 +1424,18 @@ class ElementType {
 
     // not TypeParameterType or Void or Union type
     if (isParameterizedType) {
-      buf.write('&lt;');
-      for (int i = 0; i < typeArguments.length; i++) {
-        if (i > 0) {
-          buf.write(', ');
+      var list = typeArguments.where((t) => t.linkedName != 'dynamic').toList();
+      if (list.isNotEmpty) {
```
sethladd: this might be easier if we use list.join(',')

sethladd:
thanks!

lib/src/model.dart
keertip: done.

sethladd:
lgtm!

PR #477, opened 2015-05-14
move the test package out of test/ and into test_package/

devoncarew:
Move the test package out of `test/` and into `test_package/`. This way tools that expect all the code in `test/` to analyze will work w/ the dartdoc project.


sethladd:
love it

PR #478, opened 2015-05-14
grind script cleanup

devoncarew:
- misc changes to the grind script - mostly some upgrades from grinder 0.7.0

### DIFF
--- a/tool/grind.dart
+++ b/tool/grind.dart
@@ -2,25 +2,23 @@
 // All rights reserved. Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-import 'dart:io';
 import 'dart:async' show Future;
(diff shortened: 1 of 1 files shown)


keertip:
lgtm! Thanks!

tool/grind.dart
```diff
@@ -167,6 +163,6 @@ Future _uploadStats(GrinderContext context, int sdkDocsGenTime) async {
         description: 'Commit ${commit}', links: [link]);
     return librato.createAnnotation('build_ui', annotation);
   } else {
-    return true;
+    return new Future.value();
```
sethladd: if this is an async function, why not just return ?

devoncarew: sgtm

devoncarew:
updated -

Issue #355, opened 2015-04-12
Mark anything as deprecated as obviously deprecated


sethladd:

seaneagan:
What about de-prioritizing all deprecated members to the bottom of a given section (methods, properties, etc.), ignoring source and alphabetical order?

sethladd:
maaaybe ? not a bad idea. Make it happen! :)

sethladd:
It would not be a bad thing to open a new issue for pushing deprecated members down to the bottom of the list.

PR #479, opened 2015-05-14
add visualization for deprecated

devoncarew:
Some work towards #355.

screen shot 2015-05-14 at 4 06 41 pm

screen shot 2015-05-14 at 4 06 51 pm

Would love some guidance on this PR. I have added the `{{#isDeprecated}}deprecated{{/isDeprecated}}` template fragment in many of the places where a deprecated style would be appropriate, but am missing it in many more places. Is there a less brute force approach to this?

### DIFF
--- a/lib/resources/styles.css
+++ b/lib/resources/styles.css
@@ -174,6 +174,10 @@ dl dt.callable .signature {
   font-style: italic;
 }
 
+.deprecated {
+  text-decoration: line-through;
+}
+
 p.firstline {
   font-weight: bold;
 }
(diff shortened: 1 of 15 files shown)


sethladd:
First off, AWESOME.

Second, there may be a slightly better way. I see a lot of duplicate here. What about using a "partial" to print the name?

See what I did for _documentation.html.

sethladd:
https://github.com/dart-lang/dartdoc/pull/470

devoncarew:
OK, updated. I used a (very small) partial.

sethladd:
lgtm thank you!

PR #488, opened 2015-05-15
generator refactor

devoncarew:
Some refactoring to `Generator` and `HtmlGenerator`. A generator now knows how to generate a specific content, but it doesn't know the package or out dir info until the `generate()` method is called. This change doesn't affect the current functionality.


sethladd:
lgtm

Issue #472, opened 2015-05-14
dartdoc doesn't generate if there's a `docs` directory present


devoncarew:
The first time dartdoc is run, it generates into `docs`. The 2nd time, it fails with the message `Warning: output directory exists: docs`. I'd want it to run the same each time; it should just generate into `docs` whether or not the directory is initially present.

dartdoc version: 0.0.1+10

sethladd:
I disagree. What if you manually created the output directory, and manually put some files in there. And then you run dartdoc and it overwrites your files. :(

Perhaps we could add a --overwrite option?

Closing for now. We've chosen the safer option. But maybe this was the wrong assumption. Are there examples of tools that overwrite files by default?

devoncarew:
It'd want the runs to be idempotent. It should do the same thing on the same input. The fact that dartdoc changes the state enough that it can't run the 2nd time is non-intuitive.

sethladd:
But what if it didn't create that output directory in the first place? Let's say that you created a docs/ dir and put an index.html file in there. And then you ran our tool and it overwrote your index.html.

a14n:
BTW the [documentation section in the Pub Package Layout Conventions](https://www.dartlang.org/tools/pub/package-layout.html#documentation) encourages to put the generated doc under `doc` :

>Instead, this is for tutorials, guides, and other hand-authored documentation _in addition to_ generated API references.

It could be worth to standardize the place to find the generated API ref.

sethladd:
Not a bad idea! Maybe generate into apidoc ?

seaneagan:
But it's bad practice to commit / publish generated docs, right?  I think we'd need to at least co-encourage adding `apidoc` to .gitignore.

seaneagan:
Also, an alternative to `apidoc` could be `doc/api` or `doc/gen`.

a14n:
@seaneagan I wondered the same thing. But what will append when a langage breaking change makes that dartdoc cannot generate the api doc for old code ?

a14n:
+1 for `doc/api`

sethladd:
Where we put it is one question.

I'm still curious about the root issue: should dartdoc overwrite files, or should it warn the user if the directory it's trying to write into isn't empty?

a14n:
Overwriting files is fine here because the generated directory should stay sync with the sources. Doing changes in the generated api doc looks really strange.

sethladd:
[...]

PR #490, opened 2015-05-15
generate docs when the output dir already exists

devoncarew:
- address #472, generate the docs even in the presence of an existing `docs/` directory
- added a constant (`defaultOutDir`) for the `docs/` dir name


sethladd:
lgtm thanks!

PR #492, opened 2015-05-16
Devoncarew package meta

devoncarew:
Refactor the package info functions into a class; add some test coverage.

- moved the top-level functions from `package_utils.dart` into a `PackageMeta` class. Abstracted the differences between a pub package and the sdk.
- removed the `--sdk-readme` flag - I assume this was needed before the sdk had a readme?
- added support for locating readme (and license files) that are in plain text format. pass the info through to the generator about whether the located files are markdown or not
- made `DartDoc.generateDocs()` return a result object, for easier testing
- added a test to verify `DartDoc.generateDocs()`

@sethladd @keertip


sethladd:
Is there a way to find the SDK readme automatically?

lib/src/html_generator.dart
```diff
@@ -523,6 +528,13 @@ String renderMarkdown(String markdown, {nestedContext}) {
   return doc.body.innerHtml;
 }
 
+/// Convert the given plain text into HTML.
+String renderPlainText(String text) {
+  if (text == null) return '';
+
+  return "<code class='fixed'>${text.trim()}</code>";
```
sethladd: shouldn't we use `<pre>` here?

test/dartdoc_test.dart
```diff
@@ -6,15 +6,40 @@ library dartdoc.dartdoc_test;
...
+    tearDown(() {
+      delete(tempDir);
+    });
+
+    test('generateDocs', () {
+      PackageMeta meta = new PackageMeta.fromDir(testPackageDir);
+      DartDoc dartdoc =
+          new DartDoc(testPackageDir, [], getSdkDir(), [], tempDir, meta);
+
+      return dartdoc.generateDocs().then((DartDocResults results) {
```
sethladd: use await ?

lib/src/html_generator.dart
devoncarew: `pre` has some selectors that make it look like this:

screen shot 2015-05-17 at 11 20 10 am

w/ `code.fixed`, it looks like this:

screen shot 2015-05-17 at 11 19 19 am

This is really just a fallback for (pretty infrequent) readme.txt files (and no readme.md).

test/dartdoc_test.dart
devoncarew: done -

sethladd: So it's a line height issue?

On Sunday, May 17, 2015, Devon Carew <notifications@github.com> wrote:

> In test/dartdoc_test.dart
> <https://github.com/dart-lang/dartdoc/pull/492#discussion_r30473023>:
>
> > +    Directory tempDir;
> > +
> > +    setUp(() {
> > +      tempDir = new Directory(path.join(Directory.systemTemp.path, 'temp'));
> > +    });
> > +
> > +    tearDown(() {
> > +      delete(tempDir);
> > +    });
> > +
> > +    test('generateDocs', () {
> > +      PackageMeta meta = new PackageMeta.fromDir(testPackageDir);
> > +      DartDoc dartdoc =
> > +          new DartDoc(testPackageDir, [], getSdkDir(), [], tempDir, meta);
> > +
> > +      return dartdoc.generateDocs().then((DartDocResults results) {
>
> done -
>
> —
> <https://github.com/dart-lang/dartdoc/pull/492/files#r30473023>.
>


-- 
Sent from my Motorola Startac

lib/src/html_generator.dart
devoncarew: And to clarify, if we were to use `pre` here, some selectors would trigger which would give it a different background than normal doc text as well as a border. It would look like inset code rather than documentation.

Conversely, `code.fixed` looks like documentation (rather than in-line code). So it renders plain text files (readme.txt, license.txt, ...) better. I would recommend sticking with the current `code.fixed` style, but will change it to whatever is desired.

bin/dartdoc.dart
```diff
@@ -103,8 +100,6 @@ ArgParser _createArgsParser() {
       help: "Location of the Dart SDK. Use if SDK isn't automatically located.");
   parser.addFlag('sdk-docs',
       help: 'Generate ONLY the docs for the Dart SDK.', negatable: false);
-  parser.addOption('sdk-readme',
```
keertip: Would like to keep this around a little while longer. The stable build of the SDK still does not have the readme included.

devoncarew: OK, added back in.

sethladd:
lgtm

PR #498, opened 2015-05-19
Devoncarew printer

devoncarew:
- extend the `HtmlPrinter` class a bit to make it easier to pass in `class` and `href` attributes
- add tests for `HtmlPrinter`
- some tweaks to dartdoc's documentation


devoncarew:
@keertip, @sethladd, PTAL :)

sethladd:
It doesn't look like HtmlPrinter is used by dartdoc today. If it's not, we should remove it and put it into a package that does use it. I wouldn't expect dartdoc to provide an HtmlPrinter class, unless its default generator uses it.

devoncarew:
Ah, I didn't realize it wasn't used by `HtmlGenerator`. I updated this PR to remove it -

keertip:
lgtm!

Issue #501, opened 2015-05-19
Linking to other items within src/, which are exported, are 404s


evanweible-wf:
Consider the following package layout:

#### `lib/test_package.dart`
```dart
library test_package;

export 'src/base.dart';
export 'src/extending.dart';
```

#### `lib/src/base.dart`
```dart
library test_package.src.base;

class BaseClass {}
```

#### `lib/src/extending.dart`
```dart
library test_package.src.extending;

import 'base.dart';

/// Extending class extends [BaseClass].
class ExtendingClass extends BaseClass {}
```

There are two links to `BaseClass` in the above setup. One in the doc comment for `ExtendingClass` and another implicit one because `ExtendingClass` extends `BaseClass`. I would expect these links to work, but they 404 because they end up resolving to the full path which is not generated because it's in `src/`:
```
test_package.src.base/BaseClass_class.html
```
Instead of the top-level path where it actually exists (since it is exported in `lib/test_package.dart`):
```
test_package/BaseClass_class.html
```

---

I'm not sure if this is a bug or if it's a side-effect of an incorrect package layout. I've been using this layout structure for a while now and was under the impression that it was the recommended pattern (the [rpc package](https://github.com/dart-lang/rpc/tree/master/lib) uses this structure as well).

sethladd:
Thanks for the report! We'll take a look.

sethladd:
Confirmed! Thanks for the report.

cc @keertip

sethladd:
I think the bug is in model.dart, in resolveReferences:

```
      if (this is Library) {
        refLibrary = this;
      } else {
        refLibrary = _getLibraryFor(refElement.library, this.package);
      }
```

We're using the actual library, instead of the library that the name is exported through.

sethladd:
Please try it again.

PR #507, opened 2015-05-23
doc references should understand exported libraries

sethladd:


sethladd:
Attempts to address https://github.com/dart-lang/dartdoc/issues/501

PR #525, opened 2015-05-28
Add test and fix resolving [new Constructor] refs

keertip:


lib/src/model_utils.dart
```diff
@@ -50,7 +51,16 @@ String replaceAllLinks(String str, String findMatchingLink(String input)) {
       }
       String codeRef = str.substring(index + _leftChar.length, end);
       if (codeRef != null) {
-        var link = findMatchingLink(codeRef);
+        var link;
+        // support for [new Constructor]
+        var refs = codeRef.split(' ');
+        if (refs.length == 2 && refs[0] == 'new') {
```
sethladd: you can always do refs.first which might be cleaner

test/model_test.dart
```diff
@@ -170,7 +170,7 @@ void main() {
 
     test('docs refs', () {
       expect(B.resolveReferences(),
-          'Extends class <a href="ex/Apple_class.html">Apple</a>');
+          'Extends class <a href="ex/Apple_class.html">Apple</a>, use new <a href="ex/Apple/Apple.html">Apple</a>');
```
sethladd: I think we should include `new` in the link itself, making it clearer that it's a constructor we're linking to

keertip: done.

lib/src/model_utils.dart
keertip: done.

lib/src/model.dart
```diff
@@ -178,13 +178,18 @@ abstract class ModelElement {
       return docs;
     }
 
-    String _getMatchingLink(String codeRef) {
+    String _getMatchingLink(String codeRef, [bool isConstructor = false]) {
```
sethladd: sorry, should have caught this the first time. generally, booleans are better has named optional params. It's the difference between:

`foo(true)`

and

`foo(specialFlag: true)`

in readability.

Your call.

lib/src/model_utils.dart
```diff
@@ -37,7 +37,8 @@ bool isInExportedLibraries(
     List<LibraryElement> libraries, LibraryElement library) => libraries
     .any((lib) => lib == library || lib.exportedLibraries.contains(library));
 
-String replaceAllLinks(String str, String findMatchingLink(String input)) {
+String replaceAllLinks(
+    String str, String findMatchingLink(String input, [bool isContructor])) {
```
sethladd: did you want a default value here?

sethladd:
lgtm with comments

lib/src/model.dart
keertip: Did think about making it named, but was not sure of the convention. Will change it, more readable the better!

PR #526, opened 2015-05-28
add a test for incorrect doc reference

sethladd:


sethladd:
cc @keertip

keertip:
lgtm! Thanks for adding tests!

PR #535, opened 2015-05-29
use path separator for output dir

keertip:
fix windows build


sethladd:
patch lgtm but looks like we have more build work.

Odd output: `Parsed 0 files in 0.7 seconds.`

keertip:
Yes, not sure what is going on here. Testing on windows machine.

PR #536, opened 2015-05-29
fix for finding library for element

keertip:


sethladd:
Can we add a test for this?

PR #540, opened 2015-05-29
add test for linking to another lib

keertip:


test/model_test.dart
```diff
@@ -205,6 +207,11 @@ void main() {
           true);
     });
 
+    test('reference to class from another library', () {
+      String comment = superAwesomeClass.documentationAsHtml;
+      expect(comment.contains('<a href="ex/Apple_class.html">Apple</a>'), true);
```
sethladd: isTrue, not true :)

sethladd:
lgtm! thanks!

PR #545, opened 2015-05-29
only build pushes to master, I think

sethladd:


sethladd:
This is a good sign.

screen shot 2015-05-29 at 2 42 38 pm

PR #546, opened 2015-05-30
cache .pub-cache on travis

a14n:


sethladd:
Love it, thanks!

kevmoo:
Very cool, @a14n!

Need ta make this part of the default recipe.

sethladd:
Definitely. Is that possible?

Do we have a default or common travis.sh file?  cc @devoncarew

devoncarew:
This is probably the closest to it: https://github.com/dart-lang/oss_project_templates/blob/master/dart-github-dart-lang/.travis.yml

sethladd:
I see this in my builds now:

screen shot 2015-06-01 at 10 00 14 am

Notice the "changes detected, uploading new archive"... do we expect that?

kevmoo:
Looking at https://github.com/travis-ci/travis-build/blob/master/lib/travis/build/script/dart.rb

I wonder if we can tweak this here...

a14n:
@kevmoo it's indeed the good place but reading http://docs.travis-ci.com/user/caching/ I don't see any cache active by default. We could eventually add a `cache: pub` and do something like in the python script for pip (see https://github.com/travis-ci/travis-build/blob/master/lib/travis/build/script/python.rb)...

About the default `.travis.yml`, may I suggest https://gist.github.com/a14n/441101c997ee8f935d9c :)

PR #556, opened 2015-06-05
add support for package-root

keertip:
add awareness of package root to the resource loader

### DIFF
--- a/bin/dartdoc.dart
+++ b/bin/dartdoc.dart
@@ -73,6 +73,11 @@ main(List<String> arguments) async {
     outputDir = new Directory(_resolveTildePath(args['output']));
   }
 
+  var packageRootDir;
+  if (args['package-root'] != null) {
+    packageRootDir = new Directory(_resolveTildePath(args['package-root']));
+  }
+
   PackageMeta packageMeta = sdkDocs
       ? new PackageMeta.fromSdk(sdkDir, sdkReadmePath: readme)
       : new PackageMeta.fromDir(inputDir);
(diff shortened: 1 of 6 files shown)


lib/resource_loader.dart
```diff
@@ -16,6 +16,15 @@ import 'package:http/http.dart' as http;
 import 'package:path/path.dart' as path;
 import 'package:pub_cache/pub_cache.dart';
 
+String _packageRootPath;
```
sethladd: why not just make this public? we have a getter and setter below for it.

lib/resource_loader.dart
```diff
@@ -107,9 +116,16 @@ Future<Uint8List> _doLoadOverFileFromLocation(
 // Meanwhile, assume packages/ is next to entry point of script
 Future<Uint8List> _doLoadFromFileFromPackagesDir(final String resourcePath) {
   var convertedPath = _convertPackageSchemeToPackagesDir(resourcePath);
-  var scriptFile = new File(Platform.script.toFilePath());
-  var baseDir = path.dirname(scriptFile.path);
-  var fullPath = path.join(baseDir, convertedPath);
+  var fullPath;
```
sethladd: would like to add a type annotation to this, as we're not initializing it.

test/resource_loader_test.dart
```diff
@@ -0,0 +1,35 @@
...
+    test('package root setting', () {
+      var root = Directory.current;
+      packageRootPath = root.path;
+      expect(packageRootPath, equals(root.path));
+    });
+
+    test('load from packages with package root setting', () {
+      var root = Directory.current;
+      packageRootPath = join(root.path, 'packages');
+      loadAsString('package:dartdoc/templates/index.html').then((string) {
```
sethladd: consider using async await

test/resource_loader_test.dart
```diff
@@ -0,0 +1,35 @@
...
+
+import 'package:path/path.dart';
+import 'package:dartdoc/resource_loader.dart';
+import 'package:unittest/unittest.dart';
+
+void main() {
+  group('Resource Loader', () {
+    test('load from packages', () {
+      loadAsString('package:dartdoc/templates/index.html')
+          .then((string) => expect(string, isNotNull));
```
sethladd: async await

bin/dartdoc.dart
```diff
@@ -73,6 +73,11 @@ main(List<String> arguments) async {
     outputDir = new Directory(_resolveTildePath(args['output']));
   }
 
+  var packageRootDir;
```
sethladd: we should type annotate variables that aren't initialized

PR #570, opened 2015-06-10
enable parse function bodies

### LINKED ISSUE #568: Wrong doc comments for methods with missing return type
E.g. 
```
/// A class
class Klass {
  /// A method
  method() {
  }

  /// Another method
  another() {
  }

  /// A shadowed method
  toString() {    
  }
}
```

becomes 
```
Instance Methods

  method
  () → dynamic
  A method
  
  another
  () → dynamic
  A method

  toString
  () → dynamic
  Another method
```

keertip:
@sethladd  By not parsing the body, dartdoc looses the information on whether to query a method as being async. We had turned it off to save time. Running it on the sdk, see that there is about 2 sec difference for parsing and not parsing function bodies. So turning it back on.

closes #568

### DIFF
--- a/lib/dartdoc.dart
+++ b/lib/dartdoc.dart
@@ -103,9 +103,7 @@ class DartDoc {
     SourceFactory sourceFactory =
         new SourceFactory(/*contentCache,*/ resolvers);
 
-    var options = new AnalysisOptionsImpl()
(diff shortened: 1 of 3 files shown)


sethladd:
Thank you for catching this.

PR #573, opened 2015-06-11
Devoncarew fail on errors

### LINKED ISSUE #489: dartdoc should fail to generate docs if any analysis errors
It should probably (possibly?) still generate docs, but should warn the user at the end that there were analysis errors encountered:

```
"4 analysis errors and 123 warnings encountered!"
```

devoncarew:
fix #489

- calculate the analysis errors for the package
- fail (`exit(1)`) if there are any error level issues
- print out the list of errors, with file path and line numbers


lib/dartdoc.dart
```diff
@@ -120,15 +124,45 @@ class DartDoc {
...
+    for (Source source in sources) {
+      context.computeErrors(source);
+      errorInfos.add(context.getErrors(source));
+    }
+
+    List<_Error> errors = errorInfos.expand((AnalysisErrorInfo info) {
+      return info.errors.map(
+          (error) => new _Error(error, info.lineInfo, packageMeta.dir.path));
+    }).where((_Error error) => error.isError).toList();
+    errors.sort();
```
sethladd: you could do:

`..sort()`

sethladd:
Can we add a test for this?

lib/dartdoc.dart
devoncarew: done

devoncarew:
> Can we add a test for this?

Yup; I'll need to do some refactoring so we can test this, otherwise the call to `exit(1)` will be a bit of an issue.

sethladd:
Thanks for adding the test!

PR #583, opened 2015-06-12
signatures now appear on the same line as the name

sethladd:


sethladd:
CL is https://chromiumcodereview.appspot.com/1184733002/

sethladd:
Committed.

PR #586, opened 2015-06-12
do not run reference to another library test on Windows

keertip:
@sethladd


sethladd:
lgtm for now.

I opened an issue to track down if this is a real problem, or just an issue with our tests.

PR #592, opened 2015-06-17
change from const to static-assets for the html and css assets dir

sethladd:


lib/src/html_generator.dart
```diff
@@ -5,9 +5,8 @@
 library dartdoc.html_generator;
 
 import 'dart:io';
-
+import 'dart:developer';
```
devoncarew: Is there a dependency on a version of the sdk for this? If so, we should update our sdk dep.

sethladd: good catch! I just saw the warning that dart:profiler was deprecated.

I'll update the SDK dep.

sethladd:
@devoncarew bumped SDK constraint

devoncarew:
lgtm!

devoncarew:
Hmm, we'll need to take out stable bot offline until 1.11is stable.

pubspec.yaml
```diff
@@ -4,7 +4,7 @@ author: Dart Team <misc@dartlang.org>
 description: A documentation generator for Dart.
 homepage: https://github.com/dart-lang/dartdoc
 environment:
-  sdk: '>=1.9.0 <2.0.0'
+  sdk: '>=1.11.0 <2.0.0'
```
devoncarew: also, it looks like the lower bounds will need to be a dev build - something like `1.11.0-dev.5.2`

sethladd:
Reverted back to 1.9 for now. It's just a hint from analyzer, so for now, no big deal. Once 1.11 lands, we can switch over to dart:developer

sethladd:
Fixed with b9860e3

Issue #591, opened 2015-06-17
Misleading behavior if user provides a single command-line arg


sethladd:
Consider this:

```
~/Code/dartdoc[new-layout*] $ dart bin/dartdoc.dart test_package
Generating documentation for 'dartdoc' into /Users/sethladd/Code/dartdoc/doc/api/.
```

The intention was "generate the docs for the code in the `test_package`. Regardless if that was a reason expectation or user error :) notice how dartdoc started generating docs for the current directory.

What do you think about throwing an error if a user passes a command-line arg like this? The tool doesn't currently do anything with that, and the behavior might be misleading.

Or... we could use the command-line arg as the location of the input.

sethladd:
@keertip @devoncarew any thoughts?

devoncarew:
+1 to either throwing if args.rest is not empty, or using that as the cwd for generating docs. the first option is simpler; the 2nd could be a later enhancement.

sethladd:
Let's go with throwing if args.rest is empty for now.

sethladd:
OK, we print an error now:

```
~/Code/dartdoc[throw-on-invalid-command-line-args] $ dart bin/dartdoc.dart foo
Error: detected unknown command-line argument(s)
Usage: dartdoc [OPTIONS]

-h, --help            Show command help.
    --version         Display the version for dartdoc.
    --dart-sdk        Location of the Dart SDK. Use if SDK isn't automatically located.
    --sdk-docs        Generate ONLY the docs for the Dart SDK.
    --sdk-readme      Path to the SDK description file. Use if generating Dart SDK docs.
    --input           Path to source directory
                      (defaults to "/Users/sethladd/Code/dartdoc")

    --output          Path to output directory.
                      (defaults to "doc/api")

    --header          path to file containing HTML text, inserted into the header of every page.
    --footer          path to file containing HTML text, inserted into the footer of every page.
    --package-root    The path to the package root.
    --exclude         Comma-separated list of library names to ignore.
    --hosted-url      URL where the docs will be hosted (used to generate the sitemap).
```

sethladd:
Fixed with d7397ab

PR #593, opened 2015-06-17
throw if user passes a command-line arg

sethladd:


sethladd:
Addresses https://github.com/dart-lang/dartdoc/issues/591

sethladd:
cc @devoncarew

bin/dartdoc.dart
```diff
@@ -19,7 +19,7 @@ main(List<String> arguments) async {
   var args = parser.parse(arguments);
 
   if (args['help']) {
-    _printUsageAndExit(parser);
+    _printHelp(parser);
```
devoncarew: will this cause docs to be generated, if we don't exit here?

devoncarew: nevermind

bin/dartdoc.dart
```diff
@@ -78,6 +78,11 @@ main(List<String> arguments) async {
     packageRootDir = new Directory(_resolveTildePath(args['package-root']));
   }
 
+  if (args.rest.isNotEmpty) {
+    print('Error: detected unknown command-line argument(s)');
```
devoncarew: perhaps also: `: ${args.rest.join(' ')}`

sethladd:
Nicer error message now.

also, s/warning/error since we halt the app if those situations are encountered.

devoncarew:
lgtm

sethladd:
Committed with d7397ab

PR #598, opened 2015-06-18
widen package constraints

keertip:


pubspec.yaml
```diff
@@ -10,9 +10,9 @@ dependencies:
   args: ^0.13.0
   cli_util: ^0.0.1
   html: ^0.12.1
-  http: ^0.11.2
-  intl: ^0.12.0
-  logging: '>=0.11.1 <0.12.0'
+  http: '>0.11.1'
```
devoncarew: Do we need the service releases, or can we use the full versions?

http: `>=0.11.0`
intl: `>=0.11.0`
logging: `>=0.9.0 <0.12.0`

keertip: Guess we could.

devoncarew:
lgtm

pubspec.yaml
```diff
@@ -10,9 +10,9 @@ dependencies:
   args: ^0.13.0
   cli_util: ^0.0.1
   html: ^0.12.1
-  http: ^0.11.2
-  intl: ^0.12.0
-  logging: '>=0.11.1 <0.12.0'
+  http: '>=0.11.0'
```
sethladd: Doesn't this mean that we're saying we'll work with any future version? That seems risky.

I thought setting an upper bound was a best practice?

Issue #609, opened 2015-06-23
exported libraries have broken links


devoncarew:
See the link for `Package` on this page: https://dartdoc.firebaseapp.com/dartdoc.generator/Generator_class.html

It works from this page however: https://dartdoc.firebaseapp.com/dartdoc/index.html

So, `Package` from the private `dartdoc.models` library is exported by the public `dartdoc` library. We generate a summary of it in the dartdoc library page, and a detail page for it as well, in the dartdoc dir. `dartdoc.generator` has a reference to it, but it thinks that is lives in the `dartdoc.models` package (which it does).

Some possible fixes:

- from `dartdoc.generator`, don't create a hyperlink for `Package`. Just treat it like a regular reference to a symbol we're not generating docs for
- in `dartdoc`, create an entry in the overview for `Package`. But generate the detail page for it in `dartdoc.models`. This would be a synthetic library that would contain all the `dartdoc.models` symbols that had been exported from other libraries.

Any other ideas or options? I like the 2nd one - it handles the case well where a class is exported by more than one library.

sethladd:
Looks like this one can be closed?

PR #611, opened 2015-06-24
fix refs for parameters

keertip:
fixes the broken link in #609. We can address the bigger issue of how/where the exported libraries are generated later.

@devoncarew

### DIFF
--- a/lib/src/model.dart
+++ b/lib/src/model.dart
@@ -1367,8 +1367,15 @@ class Parameter extends ModelElement {
   Parameter(ParameterElement element, Library library)
       : super(element, library) {
     var t = _parameter.type;
-    _modelType = new ElementType(t, new ModelElement.from(
-        t.element, new Library(t.element.library, library.package)));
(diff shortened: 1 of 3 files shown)


lib/src/model.dart
```diff
@@ -1367,8 +1367,15 @@ class Parameter extends ModelElement {
   Parameter(ParameterElement element, Library library)
       : super(element, library) {
     var t = _parameter.type;
-    _modelType = new ElementType(t, new ModelElement.from(
-        t.element, new Library(t.element.library, library.package)));
+    var lib;
+    try {
+      lib = library.package.libraries
+          .firstWhere((l) => l.hasInNamespace(t.element));
```
devoncarew: I'd `orElse: () => null` to catching the StateError...

devoncarew:
lgtm

lib/src/model.dart
keertip: done

PR #613, opened 2015-06-27
remove redundant check

guillermooo:
It looks like there should be two different checks or one of them should be removed.


devoncarew:
lgtm - merging

Issue #607, opened 2015-06-23
sort order for the different sections


devoncarew:
I'm used to putting constants at the top of my source files - dartdoc generates the const section after classes. Do we have a sense for the ideal presentation order? Here's what we have currently.

For libraries:

- classes
- enums
- properties
- constants
- functions
- typedefs
- exceptions

And for classes:

- constructors
- constants
- static properties
- static methods
- properties
- methods
- operators

I think we should move constants up in the list, and possibly statics above ctors.

keertip:
How about

For libraries:
- constants
- typedefs
- properties
- functions
- enums
- classes
- exceptions

And for classes:
- constants
- static properties
- static methods
- properties
- constructors
- methods
- operators

devoncarew:
Like it! Maybe swap methods and operators? Don't know that I have a strong opinion.

keertip:
Swapping methods and operators works for me.

seaneagan:
Technically a constant *is* a property, so what about combining those two groups, and adding some visual indication on properties that happen to be declared as "const", maybe just the word "const"?  Or at least co-locate those two groups (above "typedefs" is listed in between them).

sethladd:
@keertip +1 for https://github.com/dart-lang/dartdoc/issues/607#issuecomment-114580669

PR #622, opened 2015-07-07
change the member generation order

devoncarew:
Change the order for the different member sections, as per https://github.com/dart-lang/dartdoc/issues/607.

@keertip @sethladd


sethladd:
nice!

lgtm

PR #626, opened 2015-07-08
change names for library and class files

### LINKED ISSUE #585: Possible clash with index.html and a top-level name index
Consider the case where a library has a top-level name called `index`.

This will clash and overwrite the `index.html` we generate inside the directory for the library docs.

keertip:
@sethladd 

closes #585

### DIFF
--- a/lib/src/model.dart
+++ b/lib/src/model.dart
@@ -507,7 +507,9 @@ class Library extends ModelElement {
 
   String get path => _library.definingCompilationUnit.name;
 
-  String get fileName => name.replaceAll(':', '-');
+  String get dirName => name.replaceAll(':', '-');
+
+  String get fileName => '$dirName-library.html';
 
   bool get isInSdk => _library.isInSdk;
 
@@ -649,7 +651,7 @@ class Library extends ModelElement {
   }
 
   @override
-  String get _href => '$fileName/$fileName-library.html';
+  String get _href => '$dirName/$fileName';
 }
 
 class Class extends ModelElement {
@@ -1018,7 +1020,7 @@ class Class extends ModelElement {
       name.hashCode, library.name.hashCode, library.package.name.hashCode);
 
   @override
-  String get _href => '${library.fileName}/$fileName';
+  String get _href => '${library.dirName}/$fileName';
 }
 
 class Enum extends Class {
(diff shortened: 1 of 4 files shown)


lib/src/model.dart
```diff
@@ -507,7 +507,7 @@ class Library extends ModelElement {
 
   String get path => _library.definingCompilationUnit.name;
 
-  String get nameForFile => name.replaceAll(':', '-');
```
sethladd: I'm not a huge fan of this change. If I see a variable called "fileName", I'm going to assume it's the filename. As it stands, it's really just the part of the filename before the suffix.

I used `nameForFile` to make that clearer. There's probably a better name, but I think `fileName` promises too much.

sethladd: If we change this to fileName, it should return the full filename... `libname-library.html`. Could we do that?

lib/src/model.dart
```diff
@@ -1018,7 +1018,7 @@ class Class extends ModelElement {
       name.hashCode, library.name.hashCode, library.package.name.hashCode);
 
   @override
-  String get _href => '${library.nameForFile}/$fileName';
+  String get _href => '${library.fileName}/$fileName';
```
sethladd: This is why `fileName` didn't make sense... because we also use it here for a path part.

lib/templates/index.html
```diff
@@ -9,7 +9,7 @@
       <dl>
         {{#package.libraries}}
           <dt>
-            <span class="name {{#isDeprecated}}deprecated{{/isDeprecated}}"><a href="{{nameForFile}}/index.html">{{ name }}</a></span>
+            <span class="name {{#isDeprecated}}deprecated{{/isDeprecated}}"><a href="{{fileName}}/{{fileName}}-library.html">{{ name }}</a></span>
```
sethladd: Should we add a getter to return the full path name? That way, we don't have to duplicate this logic in the template. We already have it in the class for the _href.

sethladd:
Thanks @keertip ! Left a few comments.

lib/src/model.dart
keertip: changing this to `dirName` and adding a `fileName` that returns `libName-library.html` to make this clear.

sethladd: `dirName` is much better name than `nameForFile` :)

sethladd:
lgtm thank!

PR #627, opened 2015-07-09
upgrade analyzer

### LINKED ISSUE #624: upgrade the version of the analyzer used
This will support the current sdk 1.12 dev release, which has moved the location of the _internal/libraries.dart files.

keertip:
@devoncarew 

closes #624

### DIFF
--- a/pubspec.lock
+++ b/pubspec.lock
@@ -4,7 +4,7 @@ packages:
   analyzer:
     description: analyzer
     source: hosted
-    version: "0.25.0+1"
+    version: "0.25.1"
   ansicolor:
     description: ansicolor
     source: hosted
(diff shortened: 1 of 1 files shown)


devoncarew:
lgtm

It looks like for stable sdk versions we don't get the latest analyzer because of env constraints on some packages (like `1.11.1`).

keertip:
Yes, you have to be on 1.12 dev and above to get the latest analyzer. Will hold off on committing this till we have the nav in and publish a version, so that users can still get dartdoc with the latest ui with 1.11 sdk.

devoncarew:
I think you can land this; it's just that on the stable sdk pub will select an older version of the analyzer package.

keertip:
Makes sense.

PR #629, opened 2015-07-09
fix refs for exported properties

### LINKED ISSUE #628: Properties not linked to correct library when exported
When creating links for exported properties, the links point to the source library instead of the one that exports it.

keertip:
@devoncarew 

closes #628


devoncarew:
lgtm

PR #638, opened 2015-07-11
fix some style issues; remove an analysis warning

devoncarew:
- restore some wrapping behavior with method params - the 2nd line and following lines are indented
screen shot 2015-07-10 at 9 05 00 pm
- normalize our colors a bit - there were a few colors that were very similar but had slightly different values (colors to represent non-primary things)
- fixed an analysis warning
- made the `inherited ` marker look more like the `read-only` et all markers
screen shot 2015-07-10 at 9 03 17 pm

### DIFF
--- a/lib/resources/styles.css
+++ b/lib/resources/styles.css
@@ -182,11 +182,11 @@ dt {
 
 dd {
   margin-bottom: 16px;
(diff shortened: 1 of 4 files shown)


sethladd:
lgtm

PR #640, opened 2015-07-11
Devoncarew nav

### LINKED ISSUE #359: Add in-page nav for major sections
For a class or library with lots and lots of stuff, it might be hard to jump to a particular section. Assumption: a user wants to see *just* the methods or *just* the properties.

Add an in-page nav or menu that helps a user jump down to a section.

devoncarew:
fix #359 - add an in-page navigation to jump between section of that page.

screen shot 2015-07-10 at 10 08 25 pm

screen shot 2015-07-10 at 10 09 41 pm

screen shot 2015-07-10 at 10 09 50 pm

This replaces underlining the items on hover with making the color brighter. It uses some of the size and font specs from https://www.google.com/design/spec/components/tabs.html#tabs-specs. We don't do the 2px solid underline from that spec - we're not a true tab in that the tab almost always scrolls off the page on select, and we don't have a way to associate the visible area with a particular tab.


sethladd:
lovely.

lgtm

Issue #644, opened 2015-07-14
Scroll bar appears below tab nav on fake library


sethladd:
![screen shot 2015-07-14 at 9 44 23 am](https://cloud.githubusercontent.com/assets/5479/8679031/055ef236-2a0d-11e5-8ece-0989d1e5d12c.png)

devoncarew:
Hmm, it's scrollable though.

We could possibly hide the scrollbar w/ something like:

```
foo::-webkit-scrollbar { 
    display: none; 
}
```

Or use some fancy nested divs, which may be a more complex solution than we want.

sethladd:
Scrollbar bad. Scrolling good. :)

PR #657, opened 2015-07-15
remove the scrollbar from the subnav

devoncarew:
fix https://github.com/dart-lang/dartdoc/issues/644

### DIFF
--- a/lib/resources/styles.css
+++ b/lib/resources/styles.css
@@ -402,6 +402,10 @@ ul.subnav {
   min-height: 24px;
 }
 
+ul.subnav::-webkit-scrollbar {
+  display: none;
+}
+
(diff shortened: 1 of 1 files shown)


sethladd:
tested on firefox? curious if that also needs a fix.

devoncarew:
Unfortunately, this is webkit/blink only.

sethladd:
The solution is webkit only, or the problem is webkit only?

devoncarew:
There's a solution on firefox, but it involves nested divs, and assumes that you can know ahead of time the width of the scrollbar.

This solution only addresses the problem for webkit based browsers.

sethladd:
ok, let's land this and open a new issue for firefox.

PR #665, opened 2015-07-16
fix showing duplicate implementors

### LINKED ISSUE #663: Implemented by on class shows duplicate classes for exported classes.
The implemented by part of class documentation will show duplicate classes for a single class if it is exported by more than one library.

Screenshot

You can see this happening on [this page](https://iot-dsa.github.io/docs/sdks/dart/dslink.common/Node-class.html).

I suggest that the true class's library be the only one that is linked.

keertip:
closes #663


sethladd:
Nice! thanks for the quick turnaround.

lgtm assuming the bots are green :)

Issue #659, opened 2015-07-15
Left nav links should take user to the dedicate page for the item


keertip:
Trying out the left nav on the sdk docs. When you click on an item in the nav, the page scrolls, but then the context is lost as there is no highlight or selection to indicate what was clicked. Takes a while to figure out what the left nav does.

Also, the left nav  and the top nav bar scroll off the page, so you have to go back to the top of the page once again to navigate.

sethladd:
>  Takes a while to figure out what the left nav does.

What did you expect to happen? aka where should have the link taken you?

> the left nav and the top nav bar scroll off the page, so you have to go back to the top of the page once again to navigate.

We could fix the left-nav to not scroll.

sethladd:
tldr: after an offline conversation, the left nav will take users to the dedicated page for the item.

sethladd:
Fixed!

PR #668, opened 2015-07-16
left nav goes to dedicated page

sethladd:


sethladd:
Addresses https://github.com/dart-lang/dartdoc/issues/659

Issue #637, opened 2015-07-10
No Dart identity visible on docs


sethladd:
We need some of our colors and logo, somewhere/somehow, to echo back to the main Dart identity.

Perhaps showing the logo is sufficient.

sethladd:
We have a favicon now.

sethladd:
Now at the bottom:

![screen shot 2015-07-16 at 2 05 58 pm](https://cloud.githubusercontent.com/assets/5479/8735185/cea0b294-2bc3-11e5-9713-7600f0c9ff3b.png)

and

![screen shot 2015-07-16 at 2 09 35 pm](https://cloud.githubusercontent.com/assets/5479/8735262/5120f792-2bc4-11e5-98be-6225ea4804a2.png)

devoncarew:
Can we make the line a bit more terse? It'll wrap now on mobile devices.

sethladd:
Suggestions?

I really want the words "API" and "docs" for SEO.

Maybe move the CC below?

sethladd:
What about something like

![screen shot 2015-07-16 at 3 20 09 pm](https://cloud.githubusercontent.com/assets/5479/8736541/449f5504-2bce-11e5-9ce5-cd38a6448ab9.png)

sethladd:
Going with this for now:

![screen shot 2015-07-16 at 4 16 26 pm](https://cloud.githubusercontent.com/assets/5479/8737375/13633ac0-2bd6-11e5-8cf2-fc415ea7b5ae.png)

sethladd:
Going with this, a little better?

![screen shot 2015-07-16 at 4 18 51 pm](https://cloud.githubusercontent.com/assets/5479/8737397/646250d2-2bd6-11e5-8c0a-fb6b4b120b5d.png)

![screen shot 2015-07-16 at 4 18 46 pm](https://cloud.githubusercontent.com/assets/5479/8737398/67304e54-2bd6-11e5-8fcd-e091f5e6971f.png)

PR #671, opened 2015-07-16
add some dart identity to the bottom of every page

sethladd:


sethladd:
Addresses https://github.com/dart-lang/dartdoc/issues/637

lib/templates/_footer.html
```diff
@@ -4,9 +4,9 @@
     <div class="container-fluid">
         <div class="container">
             <p class="text-center">
-                <span class="no-break">{{^package.isSdk}}package{{/package.isSdk}} {{package.name}}</span>
-                &bull;
-                <span class="no-break">version {{package.version}}</span>
+                <img src="static-assets/favicon.png" alt="Dart" title="Dart"
+                  width="24" height="24" style="margin-right:8px">
+                <span>API docs for {{package.name}}, version {{package.version}}</span>
```
devoncarew: instead of `,`, we should use `&bull;`, for consistency w/ the cc license field

sethladd: I wanted , to read like "the package, at version". Using the bullet looked too ambiguous. It could have been read that the version of the dartdoc or the version of the output.

devoncarew: I see them as visual separators. I'd prefer we used both `,` or both bullets for consistency. I have a slight preference for bullets for the aesthetics.

sethladd: If we use bullets, I'd like to make it more clear: the version is the version of the package that we used to generate the docs. I wasn't reading that interpretation strongly before. The bullets, to me, mean that each thing separated is a separate thing. Let's chat about this face to face, I'm confident there's a way to do it so it's easy to read and terse.

Issue #645, opened 2015-07-14
Display of const is inconsistent


sethladd:
For constants, we show:

![screen shot 2015-07-14 at 1 27 47 pm](https://cloud.githubusercontent.com/assets/5479/8684235/20b7f09a-2a2c-11e5-8e29-a596ff9d8146.png)

For constant constructors, we show:

![screen shot 2015-07-14 at 1 28 52 pm](https://cloud.githubusercontent.com/assets/5479/8684270/47a319be-2a2c-11e5-8fa4-1162f6678c24.png)

PR #673, opened 2015-07-16
show const below name, like other places

sethladd:


sethladd:
Addresses https://github.com/dart-lang/dartdoc/issues/645

PR #677, opened 2015-07-17
publish 0.1.0+4

keertip:


sethladd:
Great!

You might have seen, but there are a few grinder tasks that can help. There's a "bump version", and there's a public (the publish checks a few things, like the versions match, and the changelog has the latest version).

cc @devoncarew for FYI

keertip:
Yes, did use the grinder task to publish!

sethladd:
Score! We saved you 4.32 seconds. That adds up, you know.

Issue #676, opened 2015-07-17
Detail pages should easily link to peers


sethladd:
For example, if you're looking at a method or a function, you want to see what the peers are. What other methods are next to me (if in a class) ? What other functions are in near me (if in a library) ?

sethladd:
Now that left-hand nav links directly to detail pages, I think this is pretty important.

Maybe @devoncarew had some ideas. The naive approach here is put peers in the left-hand nav.

sethladd:
Going to try a few things. Ideas welcome!

sethladd:
In prep, the last breadcrumb is no longer linked, and is darker.

![screen shot 2015-07-17 at 10 25 23 am](https://cloud.githubusercontent.com/assets/5479/8753155/2f2b719e-2c6e-11e5-9490-63b4526efa34.png)

sethladd:
OK, so here's one idea: if you're at a "leaf" page, show all your peers:

![screen shot 2015-07-17 at 10 53 25 am](https://cloud.githubusercontent.com/assets/5479/8753624/1b1d5416-2c72-11e5-952a-e24e661687b4.png)

sethladd:
OK, to move forward, here is a peers on the left-hand side. I'd like to land this, and then play with it more.

![screen shot 2015-07-17 at 11 36 35 am](https://cloud.githubusercontent.com/assets/5479/8754402/29608786-2c78-11e5-9c45-55ba00af4f38.png)

devoncarew:
Cool, I like this last mock! There should be some visual callout for `fly` on the left hand side, to let people know that it corresponds to the page they're currently on.

And the siblings nav is only active for leaf nodes? it changes meaning between the different type of pages?

sethladd:
> And the siblings nav is only active for leaf nodes? it changes meaning between the different type of pages?

This is the crux of the issue. :)

sethladd:
First cut landed. Will need tweaks.

PR #680, opened 2015-07-17
Leaf pages link to siblings

sethladd:
display a notice when something doesn't have docs

put peers on left side and simplify styles


sethladd:
Addresses https://github.com/dart-lang/dartdoc/issues/676

sethladd:
Didn't upload to firebase, but that's OK for now.

Issue #679, opened 2015-07-17
Padding/spacing is weird in left-nav with titles


sethladd:
Not consistent.

![screen shot 2015-07-17 at 11 35 49 am](https://cloud.githubusercontent.com/assets/5479/8754389/0669ef74-2c78-11e5-8ad3-da546ce6e3ce.png)

sethladd:
Now it looks like this:

![screen shot 2015-07-17 at 3 23 09 pm](https://cloud.githubusercontent.com/assets/5479/8758347/c16b7dc8-2c97-11e5-8cff-f6704b232750.png)

PR #683, opened 2015-07-17
fix some subtle spacing issues

sethladd:


sethladd:
Addresses https://github.com/dart-lang/dartdoc/issues/679

PR #693, opened 2015-07-20
add an .analysis_options file

devoncarew:
Add an .analysis_options file. This will cut down on false positives (`test_package_bad/`) and on unnecessary analysis (`doc/api/`).

@sethladd @johnmccutchan

### DIFF
--- a/.analysis_options
+++ b/.analysis_options
@@ -1,0 +1,6 @@
+# Analyzer options file.
+
+analyzer:
+  exclude:
+    - test_package_bad/
+    - doc/api/


.analysis_options
```diff
@@ -0,0 +1,6 @@
+# Analyzer options file.
+
+analyzer:
+  exclude:
+    - test_package_bad/
```
johnmccutchan: These might need: '**' appended to the end.

sethladd:
Hot!

.analysis_options
devoncarew: I can add if you think it's be necessary; I don't _think_ they are required in .gitignore files (https://github.com/dart-lang/dartdoc/blob/master/.gitignore#L13).

devoncarew: I'll commit as is an adjust if it turns out the trailing `**`'s are necessary.

Issue #11, opened 2014-11-12
Display source code for methods or functions


sethladd:
A developer should be able to view the source code for a method or function, in order to learn how the method or function is implemented.

We need to do something smart with abstract classes (like List).

sethladd:
I've been looking at this. We can get the source code with `_method.computeNode().toSource()`, but there are at least three issues:

* we'll need to format it
* "native" or "external" functions are not clearly marked, so sometimes you see what looks like an empty method. For example:

![screen shot 2015-07-21 at 11 26 37 am](https://cloud.githubusercontent.com/assets/5479/8809114/5c723874-2f9b-11e5-9286-8af160f81905.png)


* the source code can contain implementation details that are not otherwise visible. For example:

![screen shot 2015-07-21 at 11 26 22 am](https://cloud.githubusercontent.com/assets/5479/8809107/53dc8eee-2f9b-11e5-9e13-cca5cfb5a621.png)

sethladd:
Making progress:

![screen shot 2015-07-21 at 3 22 14 pm](https://cloud.githubusercontent.com/assets/5479/8813981/87aac9d6-2fbc-11e5-80ef-bd4c41a66c77.png)

sethladd:
Little better:

![screen shot 2015-07-21 at 3 28 04 pm](https://cloud.githubusercontent.com/assets/5479/8814043/1b244e6c-2fbd-11e5-8f82-542e7a4968c3.png)

devoncarew:
What about having the `source` section title styled the same as the other section headers? Classes, Libraries, Methods, ...

It would also be nice to try the source code left aligned with the title, rather than indented.

sethladd:
> What about having the source section title styled the same as the other section headers? Classes, Libraries, Methods, ...

If you think they are of the same "type", then sure, we can do that.

> It would also be nice to try the source code left aligned with the title, rather than indented.

I agree, but I thought you wanted source code blocks to be indented? +1 for not indented here.

sethladd:
source code is now aligned to the left.

devoncarew:
I like the indenting for in-line code snippets. I do think we want some style difference between in-line snippets and method source. The indentation difference may be enough -

sethladd:
Source is now displayed for functions and methods.

PR #699, opened 2015-07-21
display source for method

sethladd:
show source code for functions, too
add tests

### DIFF
--- a/lib/src/model.dart
+++ b/lib/src/model.dart
@@ -1106,7 +1106,26 @@ class Enum extends Class {
   }
 }
 
-class ModelFunction extends ModelElement {
+abstract class SourceCodeMixin {
+  String get sourceCode {
+    String contents = element.source.contents.data;
+    var node = element.node; // TODO: computeNode once we go to 0.25.2
+    // find the start of the line, so that we can line up all the indents
+    int i = node.offset;
+    while (i > 0) {
+      i -= 1;
+      if (contents[i] == '\n' || contents[i] == '\r') {
+        i += 1;
+        break;
+      }
+    }
+    return contents.substring(node.offset - (node.offset - i), node.end);
+  }
+
+  Element get element;
+}
+
+class ModelFunction extends ModelElement with SourceCodeMixin {
   ModelFunction(FunctionElement element, Library library)
       : super(element, library) {
     _modelType = new ElementType(_func.type, this);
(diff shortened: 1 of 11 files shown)


sethladd:
Addresses https://github.com/dart-lang/dartdoc/issues/11

sethladd:
@devoncarew if you want to take a look

devoncarew:
Sure! Can you add some screenshots?

sethladd:
See https://github.com/dart-lang/dartdoc/issues/11

sethladd:
Hm, looks like many the latest analyzer only works with 1.12-dev SDK? Need to dig in.

lib/templates/_footer.html
```diff
@@ -56,10 +56,31 @@
...
 }
 
+function initSourceToggle() {
+  var toggle = document.querySelector('.source-code h4');
+  var toggleArrow = document.querySelector('.source-code-toggle');
+  var code = document.querySelector('.source-code pre');
+
+  if (toggle && toggleArrow && code) {
+    toggle.addEventListener('click', function(e) {
+      if (toggleArrow.classList.contains('closed')) {
```
devoncarew: This might be a bit more terse:
```
bool closed = toggleArrow.classes.contains('closed');
toggleArrow.classes.toggle('closed', !closed);
toggleArrow.classes.toggle('open', closed);
code.style.display = closed ? 'block' : 'none';
```

lib/src/model.dart
```diff
@@ -1121,6 +1121,21 @@ class ModelFunction extends ModelElement {
 
   @override
   String get _href => '${library.dirName}/$fileName';
+
+  String get sourceCode {
```
devoncarew: We should make a common method out of this and the below one. We should also take the common indent (or at least the indent of the first line) and remove it from all the source lines.

sethladd: Done

lib/templates/_footer.html
sethladd: done

sethladd:
Now locked to analyzer 0.25.0 which works with Dart 1.11

lib/src/model.dart
```diff
@@ -1105,7 +1105,26 @@ class Enum extends Class {
   }
 }
 
-class ModelFunction extends ModelElement {
+abstract class SourceCodeMixin {
```
devoncarew: a mixin in the wild :)

devoncarew:
lgtm :)

PR #700, opened 2015-07-22
css tweaks

devoncarew:
- increase the line-height for the left nav (it looked too compressed as is)
- make the footer more symmetrical

screen shot 2015-07-21 at 6 00 30 pm

screen shot 2015-07-21 at 6 00 22 pm

screen shot 2015-07-21 at 6 00 44 pm

Also, thoughts on the section headers? Should they be all-caps, a bit smaller, and have their underlined slightly closer to the text?
screen shot 2015-07-21 at 6 01 00 pm

@sethladd

### DIFF
--- a/.analysis_options
+++ b/.analysis_options
@@ -5,3 +5,4 @@ analyzer:
     - test_package/doc/api
     - test_package_bad/
     - doc/api/
+    - pub.dartlang.org/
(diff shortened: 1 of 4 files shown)


devoncarew:
Updated the footer:

screen shot 2015-07-21 at 7 17 13 pm

sethladd:
lgtm

Issue #587, opened 2015-06-15
Track down issue with test on Windows... is it a real bug?


sethladd:
Tracking issue to find root cause for https://github.com/dart-lang/dartdoc/pull/586

Also, is this a real issue on windows?

keertip:
Looking at the generated docs for test package on Windows, the link to class in another library is  correct and works. This is the same link that is in used in the test. So not sure why the test is failing.

sethladd:
Hi @keertip can you take a minute to see if this is a real issue? I can't remember if we silenced a test on the windows bots, and if we need to check into that before releasing. Thanks!

keertip:
Last time I looked, the failure was in finding the library for the class, the test could not find the library in the package and so no link was generated.

sethladd:
No action here. Closing. Feel free to reopen if it shows up again.

PR #701, opened 2015-07-22
fix a bug generating windows element links

devoncarew:
Fix https://github.com/dart-lang/dartdoc/issues/587 - a bug with a windows test and an issue generating links to elements in dartdoc on windows.

- re-enable two windows tests and one general test
- fix an issue with some tests on windows and `\r\n` line endings
- fix a regression where the docs for many properties were "`null`"

### DIFF
--- a/lib/markdown_processor.dart
+++ b/lib/markdown_processor.dart
@@ -253,21 +253,18 @@ String _getMatchingLink(
       }
     }
   }
-  if (refElement == null) {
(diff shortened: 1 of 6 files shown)


keertip:
Thanks!

lgtm on tests passing.

sethladd:
thanks for taking a look at this!! :ice_cream:

PR #709, opened 2015-07-22
always show source for methods

devoncarew:
- fixed a minor UI item where field properties would should as `read-only , inherited` rather than `read-only, inherited`
- always show source code; remove the disclosure toggle; use the section h2 instead of an h4
- strip the common indent off the displayed source

### DIFF
--- a/lib/src/model.dart
+++ b/lib/src/model.dart
@@ -1119,7 +1119,15 @@ abstract class SourceCodeMixin {
         break;
       }
     }
-    return contents.substring(node.offset - (node.offset - i), node.end);
+
+    // Trim the common indent from the source snippet.
+    String source =
+        contents.substring(node.offset - (node.offset - i), node.end);
+    String remainer = source.trimLeft();
+    String indent = source.substring(0, source.length - remainer.length);
+    return source.split('\n').map((line) {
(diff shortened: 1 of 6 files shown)


lib/src/model.dart
```diff
@@ -1119,7 +1119,15 @@ abstract class SourceCodeMixin {
...
       }
     }
-    return contents.substring(node.offset - (node.offset - i), node.end);
+
+    // Trim the common indent from the source snippet.
+    String source =
+        contents.substring(node.offset - (node.offset - i), node.end);
+    String remainer = source.trimLeft();
+    String indent = source.substring(0, source.length - remainer.length);
+    return source.split('\n').map((line) {
```
sethladd: will this work on windows?

sethladd:
could you rebase this before we merge?

`git rebase -i master`

and then:

`git push --force`

lib/src/model.dart
devoncarew: Yup, it'll split `\r\n` lines into things that end in `\r`, and then we'll add the `\n` back on with the join(). We could also right-trim the strings...

devoncarew:
Will do re: rebasing

sethladd:
Thanks!

Issue #716, opened 2015-07-23
Different formatting of code in the output


Janamou:
Add this comment to some method:

```dart
/// [:ul:] `li`.
```

```ul``` and ```li``` are formatted differently in the output. The [::] uses prettifier, the `` does not.

And in the list of methods, the code is not formatted at all. It is just a text. Maybe it should be also formatted?

sethladd:
I think I noticed that... that's for the description.

sethladd:
I think I found the cause. Now we look like this:

![screen shot 2015-07-23 at 4 46 55 pm](https://cloud.githubusercontent.com/assets/5479/8864739/78b0a21c-315a-11e5-9d05-535276698c5e.png)

sethladd:
Fixed! Thanks for the report.

PR #726, opened 2015-07-23
pretty print all code elements

sethladd:


sethladd:
Addresses https://github.com/dart-lang/dartdoc/issues/716

PR #728, opened 2015-07-24
Dual Nav

sethladd:


sethladd:
WIP

devoncarew:
Just checking, that this is not ready to review yet? Is it ready enough to post screen-shots?

sethladd:
Definitely not ready for review. WIP, and mostly testing `git reflow` workflow.

sethladd:
Here we are with the biggest width:

screen shot 2015-07-24 at 2 12 03 pm

sethladd:
@devoncarew this PR is ready for you to check out. Please do a local build with it and try it out. Before I tweak it, I'd like to know if you're OK with this general direction.

Try it on phone, tablet, and desktop layouts. Use dev tool's mobile emulation.

TODO: animation, if I can figure it out. And changing the right-button from hotdog to three dots (or other).

Thanks!

sethladd:
@devoncarew the PR moves the entire page to the left or right to make room for the navs.

As an alternative, consider this animated drawer which draws on top:

animated drawer

(the PR does NOT have this animation... this is here for your review)

I personally like the animation a bit better. But is does draw on top.

devoncarew:
Some initial thoughts then:

- the css for the package page is wonky (also, classes):
screen shot 2015-07-25 at 7 01 27 pm
- in desktop (wide) mode at least, we'll need some visual way to identify the right nav as the subordinate nav. Perhaps a 3-7-2 column split instead of a 3-6-3? It'll also help with the center column appearing slightly too narrow in the desktop view.
- in a perfect world, for the medium and narrow views we'd want the right nav to 'come out' of the overflow menu icon. It would look like a pulldown menu, a div on a higher z-order that looked like a pulldown menu starting from the overflow menu. The slide in behavior (rather than the push-the-page-over behavior) is closest to this. Another option is slide-in behavior for the nav button (the left nav) and pull-down menu behavior for the overflow button.

sethladd:
@devoncarew thanks for the feedback. Much of that is trivial to tweak, and the implementation is nowhere near to final.

What I need to know: are you OK with the right nav sliding out, as the animated GIF shows?

Don't worry about how we trigger that, we can tweak the icon in the upper right.

What I'm worried about: me spending 8 hours today getting the behaviors right, and then not using it. So I want to get your sign off on the design before I do anything else.

Assume we'll pretty up the source and tighten up the styles in the CSS.

sethladd:
Here's what I'm going to do for now:

* make the left nav slide in, in response to clicking/tapping the "hotdog" menu icon in the upper left. (this seems like the most obvious design choice... let me know if that's not what we want)

I'll wait for you before I tackle the right-hand nav.

sethladd:
Left nav:

left nav

sethladd:
@devoncarew ignoring the right-nav for now... I'm pretty confident in the behavior of the left nav. I tested it all the way down to an iphone4. Seems to work pretty well.

If you're happy with left-hand nav behavior, we can talk about what goes into it, and what we can remove from the top nav.

devoncarew:
Yup, to be clear, for left nav, I'm thinking of something like this site: https://elements.polymer-project.org/browse?package=iron-elements

It could have a title and some breadcrumbs as well, though just what we have in the left nav would work as is.

And yup, the paper-menu-button was what I was thinking of. The menu there will be quite a bit bigger than they normally are for overflow menus. It would perhaps be fixed position on the page, with something like:

```css
right: 10px;
top: 10px;
bottom: 10px;
```

sethladd:
Is that an LGTM for the left-nav behavior?

Next step is to ensure the left nav drawer has the right stuff. But I want your LGTM on the _idea_ and _behavior_ of the left nav. :)

(ignoring individual style tweaks, of course...)

devoncarew:
Yup, lgtm on the idea, and I really like that animation of the toolbar collapse behavior.

Having the left nav always visible on desktop makes navigation faster. If it cleans up the implementation to have it always toggleable, lets do that, with a lower priority issue to circle back and see if we want re-enable always on desktop nav.

sethladd:
> Yup, lgtm on the idea, and I really like that animation of the toolbar collapse behavior.

Great. thanks for the review!

> Having the left nav always visible on desktop makes navigation faster.

I managed to tweak a bunch of other rules so on desktop we'll keep the left-nav open.

sethladd:
@devoncarew I'd like to merge this and then keep going.

devoncarew:
lgtm!

Issue #736, opened 2015-07-27
Display breadcrumbs in left drawer


sethladd:

sethladd:
Will take a look.

sethladd:
Now we have:

desktop:

![screen shot 2015-07-29 at 1 04 30 pm](https://cloud.githubusercontent.com/assets/5479/8968484/68aca144-35f2-11e5-88e2-aa5cd422cfef.png)


mobile:

![screen shot 2015-07-29 at 1 04 35 pm](https://cloud.githubusercontent.com/assets/5479/8968491/6c2d6f88-35f2-11e5-8f59-6177435dac78.png)

PR #743, opened 2015-07-29
sl-drawer-breadcrumbs

sethladd:


sethladd:
Addresses https://github.com/dart-lang/dartdoc/issues/736

sethladd:
@devoncarew

devoncarew:
lgtm!

q: why do you put the screenshots in the issue rather than the PR?

sethladd:
> q: why do you put the screenshots in the issue rather than the PR?

I hadn't realized that, actually. :)  Whatever window was open?

PR makes more sense. Thanks for pointing that out!

Issue #742, opened 2015-07-29
Resizing from mobile to desktop, after drawer is open, keeps background dark


sethladd:
* resize down to mobile
* open drawer
* resize back up to desktop

expected: drawer disappears and dark overlay disappears

actual: drawer disappears, but dark overlay remains

sethladd:
We are dynamically inserting a div when the drawer opens. The div, however, is not removed when the window resizes.

sethladd:
Will take a look now.

PR #744, opened 2015-07-29
sl-remove-overlay-on-resize

sethladd:


sethladd:
Addresses https://github.com/dart-lang/dartdoc/issues/742

sethladd:
cc @devoncarew

lib/resources/styles.css
```diff
@@ -562,6 +562,17 @@ button {
     margin-right: 24px;
   }
 
+  #overlay-under-drawer.active {
+    opacity: 0.7;
+    width: 100%;
```
devoncarew: can also do `right: 0; bottom; 0` instead of width and height

lib/resources/styles.css
```diff
@@ -562,6 +562,17 @@ button {
     margin-right: 24px;
   }
 
+  #overlay-under-drawer.active {
```
devoncarew: What properties does `#overlay-under-drawer` have if not active? It looks like a 0 width height div. Perhaps have it be display: none?

devoncarew:
lgtm

lib/resources/styles.css
sethladd: Happy to change it if that's better. Are you saying to change it? :)

lib/resources/styles.css
sethladd: Good call. I'll make sure it's display: none.

lib/resources/styles.css
devoncarew: I would. I don't really trust percentages w/ the dom. I think the fixed positioning will be more reliable.

sethladd:
lgtm

PR #755, opened 2015-07-31
exclude test directories from analysis

devoncarew:
- exclude test directories from analysis
- reset us down to zero analysis warnings (including deprecation warnings)

### DIFF
--- a/test_package/.analysis_options
+++ b/test_package/.analysis_options
@@ -1,0 +1,1 @@
+{ analyzer: { exclude: ['**'] } }
--- a/lib/src/html_generator.dart
+++ b/lib/src/html_generator.dart
@@ -208,6 +208,8 @@ class HtmlGeneratorInstance {
     });
 
     //if (url != null) generateSiteMap();
+
+    await _copyResources();
   }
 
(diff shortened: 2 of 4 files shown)


devoncarew:
@sethladd

sethladd:
Can you make this PR just the `.analysis_options` ?

My other PR removes the dart:profiler related code.

test_package/.analysis_options
```diff
@@ -0,0 +1 @@
+{ analyzer: { exclude: ['**'] } }
```
sethladd: huh, I thought this file was a YAML file ?

devoncarew: yaml's a superset of json :)

sethladd: +1 for the YAML format. It's easier to read.

If there's a reason to use the JSON version, that's fine, too.

devoncarew: I think of this file as just a 1-line marker file - don't look in this directory.

Hopefully the AS issue I filed will be addressed - the parent pubspec can exclude this directory, and this file can just be deleted.

PR #756, opened 2015-08-01
sl-doc-comments

sethladd:


sethladd:
@devoncarew thanks for taking a look!

sethladd:
TBR

PR #769, opened 2015-08-05
switch to the new Resources api

devoncarew:
Placeholder until 0.12.0-dev5 is out.

### DIFF
--- a/lib/resource_loader.dart
+++ b/lib/resource_loader.dart
@@ -2,8 +2,6 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-// TODO: Consider making this a stand-alone package, if useful.
-
 /// Make it possible to load resources, independent of how the Dart app is run.
 ///
 ///     Future<String> getTemplateFile(String templatePath) {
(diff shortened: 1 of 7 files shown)


lib/resource_loader.dart
```diff
@@ -12,163 +10,20 @@
...
+/// Loads a `package:` resource as a [List<int>].
+Future<List<int>> loadAsBytes(String path) {
   if (!path.startsWith('package:')) {
     throw new ArgumentError('path must begin with package:');
   }
-  return _doLoad(path);
-}
-
-/// Determine how to do the load. HTTP? Snapshotted? From source?
-Future<Uint8List> _doLoad(final String path) {
```
sethladd: this will break on pub global activated installs, because those don't write .packages yet. We might not want to remove all this code quite yet.

devoncarew: Hmm, so try the new resource api, and if that fails, fall back to the `pub_cache` method?

sethladd: For now, yes. Just test this with `pub global activate` before you commit.

(you can globally activate without pushing to pub, fwiw)

devoncarew: kk, sounds good

PR #784, opened 2015-08-07
fix issues with docs for names with multiple underscores

sethladd:


sethladd:
cc @keertip

keertip:
lgtm!

Looks like pub is failing on the windows bot.

sethladd:
pub is timing out for the download. Seems transient.

PR #786, opened 2015-08-07
Tweak <pre> tags a bit.

munificent:
- Keep the flush left and right. Looks better and makes a bit more room.
- Shrink the text a bit to fit more code in.
- Lighten the background.


sethladd:
lgtm

PR #787, opened 2015-08-07
Document property types on the right like methods.

munificent:
This keeps the property name nice and flush left so it's easy to find.


sethladd:
I'm a fan. In fact, I think we had this at one point.

The question is, does a property "return" a type, like a method/function does? Or is a property "a" type?

munificent:
I think the semantics make sense. The left side of the arrow is what you write to invoke it and the right side is what you get back. The only difference between a property and a method is there no parentheses.

sethladd:
lgtm

Issue #785, opened 2015-08-07
remove workaround now that list sort bug is fixed


sethladd:
Tracking issue so we don't forget.

Look for:

```
// XXX working around a VM SDK issue.
```

in model.dart

PR #789, opened 2015-08-07
Undo workaround for sorting an empty ungrowable list

sethladd:
Don't land this until next dev channel build.


sethladd:
Addresses https://github.com/dart-lang/dartdoc/issues/785

PR #793, opened 2015-08-07
Make section titles lighter (and a bit bigger).

munificent:
Since they don't contain real content, they don't need to stand out
that much. Lightening does that, and making the font a bit bigger makes
it a bit easier to read when you do want to.

Also, darkened the ">" in the breadcrumbs to show that it's not part
of the link.

### DIFF
--- a/lib/resources/styles.css
+++ b/lib/resources/styles.css
@@ -324,7 +324,7 @@ footer .container-fluid {
 }
 
 .gt-separated li:before {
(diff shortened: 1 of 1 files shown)


sethladd:
Is it easy to add a screenshot?

sethladd:
lgtm

munificent:
I'll do that next time if you'd like. Want me to merge this, or should you?

sethladd:
Thanks! A Cmd-Shift-4 on a mac is a very quick way to grab a section.

meanwhile, I'll smash button.

PR #794, opened 2015-08-07
use the new Resource api

devoncarew:
- redux of https://github.com/dart-lang/dartdoc/pull/769 - keep the old resource loading as a backup, until pub global activate generates .packages files
- use the new Resource API
- only support SDK 1.12.0 and newer

### DIFF
--- a/lib/resource_loader.dart
+++ b/lib/resource_loader.dart
@@ -2,8 +2,6 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-// TODO: Consider making this a stand-alone package, if useful.
-
 /// Make it possible to load resources, independent of how the Dart app is run.
 ///
 ///     Future<String> getTemplateFile(String templatePath) {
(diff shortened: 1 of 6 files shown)


lib/resource_loader.dart
```diff
@@ -24,20 +22,25 @@ import 'package:pub_cache/pub_cache.dart';
...
 /// Loads a `package:` resource as a String.
-Future<String> loadAsString(String path) async {
+Future<String> loadAsString(String path) {
   if (!path.startsWith('package:')) {
     throw new ArgumentError('path must begin with package:');
   }
-  Uint8List bytes = await _doLoad(path);
-  return new String.fromCharCodes(bytes);
+  return new Resource(path).readAsString().catchError((_) async {
+    var bytes = await _doLoad(path);
```
sethladd: maybe a note that we can remove this when a certain issue is closed?

devoncarew:
@sethladd

sethladd:
Thanks for doing this!

lgtm but can you add a note saying when we can delete the catchError bits?

lib/resource_loader.dart
devoncarew: will do

devoncarew:
We need to disable the windows build until 1.12 is available on chocolatey (or scoop).

devoncarew:
and, squashed

sethladd:
Yay! This is great!

PR #797, opened 2015-08-08
silence warnings for anon libs without docs, also silence appveyor build for now

sethladd:


devoncarew:
Lgtm; nice fix for the undocumented libraries warning

PR #799, opened 2015-08-08
find as you type search

sethladd:
WIP

@devoncarew @keertip the plan is to generate a simple JSON file, enough to power a find-as-you-type search.

Do either of you have an opinion on how to implement the HTML bits for this? (e.g. found a good example?)


sethladd:
cc @keertip @devoncarew please take a look. I think this is good enough to try out and send feedback.

lib/src/html_generator.dart
```diff
@@ -166,13 +168,34 @@ class HtmlGeneratorInstance {
 
     if (package != null) {
       _generateDocs();
+      _generateJson();
```
devoncarew: Perhaps `_generateIndex()` or `_generateSearchIndex()`?

lib/src/model.dart
```diff
@@ -48,7 +50,19 @@ void _addToImplementors(Class c) {
   }
 }
 
-abstract class ModelElement implements Comparable {
+/// An element that is enclosed by some other element.
+///
+/// Libraries are not enclosed.
+abstract class EnclosedElement {
```
devoncarew: +1, nice addition to the model

lib/resources/styles.css
```diff
@@ -606,8 +616,81 @@ button {
...
+}
+
+.tt-hint {
+  color: #999
+}
+
+.tt-menu {
+  right:0;
+  left: inherit !important;
+  width: 422px;
```
devoncarew: It's not clear from the screenshots how this works on mobile. My sense is that smaller then some screen width, the search menu should display as wide as the screen.

devoncarew:
Looks great! A question about mobile behavior.

lib/src/html_generator.dart
sethladd: done!

lib/resources/styles.css
sethladd: I haven't taken a look at mobile yet. That will certainly need another batch of fixes/changes.

sethladd:
Ready for review. It's not perfect, but I believe it's enough to get feedback on the actual behavior.

This also fixes a few issues with paddings and lines.

Search box is hidden for very small widths, will need to revisit. Should be visible for nearly all desktop users.

cc @devoncarew

devoncarew:
It's a big one! Awesome to have this feature. :fire:

One bit of feedback on the UI. I'm not sure what the material design spec for search boxes is, but the square box doesn't say search to me. Consider putting a `border-radius: 2px` on it to round the edges.

lgtm

sethladd:
> Consider putting a border-radius: 2px on it to round the edges.

Sure, I'll give it a try.

Thanks for the review!

Issue #775, opened 2015-08-07
source code be shown with 80 chars on desktop


sethladd:
Looks like we're too narrow.

![screen shot 2015-08-07 at 9 05 05 am](https://cloud.githubusercontent.com/assets/5479/9140101/8342eae8-3ce3-11e5-9898-09a6560cb95e.png)

jakemac53:
I noticed this as well, on my computer I needed to make the max width a bit over 1400px in order to fit all 80 chars. I think the actual issue here is the 3 column layout, but switching to a 2 column layout would be a bit more radical of a change :).

jakemac53:
Actually, I tried changing the middle column to 7 column width (from 6) and the right sidebar to 2 (from 3) in the desktop layout, and I think thats a lot better. Causes a bit of overflow in the names on the right column but you can still see enough to be useful.

PR #800, opened 2015-08-10
change sidebars to 2 cols and center to 8 cols

jakemac53:
Just floating this as an option, I don't think the sidebars need to be as large as they currently are, but I am sure others might disagree :).

This just applies to larger screens, and resolves https://github.com/dart-lang/dartdoc/issues/775, at least in the case of a full sized browser window.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ packages
 .packages
 .settings/
 .DS_Store
+*.iml
 
 pub.dartlang.org/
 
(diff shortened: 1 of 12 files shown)


sethladd:
Thanks! Can you add some screenshots?

jakemac53:
I added grey background to the right column so you can see the size of it, but that is not actually a propsed change.

Original:
original

Proposed:
new

As you can see it does cause some elipses in the right hand column on long names, but I think there is still more than enough to be useful/searchable on the page. Also google search would still see the full names. Chrome is also smart enough to let you search for the full text on the page even though its cut off with the elipses. For example:

search
:

sethladd:
Nice. Can you merge with origin master?

polux:
Woohoo! Thanks for fixing this @jakemac53, that was my number 1 pet peeve with the new dartdoc :)

Issue #806, opened 2015-08-11
homepage column widths don't add up to 12


sethladd:
This affects alignment of the search box:

![screen shot 2015-08-10 at 6 43 31 pm](https://cloud.githubusercontent.com/assets/5479/9190192/4e6ebfec-3fa7-11e5-9cf2-7cbac817e706.png)

Only affects the homepage because we don't have a left column.

PR #809, opened 2015-08-11
display descriptions with ellipses

### LINKED ISSUE #808: Provide a hint where text is hidden behind a link in the overview
It should be more visible that clicking on a field/method in the class overview leads to more text.

I find it confusing is that when you look at a class like in

../dart-mirrors/MirrorsUsed-class.html

it is non-obvious that there is more text for a property if you click on it. Maybe we can add a little visual clue, like a 'more >>' or just three '...'?

devoncarew:
- fix #808; display one line of a description summary - use ellipses if it doesn't fit. This had been our behavior; not sure if this regressed or it was a conscious choice to move away from it.
- display the glass pane w/ slightly less opacity
- the columns on the index page now add up to 12
- move the nav from right to left on the index page. This way the libraries nav is in a consistent place when it's visible on the page.

screen shot 2015-08-11 at 1 02 48 am

### DIFF
--- a/lib/resources/styles.css
+++ b/lib/resources/styles.css
@@ -210,6 +210,9 @@ dd.callable, dd.constant, dd.property {
 
 dd p {
   margin-top: 4px;
+  white-space: nowrap;
+  overflow-x: hidden;
+  text-overflow: ellipsis;
 }
 
 section.summary h2 {
@@ -503,6 +506,7 @@ span.top-level-variable-type {
   padding: 16px;
 }
 
+.sidebar h5,
 .sidebar ol li {
   text-overflow: ellipsis;
   overflow: hidden;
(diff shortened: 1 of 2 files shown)


devoncarew:
fix #806

sethladd:
Thanks!

> move the nav from right to left on the index page.

I don't care if that particular nav is on the left or right, but I'd like it consistent across all the pages. Can we move all the other pages, too?

devoncarew:
The library nav is on the left for all the pages.

page | left nav | right nav
---- | -------- | --------
index | libraries | _none_
library | libraries | library members
class | library members | class members
class member | class members | _none_

devoncarew:
I think this is the least surprising structure for users. The first and last pages (tree root and leaf nodes) are a bit of special cases.

sethladd:
Bonus points for using a table! :)

I find it inconsistent, but the world is a messy place.

lgtm!

PR #812, opened 2015-08-11
remove calls to deprecated methods

devoncarew:
- remove calls to deprecated methods (`node` from the analyzer package ==> `computeNode ()`)


sethladd:
lgtm

thanks for taking care of this one!

Issue #824, opened 2015-08-19
https://dartdoc.firebaseapp.com/ isn't updated


sethladd:
Any clues as to why?

We link to this from the README... if we don't guarantee that https://dartdoc.firebaseapp.com/ will be up to date, let's unlink from the README.

Thoughts?

sethladd:
cc @devoncarew @keertip

devoncarew:
Don't know - I'll take a look. We could also switch this to the SDK docs
link. Or remove completely. The sample link was more useful early on.

sethladd:
Thanks. If you don't see an immediate cause, let's just remove the link.

d2m:
not sure this is the real issue, but is suspect the firebase job bails at
    `https://github.com/dart-lang/dartdoc/blob/master/tool/grind.dart#L230`

```
@Task('Push a copy of the dartdoc docs to firebase')
firebase() {
  Map env = Platform.environment;

  if (env['FIREBASE_USER'] == null) return;
  if (env['TRAVIS_DART_VERSION'] != 'stable') return;
```

and travis-ci runs the dart dev release

devoncarew:
Ah ha! Good catch @d2m. We changed to just building dartdoc for dev channel when we started consuming the new `Resource` API.

d2m:
@devoncarew - Great, thanks!

sethladd:
Thanks @d2m and @devoncarew !

PR #828, opened 2015-08-20
fix upload to firebase

devoncarew:
Fix https://github.com/dart-lang/dartdoc/issues/824 - upload the built docs to firebase during the build, when building on the dev channel bot.


sethladd:
Thanks!

Issue #832, opened 2015-08-20
Combined getter/setter docs are mangled


munificent:
Here's the docs for `List.length`:

<img width="552" alt="screen shot 2015-08-20 at 12 52 26 pm" src="https://cloud.githubusercontent.com/assets/46275/9393982/55c78ee6-473a-11e5-82c1-e1b19a7cc3a2.png">

It looks like it's making a single page for both the getter and setter (good idea!) but isn't combining the docs correctly.

sethladd:
Yikes, good call! Thanks for the file.

sethladd:
OK, I have a repro. Occurs when the classical `/**` docs are used, not `///`. Fix incoming!

sethladd:
Fixed!  (though, we probably need to make a nice distinction in the page about which docs are for the setter, and which are for the getter. will do in another patch.)

![screen shot 2015-08-20 at 4 05 43 pm](https://cloud.githubusercontent.com/assets/5479/9397825/66113c0a-4755-11e5-9687-b184e55c8280.png)

PR #839, opened 2015-08-20
fix docs for props with explict getter and setter

sethladd:


sethladd:
Addresses https://github.com/dart-lang/dartdoc/issues/832

sethladd:
cc @keertip

keertip:
Awesome! Thanks.

PR #846, opened 2015-08-21
update creating source for analyzer

keertip:


keertip:
@devoncarew

sethladd:
lgtm

thanks!

sethladd:
(what bug does this fix?)

devoncarew:
lgtm

It's for fixing generating docs for the sky package. We normalize references to libraries from relative file paths and self reference package: links.

devoncarew:
Oddly enough, I'm still seeing this issue locally after a pub global activate (getting the latest version) and running `dartdoc`. I _don't_ see it after activation if I instead run `pub global run dartdoc`. Perhaps an older version is cached for me? Or perhaps a path issue.

devoncarew:
Ah! `dartdoc` picks up the version from the sdk, which doesn't have this fix. `pub global activate` picks up the latest latest, which does.

PR #849, opened 2015-08-21
add ellipsis when there's more than one liner

sethladd:


sethladd:
Addresses https://github.com/dart-lang/dartdoc/issues/808

sethladd:
cc @devoncarew @keertip

lib/src/model.dart
```diff
@@ -62,13 +62,23 @@ abstract class Nameable {
...
+}
+
+abstract class ModelElement implements Comparable, Nameable, Documentable {
   final Element element;
   final Library library;
 
   ElementType _modelType;
   String _rawDocs;
-  Documentation _documentation;
+  Documentation __documentation;
```
devoncarew: the rare double underscore

lib/templates/_has_more_docs.html
```diff
@@ -0,0 +1,3 @@
+{{#hasMoreThanOneLineDocs}}
+<a href="{{href}}">&hellip;</a>
```
devoncarew: Interesting - did the … char not make it through the various file encodings?

You could also remove the `a href`. If just the line is long and gets truncated by the browser, it'll show an ellipses from the text-overflow css style. So, a plain ellipses char from a 2nd paragraph would match that.

devoncarew:
lgtm! I like the solution.

lib/src/model.dart
sethladd: Yeah, I'm not proud of that one. :(

lib/templates/_has_more_docs.html
sethladd: > … char not make it through the various file encodings?

What is the ... char?

> You could also remove the a href.

And still show the hellip ?

PR #858, opened 2015-08-24
remove --package-root, use --input for the analysis root

devoncarew:
- Removed the `--package-root` option

This was implemented from the pov of the dartdoc tool (where to I load my code from?), rather than from the pov of the user (where are the packages by libraries needs to analyze correctly?). Even re-purposed, the --package-root would not have handled cases like libraries using sdk extensions.

I removed it in favor of using the existing `--input` option to help us locate an analysis context. dartdoc can now generate docs for sky, when driven by their skydoc.py script, which runs from a different cwd than the `sky` package.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## unreleased
+* Removed the `--package-root` option. Dartdoc now uses the `--input` flag as
+  the place to start looking for an analysis root. This better supports the
+  `.packages` file and use cases where dartdoc is invoked from locations other
+  than the project directory.
+
 ## 0.6.0+1
 * [bug] fix for getting uri for sources in lib folder
 
--- a/bin/dartdoc.dart
+++ b/bin/dartdoc.dart
@@ -74,11 +74,6 @@ main(List<String> arguments) async {
     outputDir = new Directory(_resolveTildePath(args['output']));
   }
 
-  Directory packageRootDir;
-  if (args['package-root'] != null) {
-    packageRootDir = new Directory(_resolveTildePath(args['package-root']));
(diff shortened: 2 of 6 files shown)


sethladd:
We're going to need to add back package-root, though, right? Just like our other tools, we'll possibly need to specify the location of this, in order to resolve the package: imports for the code that we're analyzing and generating docs for.

devoncarew:
We could add it if people end up needing it. It's important to note that the current implementation would not have let people analyze their code against a package-root.

devoncarew:
https://github.com/dart-lang/dartdoc/issues/859

sethladd:
>  It's important to note that the current implementation would not have let people analyze their code against a package-root.

Yeah, glad you caught that. That would have been surprising. Good to know we can add it back as a means to resolve package: URIs _for the code we're documenting_.

sethladd:
lgtm

keertip:
@devoncarew, we are using the --package-root option on the bots to generate docs for the sdk.

sethladd:
> we are using the --package-root option on the bots to generate docs for the sdk.

What are the semantics of --package-root when building on the bots?

1) Find the code for dartdoc _here_
2) Resolve package: URIs in the code that dartdoc is analyzing _here_

keertip:
Looking at the script, the --package-root to dartdoc can be removed, it is not needed. Just something to be done when we roll in this version into the sdk.

sethladd:
> Just something to be done when we roll in this version into the sdk.

What command-line flags will we need for the SDK build process?

devoncarew:
merging

PR #862, opened 2015-08-24
dartfmt

devoncarew:


sethladd:
I'm running dartdoc on every commit locally, plus we have it installed on the bots.

Why didn't they catch these formats?

devoncarew:
We disabled it when the newer dartfmt came out - it formats differently than the older one. The stable sdk build would fail when the dev one succeeded, and vice versa.

PR #892, opened 2015-09-03
links for inherited elements, reduce file generation

keertip:
@sethladd , @devoncarew 

We now link to the inherited element, if in the package. Also reduced the files generated, back to 150 secs to generate sdk docs.

### DIFF
--- a/lib/src/html_generator.dart
+++ b/lib/src/html_generator.dart
@@ -224,15 +224,15 @@ class HtmlGeneratorInstance {
           generateProperty(package, lib, clazz, property);
         });
 
-        clazz.allInstanceProperties.forEach((property) {
+        clazz.propertiesForPages.forEach((property) {
           generateProperty(package, lib, clazz, property);
         });
 
-        clazz.allInstanceMethods.forEach((method) {
+        clazz.methodsForPages.forEach((method) {
           generateMethod(package, lib, clazz, method);
         });
 
-        clazz.allOperators.forEach((operator) {
+        clazz.operatorsForPages.forEach((operator) {
           generateMethod(package, lib, clazz, operator);
         });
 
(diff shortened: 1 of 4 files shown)


devoncarew:
Looks like the checkLinks task found some issues (https://travis-ci.org/dart-lang/dartdoc/jobs/78612803#L216); nice validation of the build system :)

sethladd:
Automation FTW! :)

keertip:
Those are false positives - if you look at the generated docs, there are no links generated for anything inherited from Duration, dartdoc uses linkedName to make sure it is in the package, whereas the test is checking href.

sethladd:
First off: awesome for working on this!

Does this handle the case of ExtraSpecialList which extends SpecialList which extends ListBase ? Where would ExtraSpecialList's length property link to?  ListBase? List?

sethladd:
The link checker is opening the HTML file and finding anchor tags. Is the link checker broken (it could be)?

keertip:
Ok, the links are there in the generated pages - they are the "..." in the one liner doc comments. Was going to remove the one line doc if the element was inherited but not in package in the next PR, but will fix that now!

sethladd:
> Was going to remove the one line doc if the element was inherited but not in package

Don't we want the one-line doc for any and all names mentioned on a page?

keertip:
The user has no way to jump to the element page to see more of the docs. So at the least we should not be showing that there is more to see.

sethladd:
@keertip do you have a copy of the docs hosted somewhere that you're now generating with this patch?

sethladd:
I don't see links generated for ExtraSpecialList:

screen shot 2015-09-03 at 10 33 34 am

Can you add a test for this, as we should be generating a link to ExtraSpecialList's length property (for example)

PR #894, opened 2015-09-06
enable search when docs are opened from filesystem (file:///)

d2m:
a bit of reconfiguration enables search in file-based docs.

### DIFF
--- a/lib/src/html_generator.dart
+++ b/lib/src/html_generator.dart
@@ -185,7 +185,7 @@ class HtmlGeneratorInstance {
   }
 
   void _generateSearchIndex() {
-    File jsonFile = createOutputFile(out, 'index.json');
+    File jsonFile = createOutputFile(out, 'index.js');
     String json = JSON.encode(documentedElements.map((ModelElement e) {
       // TODO: find a better string for type
       Map data = {'name': e.name, 'href': e.href, 'type': e.kind};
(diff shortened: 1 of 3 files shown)


lib/templates/_footer.html
```diff
@@ -26,6 +26,7 @@
 <script src="https://code.jquery.com/jquery-2.1.4.min.js"></script>
 <script src="static-assets/typeahead.bundle.min.js"></script>
 <script src="static-assets/prettify.js"></script>
+<script src="index.js"></script>
```
sethladd: is there a more descriptive name that we could use?

lib/src/html_generator.dart
```diff
@@ -185,7 +185,7 @@ class HtmlGeneratorInstance {
   }
 
   void _generateSearchIndex() {
-    File jsonFile = createOutputFile(out, 'index.json');
```
sethladd: One downside is that we're preventing potential interesting uses of our docs by not publishing a JSON file. I'm tempted to continue to generate this file, in order to allow other tools to experiment.

lib/templates/_footer.html
devoncarew: We may also want to consider an `async` or a `defer` here so that loading the script doesn't block page rendering.

d2m:
Well put, i'll come up with another PR later today.

PR #899, opened 2015-09-13
Print annotations on separate lines

### LINKED ISSUE #780: Annotations should be one per line?
If I look at https://new-dartdocs-dot-dartlang-api.appspot.com/1.12.0-dev.5.1/dart-html/Element-class.html it seems odd to have

```
Annotations     @DocsEditable() @DomName('Element')
```

It seems like it would be better to have each annotation on its own line.

srawlins:
Fixes #780 

I think this looks pretty good. Two examples:

---

Element annotations

---

CDataSection

### DIFF
--- a/lib/resources/styles.css
+++ b/lib/resources/styles.css
@@ -425,10 +425,6 @@ nav .self-name {
   display: inline;
 }
 
-.annotation-list li {
-  display: inline;
-}
-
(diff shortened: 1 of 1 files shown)


devoncarew:
:+1:, this makes the annotations more readable

keertip:
Thanks Sam!

PR #900, opened 2015-09-13
Change callable metadata to be the same as class's

srawlins:
Very related to #780, but it's not 100% the same.

Currently the metadata of "callables" are listed out inline, and with commas. [Example](https://api.dartlang.org/1.12.1/dart-html/DatabaseCallback.html).

This changes the latter, and adds the `annotation-list` class, so that #899 fixes the former.

Of note, I don't understand why I found this code 2 times, in `lib/src/`, and `lib/templates/`. I fixed it in both places. *shrugs*

Example:

DatabaseCallback


keertip:
Looks good! Thanks for the change!

PR #902, opened 2015-09-18
publish 0.6.4

keertip:


sethladd:
Other than the build breaks, lgtm! Thanks for doing this :)

Issue #904, opened 2015-09-18
404 for linking from dart:io to Future


sethladd:
See https://api.dartlang.org/1.12.1/dart-io/HttpClient/getUrl.html

Specifically, see the Future as return type is linked to dart-io/Future... instead of dart-async/Future...

FWIW the page was generated by dartdoc version 0.6.2

sethladd:
BTW should we turn the link checkers on for the SDK build? It would have caught this.

cc @keertip @devoncarew

keertip:
Generated the sdk docs locally and I still see the problem.

sethladd:
Thanks for checking.

We don't run the link checker on the Core SDK docs, only docs from our test_package. Maybe we should run the link checker on the core SDK docs as generated by Travis, too.

keertip:
Yes, we should. Will look at fixing the issue and then enable the checking too.

sethladd:
Rock on!

PR #905, opened 2015-09-19
pub upgrade; readme tweaks

devoncarew:
- run pub upgrade
- tweak some docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,9 @@
 # dartdoc
 
-Use _dartdoc_ to generate HTML documentaton for your Dart package.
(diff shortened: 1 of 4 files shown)


keertip:
@devoncarew , can we wait on the pub upgrade? Would like to get #904 fixed and running on the bots. The pub upgrade might need tweaking of the dependent packages on the bots. Prefer to do that after the fix is in.

Thanks!

devoncarew:
Sure - reverted the changes to the lock file.

keertip:
Thanks!

devoncarew:
(rebased)

PR #909, opened 2015-09-21
show source for ctors

devoncarew:
- show source code for constructors
- minor updates to the readme
- make the style of the source box more consistent with source code boxes in dartdoc
- make long source code scroll instead of word-wrap; reading word-wrapped, indented source code is pretty hard


keertip:
lgtm!

PR #915, opened 2015-09-21
Sdk docs

keertip:


keertip:
closes #904 

Not calling the check sdk links as yet, cause the of the one failure in `dart-developer/log.html` which will be fixed when the comment fix gets into the sdk builds.

PR #921, opened 2015-09-22
don't show method sigs for methods with source code

devoncarew:
For methods with any associated source (everything but no-body constructors), don't show a method sig at the top of a file. Lead with the method's dartdoc, and let the `Source` section show the method and params. With the source section, we already have the method name on the page 3 times.

screen shot 2015-09-22 at 11 50 10 am


keertip:
lgtm! Like it! Page looks a lot cleaner.

Issue #813, opened 2015-08-12
Should constants use an arrow and type on the right?


sethladd:
![screen shot 2015-08-11 at 10 18 41 pm](https://cloud.githubusercontent.com/assets/5479/9217002/f8221708-4076-11e5-8a2c-1e180f9a5efd.png)

We put the type annotation on the right for properties.

cc @munificent

munificent:
I was thinking about that too. I'm all in favor of it.

seaneagan:
Agree since constants are just a special kind of property. But it might be a bit hard to read with the value literals e.g. `APRIL → int = 4`.  Maybe move the value to the next line and get rid of `const` (which is redundant since the section is already labeled "Constants")?:

```
APRIL → int
= 4
```

PR #927, opened 2015-09-27
improve the display of constants

devoncarew:
address https://github.com/dart-lang/dartdoc/issues/813

Improve the display of constants:

screen shot 2015-09-26 at 9 19 51 pm

- change to use the name on the left, type on the right notation for constants
- include the constant value on a separate line, after the `const` keyword

I played with a few other looks - this one seemed like the best result. The single line ones didn't read well (`name -> type, value`, `name -> type: value`, `name -> type (value)`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## unreleased
 * [health] remove --url-mappings option, the analyzer picks up sdk extensions
   from the package
+* [style] changed how we display constant values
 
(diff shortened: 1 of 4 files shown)


devoncarew:
also:
screen shot 2015-09-27 at 11 47 03 am

sethladd:
Thanks! Glad we kept the display of the actual value.

sethladd:
lgtm

PR #934, opened 2015-09-30
fix type info for Map with generics

### LINKED ISSUE #917: Map type is sometimes shown with only one generic argument type
for the invocation-class [0] the Map type for the named arguments states that it only has one generic argument. I'm guessing the second one is `dynamic`, but it's not clear.

[0] https://api.dartlang.org/1.12.1/dart-core/Invocation-class.html

### LINKED ISSUE #919: Missing 'dynamic' in type parameters
`Map<Symbol>` below should be `Map<Symbol, dynamic>`
Omit the type parameters only when all are dynamic.

https://api.dartlang.org/1.12.1/dart-core/Function/apply.html
```
dynamic apply( 
    Function function,
    List positionalArguments,
    [Map<Symbol> namedArguments]
)
```

keertip:
closes #919 
closes #917

### DIFF
--- a/test/model_test.dart
+++ b/test/model_test.dart
@@ -558,7 +558,7 @@ void main() {
     });
 
     test('get methods', () {
-      expect(Dog.instanceMethods, hasLength(3));
+      expect(Dog.instanceMethods, hasLength(4));
     });
 
     test('get operators', () {
@@ -605,7 +605,7 @@ void main() {
     });
 
     test('F has many inherited methods', () {
-      expect(F.inheritedMethods, hasLength(6));
+      expect(F.inheritedMethods, hasLength(7));
       expect(
           F.inheritedMethods.map((im) => im.name),
           equals([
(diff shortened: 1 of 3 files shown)


test/model_test.dart
```diff
@@ -839,6 +844,11 @@ String topLevelFunction(int param1, bool param2, Cool coolBeans,
       expect(isGreaterThan.modelType.createLinkedReturnTypeName(), 'bool');
     });
 
+    test('parameter has generics in signature', () {
+      expect(testGeneric.parameters[0].modelType.linkedName,
+          'Map&lt;String,dynamic&gt;');
```
sethladd: should we put a space after the comma ?

keertip: yes, we should. Thought I had added it.

sethladd:
If you don't mind rebasing, please do.

other, lgtm! And thanks!

Issue #937, opened 2015-10-04
abstract classes and factory constructors - qualifiers not part of docs


chalin:
It seems to me that **dartdoc** generated documentation doesn't indicate whether a class is `abstract` (e.g., Iterator), or am I missing something? I believe that the abstract qualifier was being output when dartdoc was in beta at some point. What was the rationale for removing it? It seems to me that it is useful to know whether a class is abstract or not.

Likewise, it would be useful to know (in the documentation) if a constructor is a `factory` or a generative constructor. Again, was there a rational for not indicating this in the documentation?

Cf. https://groups.google.com/a/dartlang.org/d/msg/misc/Hs0vyln6PfA/pZLUHB_FAwAJ

sethladd:
re: abstract, that makes sense. We should add that.

re: factory constructor: curious, why does that matter? From a user's perspective, they don't know if it's a factory or not.

thanks for the feedback!

chalin:
E.g., whether a superclass constructor is generative or a factory can affect how you write subclass constructors:
```dart
class P {
  P.one() {}
  factory P.two() => ...
}

class C extends P {
  C.one() : super.one(); // ok P.one() is generative
  C.two() : super.two(); // error P.two() is a factory.
}
```
Also, e.g., factory constructors can return `null`; generative constructors always return (non-null) fresh instances.

sethladd:
> whether a superclass constructor is generative or a factory can affect how you write subclass constructors

Good point!

chalin:
+1 for adding fix for `const` constructor qualifier too.

PR #944, opened 2015-10-07
show abstract in class title

keertip:
address #937, showing whether class is abstract or not.


lib/src/html_generator.dart
```diff
@@ -547,8 +547,13 @@ class ClassTemplateData extends TemplateData {
       '${clazz.name} ${clazz.kind} - ${_library.name} library - Dart API';
   String get metaDescription =>
       'API docs for the ${clazz.name} ${clazz.kind} from the ${_library.name} library, for the Dart programming language.';
+  String get _fullkind {
```
sethladd: consider for later to move this to the model, and that way we can add a test for it :)

PR #948, opened 2015-10-08
show factory for constructors

### LINKED ISSUE #937: abstract classes and factory constructors - qualifiers not part of docs

keertip:
closes #937

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## unreleased
+* [enhancement] indicate when a class is an abstract class
+* [enhancement] indicate when a constructor is a factory cosntructor
+
 ## 0.7.0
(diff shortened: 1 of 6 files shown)


sethladd:
lgtm !

Issue #946, opened 2015-10-07
please show signature at top of method docs


trinarytree:
The class doc shows method signatures like
`Map.fromIterable(iterable, {K key(element), V value(element)})`
example: https://api.dartlang.org/1.12.1/dart-core/Map-class.html
but when you click on the method, it doesn't show the signature at the top.
example: https://api.dartlang.org/1.12.1/dart-core/Map/Map.fromIterable.html
Instead it shows the source code at the bottom.
(1) I don't really want to see the source code.
I mean, it's fine if I have a link to it, but I'm really here for documentation.
Source code has all kinds of distracting garbage, like the word 'factory', an equals sign,
the implementation, not to mention worse syntax highlighting and less links.
(2) The source code is often below the fold (i.e. you can't see it unless you scroll down),
it certainly is for Map.fromIterable.  I want to see the signature prominently at the top,
and in the same format as on the class summary page, e.g.
where return types are shown like -> R.

sethladd:
cc @devoncarew 

I would agree that a formal signature at the top of the method is a useful thing.

keertip:
@devoncarew, what do you think? It does look cleaner without the signature, but we don't want to lose out on ease of use.

devoncarew:
For undoc'd methods it seemed like lots of duplication (esp. w/ the name already on the page in another place). Much of the same information was repeated 3 times in close proximity on the pafe. I agree that it would help for for methods that do have docs. Not sure if there's a compromise or if we should just bring back the method sig.

sethladd:
Just ran into this issue myself. Check out https://api.dartlang.org/134069/dart-isolate/Isolate/spawnUri.html

sethladd:
@devoncarew mind if we add it back to the top until we find something better?

devoncarew:
Makes sense re: adding it back in.

PR #956, opened 2015-10-11
display the method sig at the top of method/ctor/function pages

devoncarew:
fix https://github.com/dart-lang/dartdoc/issues/946

- restore the method sig at the top of method/ctor/function/operator pages
- style the method sig so that it's similar (but not quite as much emphasis) as the h2 section headers, like `Source`
- display using one line, similar to how it would look in a dart file. when wrapping, indent; ensure we only wrap between params, not between a param type and name

@sethladd 

screen shot 2015-10-10 at 7 20 11 pm

screen shot 2015-10-10 at 7 20 33 pm

screen shot 2015-10-10 at 7 20 48 pm

screen shot 2015-10-10 at 7 21 03 pm

screen shot 2015-10-10 at 7 21 17 pm

### DIFF
--- a/lib/resources/styles.css
+++ b/lib/resources/styles.css
@@ -145,7 +145,7 @@ pre.prettyprint {
   font-family: 'Source Code Pro', monospace;
   color: black;
   border-radius: 4px;
-  font-size: 13px;
+  font-size: 14px;
   word-wrap: normal;
   line-height: 1.4;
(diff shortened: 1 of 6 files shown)


keertip:
Thanks! looks good!

sethladd:
lgtm thanks!

PR #958, opened 2015-10-11
close a span

devoncarew:
TBR, close a span.


sethladd:
Oh, thanks for the catch!

PR #964, opened 2015-10-14
Add redirector script from the old dartdocs.org path schema

astashov:
The old schema was something like
`index.html#package/library.Class@id_method`. So, we parse it and try to
redirect to the new dartdoc's path. For that, we first make a HEAD
request to check if that location actually exists, and if yes - change
`location.href` to that new location

It doesn't always work, because there are some changes between the old
schema and the new schema, so sometimes the redirect would lead to 404.
Known cases:

* Operators. Old dartdocs.org doesn't make a distinction between
  an operator and e.g. a method, but the new dartdoc does. So, there is
  no way to know if this is an operator or not, so the redirect won't
  work.

* Inherited methods. If class B extends A, and B has inherited methods
  from A, then dartdocs.org will show them as `#package/lib.B@id_foo`, and
  dartdoc will consider it as `lib/A/foo.html`, so the redirect won't work
  either.

There may be other cases, but most of the time it should work fine.

### DIFF
--- a/lib/resources/redirector.js
+++ b/lib/resources/redirector.js
@@ -1,0 +1,85 @@
+(function () {
+    function getFromHash(hash, regexp, body) {
+        var match = hash.match(regexp);
+        var result;
+        if (match) {
+            result = body ? body(match[1]) : match[1];
+        }
+        return [result, hash.replace(regexp, "")];
+    }
+
+    function getPackage(hash) {
+        return getFromHash(hash, /^([^\/]+)\//, function (m) { return m.replace("-", "."); });
+    }
+
+    function getLibrary(hash) {
+        return getFromHash(hash, /^([^@\.]+)/, function (m) { return m.replace("-", "."); });
+    }
+
+    function getClass(hash) {
+        return getFromHash(hash, /^\.([^@]+)/);
+    }
+
+    function getFunction(hash) {
+        return getFromHash(hash, /^@id_(.+)$/);
+    }
+
+    function getMethod(hash) {
+        return getFromHash(hash, /^@id_(.+)$/, function (m) {
+            // if it contains ',', we should discard everything after that,
+            // that was probably named parameter arguments
+            // if it ends with '-', it may be a constructor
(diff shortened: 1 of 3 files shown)


sethladd:
Thanks! Just to double check, are we sure this should live in dartdoc? Is this specific to dartdocs.org ?

astashov:
I'd love to, but I don't know how to make it work without modifying dartdoc, frankly. (I mean - we could use a proxy server, which parses HTML and injects that script into it, and then send all the traffic to this proxy server, but IMHO it's fragile, complicated, and may break with further dartdoc changes)

I need that script on the dartdoc's generated html pages. So people would go to something like http://www.dartdocs2.com/documentation/stream/1.5.0/index.html#stream/stream.Browser@id_android, and that should redirect to http://www.dartdocs2.com/documentation/stream/1.5.0/stream/Browser/android.html.

I think it kind of makes sense, it's a feature in dartdoc to provide backwards-compatibility with the previous Dart documentation generator.

I'm open to other suggestions, but for now I don't really see how this could be done differently.

astashov:
> dartdoc supports the ability to inject content into the <head> of a document. dartdocs.org could use this to inject custom scripts.

Ah, interesting. Didn't know that. I'll look into that.

sethladd:
This might also be really useful for api.dartlang.org! Perhaps we do need to include it in dartdoc directly (if we did, we'd probably want to put it into script.js to eliminate a download)

astashov:
Is there any way to access old api.dartlang.org? The way it works with SDK may differ from the pub packages'.

sethladd:
> Is there any way to access old api.dartlang.org?

Hm, we have the old code here: https://github.com/dart-lang/api.dartlang.org/tree/c986d54b2b933ca5db2234817a80154e2f5f86e3

However, that was using the old dartdoc-viewer, so hopefully it's the same kind of URLs and hashing structure.

astashov:
I think I'll just add that in dartdocs.org for now only.

PR #965, opened 2015-10-15
Fix incorrect 'Link to Crossdart' links

astashov:
They were generated incorrectly (whoops!)
Fixed that.

### DIFF
--- a/lib/src/model.dart
+++ b/lib/src/model.dart
@@ -22,6 +22,7 @@ import 'package_meta.dart' show PackageMeta, FileContents;
 import 'cache.dart';
 
 import '../markdown_processor.dart' show Documentation, renderMarkdownToHtml;
+import 'package:analyzer/src/generated/source_io.dart';
 
 int byName(Nameable a, Nameable b) =>
     a.name.toUpperCase().compareTo(b.name.toUpperCase());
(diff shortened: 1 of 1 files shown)


devoncarew:
@keertip @sethladd

sethladd:
Looks like tests are failing because the version of analyzer is still using dart:profiler.

```
The built-in library 'dart:profiler' is not available on the stand-alone VM.
'package:analyzer/src/generated/utilities_general.dart': error: line 10 pos 1: library handler failed
import 'dart:profiler';
^
```

sethladd:
The tests on stable are passing, so I think we can merge this in.

LGTM

lib/src/model.dart
```diff
@@ -22,6 +22,7 @@ import 'package_meta.dart' show PackageMeta, FileContents;
 import 'cache.dart';
 
 import '../markdown_processor.dart' show Documentation, renderMarkdownToHtml;
+import 'package:analyzer/src/generated/source_io.dart';
```
sethladd: We shouldn't import a file from src. Is this absolutely necessary?

keertip: We can make an exception for the analyzer package. It has public api in `src`.

sethladd: hah. well, ok then :)

sethladd:
deferring review to @keertip

sethladd:
Once we address https://github.com/dart-lang/dartdoc/issues/972 then this PR should pass its Travis builds.

keertip:
restarting build

keertip:
lgtm!

PR #967, opened 2015-10-17
Fix #966 by not stripping comments twice

### LINKED ISSUE #966: Indented code blocks don't seem to be respected
See [ListMixin.replaceRange](https://api.dartlang.org/133561/dart-collection/ListMixin/replaceRange.html). The line starting with `List<int> list` should be a code block, but is just a paragraph. The [source](https://github.com/dart-lang/sdk/blob/master/sdk/lib/core/list.dart#L446) for this comment is:

```dart
/**
 * Removes the objects in the range [start] inclusive to [end] exclusive
 * and inserts the contents of [replacement] in its place.
 *
 *     List<int> list = [1, 2, 3, 4, 5];
 *     list.replaceRange(1, 4, [6, 7]);
 *     list.join(', '); // '1, 6, 7, 5'
 *
 * An error occurs if [start]..[end] is not a valid range for `this`.                                                                                       *
 * This method does not work on fixed-length lists, even when [replacement]
 * has the same number of elements as the replaced range. In that case use
 * [setRange] instead.
 */
```

Which seems properly formatted.

Another example is [ListMixin.asMap](https://api.dartlang.org/133561/dart-collection/ListMixin/asMap.html).

srawlins:


sethladd:
Can you please add a test so we catch this regression next time? Thanks!

srawlins:
Thanks for looking, @sethladd , I couldn't really find tests in this package asserting on comment content. Can you point me to where I should add a test case?

sethladd:
Maybe model_test.dart ?

srawlins:
Regression test added. Thanks, @sethladd !

sethladd:
Fantastic!

lgtm

PR #969, opened 2015-10-18
Move html_utils.dart to utils.dart as per TODO; add tests

srawlins:


sethladd:
Beautiful! and thanks :)

PR #970, opened 2015-10-18
Correct getFileNameFor documentation

srawlins:


sethladd:
wow, thanks and good catch!

PR #971, opened 2015-10-18
ignore pubspec.lock

kevmoo:
Different versions of the SDK cause different versions of the analyzer to be
installed. This breaks tests unless 'pub get' starts fresh on the CI server

[Review on Reviewable](https://reviewable.io/reviews/dart-lang/dartdoc/971)


kevmoo:
CC @sethladd 

@srawlins I think your CLs will pass on travis if you rebase on this – although we should wait to land this before we commit your patches

sethladd:
dartdoc is an application. Shouldn't it check in its pubspec.lock ?

kevmoo:
Addressed by https://github.com/dart-lang/dartdoc/pull/973

PR #984, opened 2015-10-21
tweak stdout message; fix html entity ref

devoncarew:
- tweak the message when we don't find library level docs; it was printing out one too many newlines
- fix an issue with an html entity

@keertip


keertip:
lgtm

kevmoo:
@devoncarew Does this change affect the output HTML? Can you show me an example?

If so, it might be good to document it in the changelog

devoncarew:
I does affect the html output; browsers will render it correctly either way however. This was more a correctness fix then a user-visible one.

Issue #641, opened 2015-07-13
Option: hide inherited fields / methods


sethladd:
Users have asked to be able to hide inherited fields / methods / "things".

Often APIs are full of methods pulled in from super classes, but the user just wants to see "what's new". For example, HTML and Sky and other widget libraries.

sethladd:
cc @johnmccutchan @abarth for FYI

sethladd:
Dupes https://github.com/dart-lang/dartdoc/issues/980

Hixie:
Bonus points if you can also show or hide those that are overriding something from the base class.

sethladd:
Addressed by https://github.com/dart-lang/dartdoc/pull/1010

sethladd:
Not sure we closed this one. We just dim inherited, yes? We'd still like to add an option to hide them.

kevmoo:
Gotcha

Hixie:
It doesn't seem like we even dim them at this point. Maybe we're on some old version?

I'd like to see just the newly introduced members e.g. in:
http://docs.flutter.io/flutter/rendering/RenderParagraph-class.html

kevmoo:
@Hixie We only dim/italic on the right side. In the main body, you see the *inherited* comment

Hixie:
I don't think the "inherited" comment is enough to make this clear to readers. I think if we want to have the members always there, maybe separating inherited members from new members, e.g. by having the inherited ones in a separate later section, would really help.

Consider a page like: http://docs.flutter.io/flutter/rendering/RenderCustomMultiChildLayoutBox-class.html

This is pretty typical for Flutter docs.

BTW, it would also be useful to be able to distinguish methods that were overridden from those that are entirely new. Similarly marking the protected methods somehow would be good (we don't have any yet but we will soon be going and adding @protected annotations all over the place).

kevmoo:
@Hixie we might-could open a new issue on this.

I went back and forth w/ @sethladd and @keertip on what's in there now.

Separate feature requests for overrides and projected would also be good, me thinks.

sethladd:
I'd like to see if we can find a way to make inherited methods easier to recognize, when looking at the list of methods provided by a class. There's probably a better way to do it than to print "inherited" below the name.

![screen shot 2016-03-18 at 1 20 40 pm](https://cloud.githubusercontent.com/assets/5479/13891079/436cadd0-ed0c-11e5-99bd-dbb09e3b406a.png)

[...]

PR #995, opened 2015-10-27
Misc cleanup

kevmoo:
[Review on Reviewable](https://reviewable.io/reviews/dart-lang/dartdoc/995)

### DIFF
--- a/lib/src/html_generator.dart
+++ b/lib/src/html_generator.dart
@@ -476,7 +476,7 @@ abstract class TemplateData {
         (element as SourceCodeMixin).hasSourceCode) {
       return [new Subnav('Source', '${element.href}#source')];
     } else {
-      return [];
+      return const [];
     }
   }
 }
--- a/lib/src/html_generator.dart
+++ b/lib/src/html_generator.dart
@@ -299,8 +299,8 @@ class _HtmlGeneratorInstance implements HtmlOptions {
         .write('\ngenerating docs for library ${lib.name} from ${lib.path}...');
 
     if (!lib.isAnonymous && !lib.hasDocumentation) {
-      stdout.write(
-          "\n  warning: '${lib.name}' has no library level documentation comments");
+      stdout.write("\n  warning: '${lib.name}' has no library level "
+          "documentation comments");
     }
 
     TemplateData data = new LibraryTemplateData(this, package, lib);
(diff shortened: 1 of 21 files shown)


kevmoo:
@keertip PTAL

lib/src/html_generator.dart
```diff
@@ -221,61 +222,61 @@ class _HtmlGeneratorInstance implements HtmlOptions {
 
     generatePackage();
 
-    package.libraries.forEach((Library lib) {
+    for (var lib in package.libraries) {
```
keertip: What is the reason for this change?

keertip:
@sethladd, could you take a look at the template changes?

lib/src/html_generator.dart
kevmoo: By using for loops, we avoid creating extra closures, which add to call stacks.

There's also nothing to step into when debugging – just a good practice to encourage and model for our users.

kevmoo:
@keertip @sethladd The template changes were just whitespace – and then adding the class attributes.

Would love a 2nd/3rd pair of eyes on em, though

lib/templates/_callable.html
```diff
@@ -1,11 +1,11 @@
-<dt id="{{htmlId}}" class="callable">
-    <span class="name {{#isDeprecated}}deprecated{{/isDeprecated}}">{{{linkedName}}}</span><span class="signature">(<wbr>{{{ linkedParamsNoMetadata }}})
-        <span class="returntype parameter">&#8594; {{{ linkedReturnType }}}</span>
-    </span>
+<dt id="{{htmlId}}" class="callable{{ #isInherited }} inherited{{ /isInherited}}">
```
sethladd: was this intentional to check for inherited here?

lib/templates/_property.html
```diff
@@ -1,8 +1,8 @@
-<dt id="{{htmlId}}" class="property">
-    <span class="name">{{{linkedName}}}</span>
-    <span class="signature">&#8594; {{{ linkedReturnType }}}</span>
+<dt id="{{htmlId}}" class="property{{ #isInherited }} inherited{{ /isInherited}}">
```
sethladd: again, checking this is intentional

sethladd:
lgtm thanks!

lib/src/html_generator.dart
keertip: got it.

keertip:
lgtm. Thanks for working on #641

lib/templates/_property.html
kevmoo: ditto

lib/templates/_callable.html
kevmoo: Yup. May want to dim/hide both the `dd` and `dt`.

PR #996, opened 2015-10-27
fix property page

### LINKED ISSUE #989: Fields with comments show up – weird
https://api.dartlang.org/1.13.0-dev.7.2/dart-io/HttpServer/autoCompress.html

It's defined as a field

```
   /**
   * Whether the [HttpServer] should compress the content, if possible.
   *
   * The content can only be compressed when the response is using
   * chunked Transfer-Encoding and the incoming request has `gzip`
   * as an accepted encoding in the Accept-Encoding header.
   *
   * The default value is `false` (compression disabled).
   * To enable, set `autoCompress` to `true`.
   */
  bool autoCompress;
```

It shows up as an undocumented field plus a weird operator-looking thing on `0.7.3`

screen shot 2015-10-22 at 12 48 27 pm

Clearly a regression https://api.dartlang.org/1.12.2/dart-io/HttpServer/autoCompress.html

keertip:
closes #989

### DIFF
--- a/lib/src/model.dart
+++ b/lib/src/model.dart
@@ -1519,6 +1519,12 @@ class Field extends ModelElement
 
   bool get readWrite => hasGetter && hasSetter;
 
+  bool get hasExplicitSetter => hasSetter && !_setter.isSynthetic;
+
+  bool get hasExplicitGetter => hasGetter && !_getter.isSynthetic;
+
+  bool get hasNoGetterSetter => !hasExplicitGetter && !hasExplicitSetter;
+
   String get typeName => "property";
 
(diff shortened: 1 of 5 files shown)


kevmoo:
@keertip Any way we can add a regression test here to make sure this doesn't break again?

keertip:
@kevmoo, we could add a grinder test to look for file/contents in the docs for sdk, but this could break if sdk changes.

kevmoo:
Could we run the generator over the sample library in the test directory and save it – then validate that the output doesn't change – or just regen the expected output on big changes.

Thoughts?

PR #1001, opened 2015-10-29
Test generated HTML output for unexpected changes

kevmoo:


keertip:
@kevmoo , looks like the test is failing.

kevmoo:
@keertip Got this fixed. It may add a tiny bit of noise to future updates – but I think it'll help us find bugs

keertip:
lgtm! Guess will have to do check ins of docs when we change output or fix bugs.

PR #1010, opened 2015-11-02
[WIP] show dim hide inherited

kevmoo:
http://kevmoo.github.io/dartdoc/dart-io/HttpServer-class.html

I want to make the radio buttons look better – open to suggestions

CC @keertip @devoncarew @sethladd

[Review on Reviewable](https://reviewable.io/reviews/dart-lang/dartdoc/1010)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+<<<<<<< HEAD
 ## 0.8.0
 * [bug] fix annotation shown as raw HTML in constructors
 * [bug] fix missing return type when Future
@@ -10,6 +11,7 @@
   Use of these libraries by third-party code is no longer supported.
 * **BREAKING** `generator.dart` is no longer exposed as a stand-alone library.
   You can access the `Generator` class by importing `dartdoc.dart`.
+* [enhancement] Add UI to show, hide, or dim inherited class members.
 
 ## 0.7.4
 * [bug] In class documentation, move constructors before instance properties
--- a/lib/resources/script.js
+++ b/lib/resources/script.js
@@ -194,11 +194,100 @@ function initSearch() {
   jsonReq.send();
 }
 
+// copied from https://developer.mozilla.org/en-US/docs/Web/API/document/cookie#A_little_framework_a_complete_cookies_readerwriter_with_full_unicode_support
+var docCookies = {
+  getItem: function (sKey) {
+    if (!sKey) { return null; }
+    return decodeURIComponent(document.cookie.replace(new RegExp("(?:(?:^|.*;)\\s*" + encodeURIComponent(sKey).replace(/[\-\.\+\*]/g, "\\$&") + "\\s*\\=\\s*([^;]*).*$)|^.*$"), "$1")) || null;
+  },
+  setItem: function (sKey, sValue, vEnd, sPath, sDomain, bSecure) {
+    if (!sKey || /^(?:expires|max\-age|path|domain|secure)$/i.test(sKey)) { return false; }
+    var sExpires = "";
+    if (vEnd) {
+      switch (vEnd.constructor) {
+        case Number:
+          sExpires = vEnd === Infinity ? "; expires=Fri, 31 Dec 9999 23:59:59 GMT" : "; max-age=" + vEnd;
+          break;
+        case String:
+          sExpires = "; expires=" + vEnd;
+          break;
+        case Date:
+          sExpires = "; expires=" + vEnd.toUTCString();
+          break;
+      }
+    }
+    document.cookie = encodeURIComponent(sKey) + "=" + encodeURIComponent(sValue) + sExpires + (sDomain ? "; domain=" + sDomain : "") + (sPath ? "; path=" + sPath : "") + (bSecure ? "; secure" : "");
+    return true;
+  },
+  removeItem: function (sKey, sPath, sDomain) {
+    if (!this.hasItem(sKey)) { return false; }
+    document.cookie = encodeURIComponent(sKey) + "=; expires=Thu, 01 Jan 1970 00:00:00 GMT" + (sDomain ? "; domain=" + sDomain : "") + (sPath ? "; path=" + sPath : "");
+    return true;
+  },
+  hasItem: function (sKey) {
+    if (!sKey) { return false; }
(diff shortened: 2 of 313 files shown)


kevmoo:
...and @hixie

Hixie:
Neat!

devoncarew:
I'd love to avoid having config options in the generated docs if possible. I think the goal here is to make people aware of the inherited members, but not give them the same weight on the page as the declared members?

We could try listing inherited members at the beginning of each section. This would be like what we do at the beginning of the class page, listing `implements` and `implemented by`. So, a `methods` section might look like:

## Methods

Inherited: `Object.toString`, `Foo.bar`

`deleteSync({bool recursive: false}) → void`
Synchronously deletes this FileSystemEntity.…

`exists() → Future<bool>`
Checks whether the file system entity with this path exists.

--

With the inherited items being hyperlinked where possible.


If we do choose to have configuration settings in the generated pages, I think we'd want to be more terse that three radio buttons. Probably a single check box: `[ ] show inherited`.

kevmoo:
@devoncarew What's wrong with config?

MSDN uses cookies to store which language you want to see APIs in – https://msdn.microsoft.com/en-us/library/bb359438(v=vs.110).aspx

This is very similar. Sometimes you want to see everything, some times you only want to see "what's new".

We could inject the HTML radio buttons w/ JS if you're worried about the size of the markup.

Hixie:
There's definitely times where the inherited members are important.

sethladd:
I think we should allow people to easily hide all inherited methods. Some suggestions:

* remove "dim" -- less options are better
* use a material design drop-down instead of putting these radio buttons directly into the page (it doesn't look great, and it won't scale)

keertip:
Would remove the "dim" option, not necessary. Just a show/hide for the inherited elements is enough.

Agree with @sethladd , the radio buttons are not quite the thing there.

devoncarew:
@kevmoo, re: config:
- less configuration ==> a simpler UI
- it takes me away from a web page feel
- I'm worried that we're adding on option here instead of noodling on one good presentation that will generally be what people want to use. As in, an option in lieu of determining the best presentation.

I still like my idea of always showing in inherited members but using a shorthand list style where they don't take up much space, and look different from members explicitly declared in the class.

Looks like https://github.com/dart-lang/dartdoc/commit/c0a5547cbf3c2eb1e169f8e8e20f69125e6422f2 is a bug fix that we'll want in regardless of the how this PR plays out.

sethladd:
We've had numerous requests to hide inherited members. It's really useful for frameworks like HTML or Flutter... so many widgets, so many inherited members.

I'm sensitive to turning dartdoc into a configurable app... I think it's a good goal to avoid that. However, in this case, I'm open to exploring a UI that toggles inherited.

Hixie:
There are definitely cases where what you want is the members of at least the bottom two or three classes in the hierarchy.

For example, we have a few classes where there's one abstract class that defines all the APIs, and then two subclasses for, e.g., horizontal and vertical. If we hid all the members, or just listed them without details, then the help page for those leaf classes would be useless.

Actually you almost want a slider that lets you decide how deep in the hierarchy to go when showing inherited members, rather than an all or nothing radio button. ;-)

devoncarew:
Hmm, is the noise just largely the `Object` inherited methods?

sethladd:
> Hmm, is the noise just largely the Object inherited methods?

No, it's much more than that. Consider HTML's Button class. Or any widget in a complex widget hierarchy. So. Many. Inherited. Members.

kevmoo:
Often, yes, but my example – http://kevmoo.github.io/dartdoc/dart-io/HttpServer-class.html – shows how this comes up w/ other types.

Things that inherit from HtmlElement like [FormElement](http://kevmoo.github.io/dartdoc/dart-html/FormElement-class.html) are examples

Hixie:
Yeah. The Flutter docs are a good example:
https://domokit.github.io/docs/sky/rendering/RenderBlock-class.html
That class has almost no non-inherited members. Each superclass (there are 5) introduces some interesting members, but which you care about at any moment in time varies based on whether you're trying to learn what's special about RenderBlock (no inherited members), how to _use_ RenderBlock as opposed to any other RenderBox (just the members from it and its parent), what APIs you might expect to reasonably use with RenderBlock (members from it, its parent, and grandparent), what rendering-level APIs are present on this class (everything up to RenderObject), what non-Dart APIs are on this class (everything up to AbstractNode), or everything that's on this class (everything up to Object).

kevmoo:
This should be pulled after https://github.com/dart-lang/dartdoc/pull/1013

Here's the updated output: https://kevmoo.github.io/dartdoc/dart-io/HttpServer-class.html

I'm making another plea for the "dim" option – I think it's *really* nice. It has a great flow when skimming the right side of the docs.

Are folks still strongly against this?

sethladd:
I strongly prefer a material design drop-down. Something like:

screen shot 2015-11-02 at 12 55 00 pm

Less config options is always better. Another way to ask is, what happens if we _don't_ include the Dim option?

Regardless, thanks for cranking on this.

kevmoo:
@sethladd 

RE style: I like material design, too – but the site is all bootstrap.

I'd argue against pulling in another pile of CSS/JS just for this dropdown. If we want to migrate the site to material, I'm all for that – but it should be separate item.

RE the option. I'm claiming the "dim" option is critical. I actually like it **more** than hide. It makes it easy to see what is inherited vs not while scanning the members on the right.

sethladd:
> I'd argue against pulling in another pile of CSS/JS just for this dropdown.

I'm not suggesting that. There's probably a way to write just enough css and JS to make it work. Definitely don't want to pull down an entirely new CSS framework for one widget :)

The current implementation doesn't look like it fits with the rest of the site (built-in widgets almost never do :( and it's not mobile-friendly:

screen shot 2015-11-02 at 1 05 34 pm

kevmoo:
Okay – another attempt at this. Took some inspiration from http://materializecss.com/forms.html and http://codepen.io/pudgereyem/pen/PqBxQx

see the updated bits here: https://kevmoo.github.io/dartdoc/dart-io/HttpServer-class.html

sethladd:
Thanks for the update!

looks like we need more padding between the lower border of the drop-down and the breadcrumbs, while in mobile.

Also, looks like the drop-down overlaps with the search box, in desktop, when you open it up.

Is Roboto used for the font inside of the drop-down, when opened up?

kevmoo:
Holding off on this. I tried to bring in MDL and ended up with a lot more CSS/JS loaded as part of the app.

Should consider other options...

PR #1015, opened 2015-11-02
fix return type not showing if Future

### LINKED ISSUE #1008: Missing return type in IOSink close, flush
Clearly a `Future`

screen shot 2015-10-29 at 1 41 55 pm
screen shot 2015-10-29 at 1 42 01 pm

keertip:
Closes #1008

### DIFF
--- a/lib/src/model.dart
+++ b/lib/src/model.dart
@@ -2032,7 +2032,7 @@ class ElementType {
     // not TypeParameterType or Void or Union type
     if (isParameterizedType) {
       if (typeArguments.every((t) => t.linkedName == 'dynamic')) {
-        _linkedName = '';
+        _linkedName = buf.toString();
         return _linkedName;
       }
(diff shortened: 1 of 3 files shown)


lib/src/model.dart
```diff
@@ -2032,7 +2032,7 @@ class ElementType {
     // not TypeParameterType or Void or Union type
     if (isParameterizedType) {
       if (typeArguments.every((t) => t.linkedName == 'dynamic')) {
-        _linkedName = '';
+        _linkedName = buf.toString();
```
kevmoo: why not `return buf.toString();` ?

keertip: Storing it so that subsequent calls can be faster.

kevmoo: Ah! It's an instance var! Sorry. :grimacing:

PR #1020, opened 2015-11-03
use a span for "kind" within the page h1

kevmoo:
div should not be used within an "h" tag

[Review on Reviewable](https://reviewable.io/reviews/dart-lang/dartdoc/1020)

### DIFF
--- a/lib/src/html_generator.dart
+++ b/lib/src/html_generator.dart
@@ -466,7 +466,7 @@ abstract class TemplateData {
   String _layoutTitle(String name, String kind, bool isDeprecated) {
     if (kind.isEmpty) kind =
         '&nbsp;'; // Ugly. fixes https://github.com/dart-lang/dartdoc/issues/695
-    String str = '<div class="kind">$kind</div>';
+    String str = '<span class="kind">$kind</span>';
     if (!isDeprecated) return '${str} ${name}';
     return '${str} <span class="deprecated">$name</span>';
   }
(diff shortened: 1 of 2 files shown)


sethladd:
Nice, thanks!

LGTM

devoncarew:
> div should not be used within an "h" tag

Good to know. Is this an html best practices, or an html spec thing?

lib/src/html_generator.dart
```diff
@@ -466,7 +466,7 @@ abstract class TemplateData {
   String _layoutTitle(String name, String kind, bool isDeprecated) {
     if (kind.isEmpty) kind =
         '&nbsp;'; // Ugly. fixes https://github.com/dart-lang/dartdoc/issues/695
-    String str = '<div class="kind">$kind</div>';
+    String str = '<span class="kind">$kind</span>';
```
devoncarew: Not part of your changes, but can you add braces around the if statement body above? Or move the comment one line above. It's formatting oddly.

kevmoo:
Both.

Blocks can contain blocks and flow

Flow should only contain flow

lib/src/html_generator.dart
kevmoo: Already committed, although I agree. Do a quick pull request on this?

devoncarew: That's what the github UI is for :)

PR #1028, opened 2015-11-12
Don't include <base> at the root of a project

kevmoo:


sethladd:
I love that you have a test set of docs to compare to.

kevmoo:
Yes! Hence why we should land #1001

keertip:
lgtm!

Issue #1032, opened 2015-11-18
Deprecated libraries are crossed out in package view, but not library view


kevmoo:
If you look at a package view, the deprecated library shows up with a line through it

![screen shot 2015-11-18 at 10 33 02 am](https://cloud.githubusercontent.com/assets/17034/11250112/c09f76b6-8ddf-11e5-87e8-01ed702270d8.png)

But not when looking at another package

![screen shot 2015-11-18 at 10 33 10 am](https://cloud.githubusercontent.com/assets/17034/11250117/c5b7c072-8ddf-11e5-92ca-adfff0947b6d.png)

PR #1033, opened 2015-11-18
Show libraries the same way in both package and library view

kevmoo:
Closes https://github.com/dart-lang/dartdoc/issues/1032


keertip:
lgtm

PR #1040, opened 2015-11-23
fix for exception seen in be builds

### LINKED ISSUE #1039: Latest Dartdoc release has a bad type assertion
The problematic line: https://github.com/dart-lang/dartdoc/blob/master/lib/src/model.dart#L415
Log from Angular CI build showing the problem: https://travis-ci.org/angular/angular/jobs/92373133#L843

keertip:
The exception is due to changes in the analyzer. 
closes #1039


keertip:
Travis dev build fails due to issue with the test package https://github.com/dart-lang/test/issues/360.

Issue #1035, opened 2015-11-19
I find the source code too prominent


alan-knight:
e.g. https://api.dartlang.org/1.13.0/dart-html/History/back.html 

the thing that catches your eye is a bunch of internal implementation details of Dartium. The actual function definition is in a lighter color, not at all prominent. It's also oddly indented relative to the annotations.

PR #1042, opened 2015-11-25
Add option to exclude source in output

kevmoo:


kevmoo:
FYI @alan-knight 

I think we should set this as the default for the SDK docs, too.

keertip:
@kevmoo , so show no source for SDK docs?

keertip:
lgtm

kevmoo:
@keertip this patch doesn't affect the behavior of --sdk-docs at all – but we might want to consider changing the script on the bots

keertip:
@kevmoo , know this patch does not affect the behavior of other flags. Wanted to know the reason why you think we should not show the source for sdk docs. From #1035, it looks like there are cases where it does not offer any value, but does this apply to most of the source in the sdk?

kevmoo:
Often it's not helpful -> https://api.dartlang.org/1.13.0/dart-convert/Converter/fuse.html

In this case it's a bit nice – https://api.dartlang.org/1.13.0/dart-core/Object/operator_equals.html – but in general it seems like more noise than value.

alan-knight:
I find that with the current layout it really grabs the eye, and the method signature, which is most often what I'd want to know, is hard to see. I can get it from the source, which works for me, but it's more work to parse it out, and it's probably not great for people who aren't used to reading the SDK sources.

PR #1047, opened 2015-12-02
Update style for inherited members: dimmed and hidden

kevmoo:
Adding UI for this gets very expensive if you want to use MDL or similar
Just changing the style for now – should revisit adding dynamic feature
later

Closes https://github.com/dart-lang/dartdoc/issues/641

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## 0.8.3
 * [enhancement] Added `--[no-]include-source` option.
+* [enhancement] Dimmed inherited members in the right sidebar.
 
 ## 0.8.2
 * [bug] fix exception due to change in analyzer, function types in parameters are no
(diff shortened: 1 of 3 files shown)


kevmoo:
@sethladd @devoncarew @keertip Would love your feedback

Sample: http://kevmoo.github.io/dartdoc/dart-html/ButtonElement-class.html

kevmoo:
A thought: we could apply this to *just* the links on the right of the page. It eases quick scanning, which is usually what I want here...

Scorpiion:
Hi, I like this, I'm thinking if it would be beneficial to add a hover effect showing "Inherited member" for those dimmed links as well? Or would that be too much?

sethladd:
lgtm

keertip:
@kevmoo , looks good. Like your idea of applying this just to the right nav bar.

PR #1049, opened 2015-12-04
prepare v0.8.3

kevmoo:


keertip:
lgtm! Thanks :+1:

test_package_docs/anonymous_library/anonymous_library-library.html
```diff
@@ -4,7 +4,7 @@
   <meta charset="utf-8">
   <meta http-equiv="X-UA-Compatible" content="IE=edge">
   <meta name="viewport" content="width=device-width, initial-scale=1">
-  <meta name="generator" content="made with love by dartdoc 0.8.3-dev">
+  <meta name="generator" content="made with love by dartdoc 0.8.3">
```
devoncarew: We should remove the version here to minimize diffs.

kevmoo: In all output files?

Or maybe only include it in `index.html`?

devoncarew: I think we should remove it from the leaf node files. Perhaps from index.html too, just for uniformity?

kevmoo: So just don't include it at all?

I think @sethladd is a fan

I kinda like it to – it's nice to be able to understand issues if they arise

devoncarew: Just to be clear, I meant `made with love by dartdoc 0.8.3` ==> `made with love by dartdoc` - just removing the version part. It introduces diffs into files that are otherwise identical, like generating a timestamp into files.

PR #1051, opened 2015-12-04
Improvements

kevmoo:


kevmoo:
@sethladd @devoncarew FYI – generator data is only in the root `index.html` page – this keeps the info around, which is good for debugging – while not churning *every* page when you upgrade to a new version of dartdoc

devoncarew:
Let's keep `content="made with love by dartdoc` for the leaf files, just not have the dartdoc version info.

The size of the diffs here is killing the github review UI :( I like committing some generated output - to see concretely what changes a PR produces - but I think the current sample is too large. I'll open an issue.

kevmoo:
> Let's keep content="made with love by dartdoc for the leaf files, just not have the dartdoc version info.

The matrix we create here is a bit nuts. I really think it's fine to just have it on the main page – but open for discussion. Otherwise I think we should add a separate meta tag that has just the version.

> The size of the diffs here is killing the github review UI :( I like committing some generated output - to see concretely what changes a PR produces - but I think the current sample is too large. I'll open an issue.

Understood. Hopefully this will be the last big change that affects every file. If you look commit-by-commit, it's pretty easy to parse through each.

kevmoo:
@keertip PTAL

keertip:
lgtm

PR #1060, opened 2015-12-11
check for the <nodoc> tag in comments to generate docs for an element

### LINKED ISSUE #642: New option: hide fields/methods/names marked with a specific annotation
User has asked for a new feature to be able to hide names marked with a certain annotation. This annotation would be provided by the user.

Use case: some API designs are useful only to implementors of a framework. Other APIs are useful for consumers and developers. When an end-developer wants to use an API, they shouldn't see APIs designated for framework implementors.

Using this new feature, SDKs and toolkits could generate two sets of docs (one for implementors, and one for end-developers)

keertip:
closes #642

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 0.8.5
+* [enhancement] do not document if there is a <nodoc> in the doc comment.
+
 ## 0.8.4
 * [enhancement] Only include generator metadata in the package `index.html` file.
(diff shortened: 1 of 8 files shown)


sethladd:
cc @abarth for FYI

abarth:
Thanks!  Does this hide the declaration entirely?

keertip:
@abarth Yes, that is the plan.

PR #1062, opened 2015-12-15
Switch header/footer description

jsayol:
The descriptions for the `--header` and `--footer` options were reversed.


sethladd:
Thanks! and good catch :)

PR #1065, opened 2015-12-19
[WIP] kevin hacks for perf

kevmoo:
Caching `isPublic` about doubles the speed of generating SDK docs

### DIFF
--- a/lib/src/model_utils.dart
+++ b/lib/src/model_utils.dart
@@ -47,26 +47,42 @@ bool isInExportedLibraries(
 
 bool isPrivate(Element e) => e.name.startsWith('_');
 
+final _privacyExpando = new Expando<bool>();
+
 bool isPublic(Element e) {
-  if (isPrivate(e)) return false;
-  // check to see if element is part of the public api, that is it does not
-  // have a '#nodoc' in the documentation comment
-  if (e is PropertyAccessorElement && e.isSynthetic) {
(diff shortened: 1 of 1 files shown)


lib/src/model_utils.dart
```diff
@@ -47,26 +47,42 @@ bool isInExportedLibraries(
...
-  // have a '#nodoc' in the documentation comment
-  if (e is PropertyAccessorElement && e.isSynthetic) {
-    var accessor = (e as PropertyAccessorElement);
-    if (accessor.correspondingSetter != null &&
-        !accessor.correspondingSetter.isSynthetic) {
-      e = accessor.correspondingSetter;
-    } else if (accessor.correspondingGetter != null &&
-        !accessor.correspondingGetter.isSynthetic) {
-      e = accessor.correspondingGetter;
+  var cachedVal = _privacyExpando[e];
```
rakudrama: This lazy-init pattern can written

bool isPublic(e) => _privacyExpando[e] ??= _isPublic(e);

where `_isPublic` is the old definition of `isPublic`.

kevmoo: Nice!

keertip:
@kevmoo , thanks! Overlooked that dartdoc was calling `isPublic` more than once.

lgtm!

kevmoo:
@keertip it seems that updating deps takes care of most of the perf issues – although we might reconsider this later...

PR #1066, opened 2015-12-19
updated dependencies

kevmoo:
causes dramatic speed increase with analyzer 0.27.1+1

...like 3x faster. Nice! CC @pq

### DIFF
--- a/pubspec.lock
+++ b/pubspec.lock
@@ -4,7 +4,7 @@ packages:
   analyzer:
     description: analyzer
     source: hosted
-    version: "0.27.0"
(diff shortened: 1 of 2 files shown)


kevmoo:
...if you use `--no-include-source`

keertip:
yes, thanks to caching doc comments in the analyzer. Thanks @pq !

@kevmoo , so  #1065  is no longer needed?

lgtm!

kevmoo:
Correct!

pq:
> ...like 3x faster. Nice! CC @pq

:+1:

PR #1093, opened 2016-01-27
add support for _embedder.yaml file

### LINKED ISSUE #1067: Find libraries listed in _embedder.yaml for doc generation
The Flutter engine (or sky_engine) package will provide dart: libraries, listed in the _embedder.yaml.  (See also https://github.com/dart-lang/sdk/issues/25011)

It would be really great if the dartdoc tool could find those libraries and generate docs for them, just like any other lib in the lib/ directory.

This is (probably?) blocking https://github.com/flutter/flutter/issues/998

Thanks!

keertip:
closes #1067 

@pq , currently using an alpha version of the analyzer. Will wait till a new version is released to land this.


lib/dartdoc.dart
```diff
@@ -128,7 +129,7 @@ class DartDoc {
   List<LibraryElement> _parseLibraries(List<String> files) {
     Set<LibraryElement> libraries = new Set();
     DartSdk sdk = new DirectoryBasedDartSdk(new JavaFile(sdkDir.path));
```
pq: You might consider moving SDK creation into the else block in case there is embedder information and it isn't needed?

lib/dartdoc.dart
```diff
@@ -144,6 +145,18 @@ class DartDoc {
     }
     resolvers.add(new FileUriResolver());
 
+    EmbedderUriResolver embedderUriResolver = new EmbedderUriResolver(
+        new EmbedderYamlLocator(packageMap).embedderYamls);
```
pq: Might packageMap be `null`?

pq:
LGTM % nits.

lib/dartdoc.dart
keertip: Yes, it can be. Will fix.

lib/dartdoc.dart
keertip: It is used otherwise too.

pq: If that's the case, we need an `EmbedderSdk` in case there's embedded SDK info...

Maybe better to discuss in person?

test_package_embedder_yaml/lib/_embedder.yaml
```diff
@@ -0,0 +1,4 @@
+embedder_libs:
```
sethladd: This has changed. It's now embedded_libs.

PR #1127, opened 2016-03-30
Add support for URL-based searches.

### LINKED ISSUE #1126: Support searching via URL
I'd like to set up a Chrome search handler to search the Dart core library APIs, but currently this isn't possible because there's no way to construct a URL for a search query. It would be great to support something like this.

One possibility for an easy implementation that avoids the need for a search results page:

* When a DartDoc page first loads, check for the presence of a query parameter named "search".
* If such a parameter exists, find the first autocomplete result that would be chosen if the search string were typed in the search bar.
* If there is such a result, navigate to its page.

nex3:
Closes #1126


keertip:
lgtm!

PR #1132, opened 2016-04-05
extend --include and --header

devoncarew:
Better support the use case of generating docs for multiple projects in one repo; this will allow Flutter to improve their hosted docs.

- added support for passing more than one `--header` or `--footer`
- added the ability to `--include` libraries referenced by (but not directly inside) the project
- rev the analyzer version used to fix an issue generating docs for Flutter

@keertip

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-# Don’t commit the following directories created by pub.
 build/
 doc/api/
 test_package/doc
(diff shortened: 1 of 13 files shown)


keertip:
@devoncarew , instead of changing the behavior of `--include`, could you add another flag for this functionality?

devoncarew:
ptal, updated to `--include-external`

keertip:
lgtm!

PR #1136, opened 2016-04-05
0.9.3+1

devoncarew:
@keertip, a bug fix release. We can document libraries twice in some cases. Changed from using a Set (the object's don't implement `==` or hashcode) to a List, and manually de-duping the items.


keertip:
lgtm

PR #1143, opened 2016-04-08
support for slightly wider docs

devoncarew:
- add support for slightly wider docs; the right-hand column is somewhat narrow when viewing a file with lots of long symbols; this makes this whole page wider on big displays, which it turns makes the right-hand column slightly wider

@keertip


lib/resources/styles.css
```diff
@@ -25,6 +25,12 @@ body {
   color: #212121;
 }
 
+@media (min-width: 1300px) {
+  .container {
+    width: 1270px;
```
devoncarew: normally 1170px wide on 1200px displays

devoncarew:
I'm not 100% sure that the wider page improves the look of the docs (it does help make long entries in the right-nav more readable).

Issue #1105, opened 2016-02-25
Provide a way to reference a sample code


luisvt:
It would be cool if we have a way to reference sample code similarly to what is been achieved in angular2 docs using jade.

I think they use something like this:

```
:makeExample('url/to/sample/file', 'name-of-block')
```

it would be cool if we could add something like that to docs and then the html get autogenerated with sample code.

sethladd:
Neat idea!

keertip:
Angular2 uses the `@example` tag to indicate sample code that should be injected into the documentation.

dartdoc should provide support for recognizing  
<code>
@example core/pipes/ts/lowerupper_pipe/lowerupper_pipe_example.ts region='LowerUpperPipe'
</code>

and adding the sample code referenced to the documentation.

keertip:
@sanfordredlich

sanfordredlich:
Thanks, it would be great to also have it create a log of @example entries it was not able to find on the file system, so we can ensure that they are created. This will make creating the documentation much more efficient.

chalin:
@keertip et all: FYI, I'll soon be looking into completing the implementation of this feature. Let me know if there are any objections and/or words of advice.

keertip:
@chalin , sounds good! 
Currently dartdoc has support for finding `@example` tags and looking for the files associated with it, and prints out a message if it is not found. The reading and injecting the example is what needs to be implemented.

PR #1154, opened 2016-04-29
add an option to inject examples

keertip:
This adds a flag `--include-examples` which for now looks for `@example` tag in the comments and looks for the file associated with the sample code.

#1105

### DIFF
--- a/bin/dartdoc.dart
+++ b/bin/dartdoc.dart
@@ -14,8 +14,6 @@ import 'package:dartdoc/src/package_meta.dart';
 import 'package:path/path.dart' as path;
 import 'package:stack_trace/stack_trace.dart';
 
-bool _showProgress = false;
-
 /// Analyzes Dart files and generates a representation of included libraries,
 /// classes, and members. Uses the current directory to look for libraries.
 main(List<String> arguments) async {
@@ -145,21 +143,7 @@ main(List<String> arguments) async {
   });
 }
 
-void _onProgress(File file) {
-  if (_showProgress) stdout.write('.');
-}
-
-/// Print help if we are passed the help option.
-void _printHelp(ArgParser parser, {int exitCode: 0}) {
-  print('Generate HTML documentation for Dart libraries.\n');
-  _printUsageAndExit(parser, exitCode: exitCode);
-}
-
-void _printUsageAndExit(ArgParser parser, {int exitCode: 0}) {
-  print('Usage: dartdoc [OPTIONS]\n');
-  print(parser.usage);
-  exit(exitCode);
-}
+bool _showProgress = false;
 
 ArgParser _createArgsParser() {
   var parser = new ArgParser();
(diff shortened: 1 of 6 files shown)


lib/src/model.dart
```diff
@@ -1605,6 +1611,27 @@ abstract class ModelElement implements Comparable, Nameable, Documentable {
...
+  String _injectExamples(String rawdocs) {
+    if (rawdocs.contains('@example')) {
+      RegExp exp = new RegExp(r"{@example .+}");
+      Iterable<Match> matches = exp.allMatches(rawdocs);
+      var dirPath = this.package.packageMeta.dir.path;
+      for (var match in matches) {
+        var strings = match.group(0).split(' ');
+        var file = new File(p.join(dirPath, 'examples', strings[1]));
+        if (file.existsSync()) {
+          // TODO(keertip):inject example
```
sethladd: did we mean to fill this out?

testing/test_package/lib/example.dart
```diff
@@ -162,6 +162,7 @@ class ConstantCat implements Cat {
 }
 
 /// implements [Cat], [E]
+/// {@example core/pipes/ts/slice_pipe/slice_pipe_example.ts region='SlicePipe_list'}
```
sethladd: What is this path relative to? We might want to document that fact in the main source for this feature?

bin/dartdoc.dart
```diff
@@ -195,6 +180,11 @@ ArgParser _createArgsParser() {
       allowMultiple: true, help: 'library names to generate docs for');
   parser.addOption('include-external',
       allowMultiple: true, help: 'additional (external) libraries to include');
+  parser.addFlag('include-examples',
+      defaultsTo: false,
+      negatable: false,
+      help:
+          "look for '@example' tags and inject the examples into the comment");
```
sethladd: Add a note for how to find an example? Or, show an example here? Also, how are the relative paths found?

lib/src/model.dart
keertip: Yes. Currently there are no examples in Angular2 dart, getting this in will give them a list of the missing examples.

testing/test_package/lib/example.dart
keertip: relative to the package. Yes, will document.

bin/dartdoc.dart
keertip: done.

devoncarew:
Two drive by comments:

- using both `--include-examples` and `@example` seems over-specified; what about just keying off a well-formed `@example` annotation?
- the other directive we recognize is `<nodoc>`; that's not consistent with the `@` annotation above; perhaps use the same style of directive?

keertip:
Thanks Devon for the comments!

Will remove the flag. For the `<nodoc>`, we can add support for a `@nodoc` format to make it consistent.

PR #1164, opened 2016-05-25
Prefer `wget latest` over chocolately.

pq:
As per discussion over on the linter (https://github.com/dart-lang/linter/issues/117), let's just stick to the latest rev and not try and keep up with chocolatey.

@keertip @devoncarew


keertip:
lgtm! Thanks for the cleanup.

PR #1167, opened 2016-05-28
Update dartdoc to use new EmbedderSdk API.

pq:
Addresses https://github.com/dart-lang/sdk/issues/26562.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -7,7 +7,7 @@ homepage: https://github.com/dart-lang/dartdoc
 environment:
   sdk: '>=1.14.0 <2.0.0'
 dependencies:
-  analyzer: 0.27.3-alpha.7
+  analyzer: 0.27.4-alpha.8
   args: ^0.13.0
   cli_util: ^0.0.1
   collection: ^1.2.0
--- a/lib/dartdoc.dart
+++ b/lib/dartdoc.dart
@@ -24,6 +24,7 @@ import 'package:analyzer/src/generated/sdk_io.dart';
 import 'package:analyzer/src/generated/source.dart';
 import 'package:analyzer/src/generated/source_io.dart';
 import 'package:path/path.dart' as p;
+import 'package:yaml/src/yaml_node.dart';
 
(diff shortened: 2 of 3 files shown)


pq:
cc @keertip @devoncarew

pubspec.yaml
```diff
@@ -7,7 +7,7 @@ homepage: https://github.com/dart-lang/dartdoc
 environment:
   sdk: '>=1.14.0 <2.0.0'
 dependencies:
-  analyzer: 0.27.3-alpha.7
+  analyzer: 0.27.4-alpha.8
```
pylaligand: It's a bit weird that the command failed even though the proper analyzer version was requested, but based on my quick look at how the dartdoc command is set up this is because we manually point the snapshot to a list of packages not necessarily built from this spec, correct?

pq: That surprised me too.  (And partly what lulled me into thinking this alpha push was OK.)

To be sure we're on the same page, you are running `dartdoc` from the dart SDK?  And the latest dev build?

devoncarew: Yeah, I think dartdoc from the sdk will be broken, since it doesn't use the pubspec constraints but the latest analyzer source in the sdk repo. If you run dartdoc via:

```
pub global activate dartdoc
pub global run dartdoc
```

_that_ version should work (and still be pegged to the older alpha analyzer release).

pylaligand: I confirm that I'm running dartdoc from the SDK.

pq: Great.  Thanks!

I've got this fixed in the SDK (as of https://github.com/dart-lang/sdk/commit/fba5cbc425e54b015d9ffefac4e0789d5502846c) and the next dev build should be good.  Sorry again for the oversight!

pq:
Closing in favor of restoring API (https://github.com/dart-lang/sdk/issues/26562).

pubspec.yaml
pq: FWIW a new dev build is available now.

PR #1177, opened 2016-06-17
`pub run test` fails if `dart` is not in your path

eseidelGoogle:
These also might fail if the `dart` in your path was
of a different version than you expected.  This happens
with Flutter since flutter downloads its own copy of the
Dart SDK which may be different from the one installed
globally on your disk.

In my case, I just was using an explicit path to `pub`
for `pub run test` and was surprised to see test failures.

@keertip

### DIFF
--- a/test/compare_output_test.dart
+++ b/test/compare_output_test.dart
@@ -60,8 +60,8 @@ void main() {
         tempDir.path
       ];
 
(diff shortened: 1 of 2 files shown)


keertip:
Thanks!

PR #1179, opened 2016-06-19
regen test docs

devoncarew:
- fix one analysis warning
- regen the test docs
- make a crossdoc test more maintainable (it depends on the precise char position of a symbol in `testing/test_package/lib/fake.dart`)

@keertip


keertip:
lgtm!

Issue #779, opened 2015-08-07
Make the left/right nav bars sticky


alan-knight:
If I'm looking at something large, e.g. https://new-dartdocs-dot-dartlang-api.appspot.com/1.12.0-dev.5.1/dart-html/Element-class.html 

and I want to find something in the index, say, a method, it's difficult to navigate. I have to hit page-down 7 times before I can see querySelectorAll, and there's not a way I can see to jump to the methods section.

sethladd:
![element_class_-_dart_html_library_-_dart_api](https://cloud.githubusercontent.com/assets/5479/9141336/c252b7c8-3cec-11e5-980e-72fe79badca8.png)


does the top-nav "tabs" work for you?

alan-knight:
Oh yeah, didn't notice those. But I was really looking for it for the index. So in a class with a lot of properties, I might want to scan the list of properties, and the index is more convenient than the full list. For example, what's the list of events in Element. I'm not sure what the one I'm looking for is called, or even if it exists.

Also, I think it would be better if those nav tabs were pinned to the top. If I'm down in the properties and realize I want a method, I have to scroll back to the top to get them.

sethladd:
> what's the list of events in Element.

How would we signal that in the UI? There's no language construct for "events"

> I think it would be better if those nav tabs were pinned to the top

That's true. It's a tweak we hope to make.

Not sure what to do with this issue. We can open a specific issue for "show tabs on top, if user scrolls". I don't think we're going to do anything about other ways to nav, because we have the list of main sections at the top already.

alan-knight:
I was just using events as an example of why I might want to scan the list of properties. For html elements that's usually a good percentage of the properties. So I just wanted the ability to jump quickly to the properties, which the top nav does give me, but better if it's always visible.

sethladd:
Agreed. We'd like to make those "table of contents" links visible even when the masthead is scrolled out of view.

keertip:
We should make the left/right navs sticky to make navigation easier.

Hixie:
For Flutter's purposes, I think this is probably more like a P2 than a P1 compared to the other P1s.

PR #1181, opened 2016-06-19
show the entire text for the first line of dartdoc

devoncarew:
Show the full text for the first line of dartdoc. This gives doc writers more control over what the summary of an item is, and gives more context for each item on the summary pages.

screen shot 2016-06-18 at 10 10 33 pm

### DIFF
--- a/lib/resources/styles.css
+++ b/lib/resources/styles.css
@@ -255,7 +255,6 @@ dd.callable, dd.constant, dd.property {
 }
 
 dd p {
-  white-space: nowrap;
   overflow-x: hidden;
   text-overflow: ellipsis;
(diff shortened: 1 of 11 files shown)


sethladd:
Nice!

keertip:
Was concerned that showing the first line full text would create noise on the page when there are a lot of methods. Turns out that more than classes with lots of methods, see libraries with many classes. The page does have more content, and navigation takes some more effort, and that's cause the left/right navs are not sticky. Increased visibility of #779 .

Lets land this and look to fixing the nav.

lgtm!

devoncarew:
Thanks the for review :)

Issue #1084, opened 2016-01-19
Incorrect HTML escaping for const expressions


sethladd:
For example:

![screen shot 2016-01-19 at 9 30 13 am](https://cloud.githubusercontent.com/assets/5479/12429673/650f4992-bea0-11e5-99d1-b8c647886b84.png)

From http://docs.flutter.io/flutter/material/material-library.html

Probably should be `const <int, List<BoxShadow>> {1 : const <BoxShadow> [` ...

sethladd:
From dartdoc 0.8.4

Hixie:
You can see this with kMaterialEdges now (same page).

keertip:
Looks like `kMaterialEdges` is missing type information after the `const`. Is that it? Don't see the `const >` as in the screenshot by Seth.

Hixie:
If you look at the HTML source you'll see there's unescaped HTML:
```html
          <div>
            <span class="signature"><code>const <MaterialType, double> {MaterialType.canvas : null, MaterialType.card : 2.0, MaterialType.circle : null, MaterialType.button : 2.0, MaterialType.transparency : null}</code></span>
          </div>
```

eseidel:
This is presumably the line:
https://github.com/dart-lang/dartdoc/blob/master/lib/templates/_constant.html#L8
Unclear why it uses {{{ instead of {{, the 3 { makes it unescaped according to the mustache manual:
https://mustache.github.io/mustache.5.html

devoncarew:
I think it needs to use the unescaped braces there because the constant string might have hyperlinks for some of the types referenced in it (`const Foo<double>()`).

PR #1187, opened 2016-06-23
Escape constants

devoncarew:
- fix an issue involving escaping constants (fix https://github.com/dart-lang/dartdoc/issues/1084)

@keertip


eseidel:
Why did a bunch of links go away in your expected results?

devoncarew:
It's a manual process to regenerate the reference docs, and the PR before me didn't include the regened docs. It makes this PR a little noisy -

test/model_test.dart
```diff
@@ -1260,12 +1260,12 @@ String topLevelFunction(int param1, bool param2, Cool coolBeans,
     });
 
     test('COLOR_ORANGE has correct value', () {
-      expect(orangeConstant.constantValue, "'orange'");
+      expect(orangeConstant.constantValue, "&#39;orange&#39;");
```
eseidel: It might be worth adding a test for documentation of a generic, since that was the failing case in flutter.  Doesn't look like current testing covers Foo<Bar,Baz>

devoncarew: Good point - added.

devoncarew:
screen shot 2016-06-22 at 9 32 41 pm

keertip:
lgtm!

There is a test that checks the changes against the checked in docs and flags for regen. Looks like the test is not quite working. opened #1190 to track.

PR #1192, opened 2016-06-29
remove den_api from dev dependencies

keertip:
- removed den_api from dev dependencies
- remove check for format on files on travis


tool/grind.dart
```diff
@@ -93,25 +178,8 @@ test() {
...
-  String libCodeContents = libCode.readAsStringSync();
-  libCodeContents = libCodeContents.replaceFirst(
-      new RegExp(r"const String version = '.*';"),
-      "const String version = '${pubspec.version}';");
-  libCode.writeAsStringSync(libCodeContents);
-
-  log('Version set to ${pubspec.version}');
-}
-
+// TODO: check http links, check links in <link>
```
devoncarew: If we're not going to do this, we can just remove these todos.

devoncarew:
I think we can delete ensure_dartfmt.sh as well.

lgtm! Great to see all those deps disappear.

tool/grind.dart
keertip: removed.

Issue #1190, opened 2016-06-23
Test for validity of generated docs not catching changes


keertip:
From the comments in #1187, it looks like we are not catching the need to regen the checked in docs with the change. 

The existing test should be fixed.

PR #1198, opened 2016-07-19
fix test that checks validity of generated docs

danrubel:
Fixes https://github.com/dart-lang/dartdoc/issues/1190

@keertip


keertip:
lgtm! Thanks for fixing the test.

PR #1199, opened 2016-07-19
cleanup bin/dartdoc to use public dartdoc API

danrubel:
This makes dartdoc initialization part of the public API
and updates bin/dartdoc to use it.
Needed by dartinodoc.

@keertip


keertip:
lgtm

Issue #1188, opened 2016-06-23
test failures on windows


devoncarew:
We're seeing two test failures on windows:

https://ci.appveyor.com/project/devoncarew/dartdoc/build/1.0.1449

PR #1200, opened 2016-07-19
fix @example replacement and crossdart on Windows

danrubel:
Update code to correctly process @example and crossdart on Windows
Fixes https://github.com/dart-lang/dartdoc/issues/1188
@keertip


keertip:
lgtm! Thanks for testing on Windows.

PR #1202, opened 2016-07-20
organize imports

devoncarew:
- organize imports
- add some missing license headers

@keertip @pq


pq:
Assuming travis is happy, LGTM!

PR #1211, opened 2016-07-21
Fix for issue with references in parameters and return type pointing …

### LINKED ISSUE #1160: Flutter: dartdocs gets wrong class for "CommonFinders"
We have two `find` top-level members:

http://docs.flutter.io/flutter/flutter_test/find.html
http://docs.flutter.io/flutter/flutter_driver/find-constant.html

Both are of type "CommonFinders", but those are different types. Yet they both point to the same class definition:
http://docs.flutter.io/flutter/flutter_driver/CommonFinders-class.html

This is wrong.

The two source files are:
https://github.com/flutter/flutter/blob/master/packages/flutter_test/lib/src/finders.dart
https://github.com/flutter/flutter/blob/master/packages/flutter_driver/lib/src/driver.dart

### LINKED ISSUE #1195: Flutter docs point to the wrong Point
The links in http://docs.flutter.io/flutter/dart-ui/Point-class.html target the Point class from dart:math rather than the dart_ui Point class. Likewise for the Canvas docs and (I would guess) other classes that refer to Point.

(From https://github.com/flutter/flutter/issues/4884)

keertip:
…to wrong element with same name

closes #1195, 
closes #1160 

@devoncarew


lib/src/model.dart
```diff
@@ -1653,6 +1653,19 @@ abstract class ModelElement implements Comparable, Nameable, Documentable {
     return '<a ${classContent}href="${href}">$name</a>';
   }
 
+  // TODO: consolidate all the find library methods
```
devoncarew: please add your username here

lib/src/model.dart
```diff
@@ -1653,6 +1653,19 @@ abstract class ModelElement implements Comparable, Nameable, Documentable {
...
+  Library _findLibraryFor(Element e) {
+    var eelement = e.getAncestor((l) => l is LibraryElement);
```
devoncarew: `eelement` ==> `element` ?

devoncarew:
lgtm

PR #1218, opened 2016-07-22
Fix for duplicate libraries in SDK docs

### LINKED ISSUE #1210: api.dartlang shows two dart:math libraries in left nav bar
This is for the 1.18 dev channel

screen shot 2016-07-21 at 9 44 06 am

keertip:
closes #1210 

@pq, @devoncarew

### DIFF
--- a/lib/dartdoc.dart
+++ b/lib/dartdoc.dart
@@ -44,19 +44,6 @@ const String version = '0.9.6+2';
 
 final String defaultOutDir = p.join('doc', 'api');
 
-/// Configure the dartdoc generation process
-void initializeConfig(
-    {Directory inputDir,
-    String sdkVersion,
-    bool addCrossdart: false,
-    bool includeSource: true}) {
-  setConfig(
-      inputDir: inputDir,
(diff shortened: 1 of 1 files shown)


lib/dartdoc.dart
```diff
@@ -196,7 +196,8 @@ class DartDoc {
       ..sourceFactory = sourceFactory;
 
     if (packageMeta.isSdk) {
-      libraries.addAll(getSdkLibrariesToDocument(sdk, context));
+      libraries
+          .addAll(new Set()..addAll(getSdkLibrariesToDocument(sdk, context)));
```
pq: Would be awesome if collections supported a `distinct()` method, so we could just say `getSdkLibrariesToDocument(sdk, context).distinct()`.  That said, a set works in the meantime! 👍

pq: Alternatively, I guess `getSdkLibrariesToDocument()` might enforce uniqueness.  Would there be any benefit to doing it there?

pq:
LGTM.

Issue #1077, opened 2016-01-05
Truncate super duper long const values?


sethladd:
For example, check out http://docs.flutter.io/flutter/material/material-library.html

What if we truncated incredibly long const values on the "index" page?

![screen shot 2016-01-05 at 9 28 04 am](https://cloud.githubusercontent.com/assets/5479/12122261/be355ef2-b38e-11e5-96c1-a9ace361ca1a.png)

Hixie:
We worked around this by moving the default value to a private constant, but the underlying bug still exists.

PR #1219, opened 2016-07-22
fix the display of long constants

devoncarew:
On the summary page, one display one line of text for constants (fix https://github.com/dart-lang/dartdoc/issues/1077).

Also, update how we style in-line code (to adjust to the markdown library rev).

@pq @keertip


keertip:
lgtm

PR #1220, opened 2016-07-22
minor change to the html generator

devoncarew:
- minor chage to the html generator to create a parent directory if it doesn't exist, instead of creating an empty file

@pq @keertip


keertip:
lgtm

PR #1221, opened 2016-07-22
html escape unresolved references

devoncarew:
- html escape unresolved square bracket references, so things like unresolved generics aren't converted into inline html

@pq @keertip


keertip:
lgtm

Issue #1146, opened 2016-04-12
Better screen real estate usage in multi-package dartdocs


yjbanov:
Flutter publishes its dartdocs as one site for all packages. It now look great! However, there's an opportunity to use the screen space more economically. Specifically, there is some redundant information in the header that increases the height of the header unnecessarily. Below is the screenshot with sections highlighted in red that could be removed without losing any useful information but shrinking the header size considerably:

![flutter-docs-screen-real-estate](https://cloud.githubusercontent.com/assets/211513/14475322/421d6aec-00b6-11e6-824f-bc096a8e3a30.png)

sethladd:
Good catch. It's not only duplicate, but incorrect. It's not really a "package" anymore. Maybe, if the categories are used, we can eliminate the word PACKAGE in the header, at least on the homepage.

devoncarew:
I could see it being more condensed vertically. It  would be nice if it lined up with the header size for https://flutter.io/ so that the page didn't jump around when you move from one part of the site to another.

This main dartdoc landing page doesn't use the space in the header for as many things as the follow on pages (a library page, or and class page). We'd want to decide if we need that content for the sub-pages. I do want to keep the header height for the various generated dartdoc pages the same, so that the header size doesn't jump around as you navigate through dartdoc pages.

devoncarew:
We could probably get pretty far here just with some small changes to our stylesheet.

devoncarew:
largely addressed - closing

PR #1223, opened 2016-07-23
don't show package label for sdk docs

devoncarew:
- if we're generating docs that use categories, don't show the 'package' label for the top-most index.html (https://github.com/dart-lang/dartdoc/issues/1146)
- fix an issue where the subnav headers wouldn't show all relevant options

screen shot 2016-07-22 at 10 02 47 pm

@keertip


keertip:
lgtm

PR #1224, opened 2016-07-25
fix issues with dispaying constants

devoncarew:
- change how we truncate long constant values on the summary page (the css method was causing display issues for long constants)
- show the full docs for enums on the summary page; just show the first line of docs for other constants

@keertip


keertip:
lgtm! Like the unit tests 👍

Issue #1233, opened 2016-08-26
[1.19.0] can't generate docs for angular2 anymore - analyzer NoSuchMethodError


chalin:
Having upgraded to 1.19.0, I can no longer build the API docs for angular2. When I try I get:
```
Generating documentation for 'angular2' into /Users/chalin/tmp/api/

Generation failed: Class 'JavaFile' has no instance getter 'path'.

NoSuchMethodError: method not found: 'path'
Receiver: Instance of 'JavaFile'
Arguments: []
dart:core                                                                                  Object.noSuchMethod
package:analyzer/source/pub_package_map_provider.dart 173                                  PubPackageMapProvider._runPubListDefault
package:analyzer/source/pub_package_map_provider.dart 74                                   PubPackageMapProvider.computePackageMap
package:dartdoc/dartdoc.dart 150                                                           DartDoc._parseLibraries
package:dartdoc/dartdoc.dart 93                                                            DartDoc.generateDocs.<async>
===== asynchronous gap ===========================
dart:async                                                                                 _Completer.completeError
package:dartdoc/dartdoc.dart 139                                                           DartDoc.generateDocs.<async>
===== asynchronous gap ===========================
dart:async                                                                                 Future.Future.microtask
package:dartdoc/dartdoc.dart                                                               DartDoc.generateDocs
/b/build/slave/dart-sdk-mac-stable/build/sdk/third_party/pkg/dartdoc/bin/dartdoc.dart 142  main.<async>.<fn>.<async>
===== asynchronous gap ===========================
package:stack_trace                                                                        Chain.capture
/b/build/slave/dart-sdk-mac-stable/build/sdk/third_party/pkg/dartdoc/bin/dartdoc.dart 141  main.<async>
dart:async                                                                                 _SyncCompleter.complete
package:dartdoc/dartdoc.dart 50                                                            initGenerators.<async>
dart:async                                                                                 _SyncCompleter.complete
package:dartdoc/src/html/html_generator.dart 67                                            HtmlGenerator.create.<async>
dart:async                                                                                 _SyncCompleter.complete
package:dartdoc/src/html/templates.dart 115                                                Templates.create.<async>
dart:async                                                                                 _SyncCompleter.complete
package:dartdoc/src/html/templates.dart 98                                                 Templates.create.<async>._loadTemplate.<async>
dart:async                                                                                 _SyncCompleter.complete
package:dartdoc/src/html/resource_loader.dart 22                                           loadAsString.<async>
[...]

PR #1234, opened 2016-08-27
Fix package regression

devoncarew:
- fix for https://github.com/dart-lang/dartdoc/issues/1233
- change from the older list-package-dirs method of locating packages to using the `.packages` file
- update several usages of now deprecated APIs


devoncarew:
@keertip

keertip:
lgtm! Thanks for much!

PR #1240, opened 2016-08-31
Widen constraint on package_config

whesse:
package_config has released version 1.0.0, which is compatible with earlier versions.  Relax constraint to <2.0.0.   This could also be changed to ^1.0.0 if you want.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,6 +1,6 @@
 name: dartdoc
 # Also update the `version` field in lib/dartdoc.dart.
-version: 0.9.7+2
+version: 0.9.7+3
 author: Dart Team <misc@dartlang.org>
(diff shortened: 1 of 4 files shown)


whesse:
This is needed for the pinned versions of packages in dart-lang/sdk/DEPS

pubspec.yaml
```diff
@@ -1,6 +1,6 @@
 name: dartdoc
 # Also update the `version` field in lib/dartdoc.dart.
-version: 0.9.7+2
+version: 0.9.7+3
```
devoncarew: Can you also update the constant in `lib/dartdoc.dart`?

devoncarew:
lgtm after the constant is updated

Issue #1230, opened 2016-08-17
dart:math is listed twice on api.dartlang.org


mit-mit:
_From @kmillikin on August 16, 2016 12:21_

When I look at api.dartlang.org I see dart:math listed twice on the left-hand side.

_Copied from original issue: dart-lang/sdk#27083_

mit-mit:
_From @kmillikin on August 16, 2016 13:21_

I just noticed that it's also listed twice in the body of the page.

devoncarew:
@keertip, could you take a look?

devoncarew:
This is fixed in the latest version of dartdoc: https://api.dartlang.org/dev/1.20.0-dev.1.0/index.html.

PR #1242, opened 2016-09-01
improve sdk docs validation

devoncarew:
- add a test for https://github.com/dart-lang/dartdoc/issues/1230; improve the sdk docs validation to also check the number of `dart:` libs listed in index.html.
- fix an issue w/ how we were using `await` in the `buildSdkDocs` task

@keertip


keertip:
lgtm

Issue #1244, opened 2016-09-01
Allow dependency html 0.13


whesse:
The html package has released version 0.13, with breaking changes
BREAKING Fix all strong mode errors and warnings. This involved adding more precise types on some public APIs, which is why it may break users.

If dartdoc works with version 0.13, please relax the version constraint to >=12.0 <14.0, or to ^13.0 if you don't want to support 0.12 anymore.

This is blocking rolling package:html in sdk/DEPS.

PR #1245, opened 2016-09-01
upgrade the html package dep

devoncarew:
- upgrade the html package dep
- fix https://github.com/dart-lang/dartdoc/issues/1244

@keertip @whesse


whesse:
I don't understand why we have the .pubspec.lock file committed into the repo.  Shouldn't this be .gitignored, and not in the repo?  If it is common practice to include this file, then OK, but I didn't think it was.

sethladd:
Previous best practice was to commit the .lock file for apps. Libraries should not have a .lock file committed. Maybe that's changed?

whesse:
Eric says that in general, the pubspec.lock should be committed in the repo for applications, but not for libraries.  But I suppose it doesn't hurt anything for a library - things using the library wouldn't even see it.

devoncarew:
I'm confused - dartdoc is an app; per conventions we should be checking in the lock file, no? 

https://github.com/dart-lang/dartdoc/commit/96fadf092744fcb74b501ac2d3ab7df207b63543#diff-7ed8db171f9d2dbefb97f0043b4eb2ff

keertip:
lgtm

PR #1258, opened 2016-10-08
prefer running from the SDK

devoncarew:
- remove documentation on installing via `pub global activate`; prefer installing via the SDK distribution

@keertip


keertip:
lgtm

PR #1269, opened 2016-11-07
Hacky annotation support on properties

rkj:


keertip:
@rkj, can you do a

`pub global activate grinder`
`grind update-test-package-docs`

and add the changes to PR to get the test package docs updated. Thanks!

devoncarew:
@rkj, ping re: this PR - we need a little work (from the comment above) to make the bots happy. Thanks!

PR #1275, opened 2016-11-17
Replace uses of deprecated analyzer APIs

bwilkerson:
@keertip


keertip:
@bwilkerson, has anything changed with how dartdoc gets all the sdk llibraries when generating docs for the sdk? See that the sdk doc generation is failing on travis for both dev and stable.

bwilkerson:
Yes. If you have some time I could use some help figuring out how to debug this.

PR #1280, opened 2016-11-26
remove deps on cli_util and which

devoncarew:
- remove deps on `package:cli_util` and `package:which`


devoncarew:
@keertip for review

eseidelGoogle:
@devoncarew Flutter was using --dart-sdk, do we no longer need to when updating past this change?
https://github.com/flutter/flutter/blob/master/dev/tools/dartdoc.dart#L65

devoncarew:
Yeah, --dart-sdk is no longer necessary. There's a new(er) API in dart which allows us to locate the sdk we're running from (https://api.dartlang.org/stable/1.21.0/dart-io/Platform/resolvedExecutable.html).

Issue #1283, opened 2016-11-28
update the dartdoc dep in the sdk repo


devoncarew:
In support of https://github.com/dart-lang/dartdoc/issues/1278 (generic methods), we need to bring the latest dartdoc into the sdk repo (and likely published a new version of dartdoc).

keertip:
https://codereview.chromium.org/2535863003/

PR #1284, opened 2016-11-28
rev version to 0.9.8

keertip:
#1283 

@devoncarew


devoncarew:
lgtm; will need to re-gen the test docs before landing

PR #1289, opened 2016-12-14
Change the --include-external flag to be more selective

### LINKED ISSUE #1236: flutter docs are missing lots of classes
The 0.9.7+2 release introduced an issue where the generated docs for flutter are missing many classes.

This page: https://docs.flutter.io/flutter/material/material-library.html

only shows ~5 classes.

keertip:
Fixes #1236 

@devoncarew

### DIFF
--- a/lib/dartdoc.dart
+++ b/lib/dartdoc.dart
@@ -252,12 +252,13 @@ class DartDoc {
     while (result.hasMoreWork) {
       result = context.performAnalysisTask();
     }
-
+    
     // Use the includeExternals.
     for (Source source in context.librarySources) {
       LibraryElement library = context.computeLibraryElement(source);
       String libraryName = Library.getLibraryName(library);
-      if (includeExternals.contains(libraryName)) {
+      var fullPath = source.fullName;
+      if (includeExternals.any((string) => fullPath.endsWith(string))) {
         if (libraries.map(Library.getLibraryName).contains(libraryName)) {
           continue;
         }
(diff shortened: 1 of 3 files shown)


devoncarew:
(@keertip, you're seeing the same failure as from https://github.com/dart-lang/dartdoc/pull/1287)

devoncarew:
Awesome! lgtm

We when this lands in the sdk, we should rev. the flutter script to use the updated params (and to pull from the dartdoc in the sdk).

sethladd:
@devoncarew should we open an issue on Flutter side so we don't forget?

devoncarew:
👍 sounds good

eseidelGoogle:
I happened to notice that Travis was a sad panda, just FYI:
https://travis-ci.org/dart-lang/dartdoc/jobs/183981580#L319

devoncarew:
Yup, fixed in https://github.com/dart-lang/dartdoc/pull/1291.

lib/dartdoc.dart
```diff
@@ -252,12 +252,13 @@ class DartDoc {
     while (result.hasMoreWork) {
       result = context.performAnalysisTask();
     }
-
+    
     // Use the includeExternals.
     for (Source source in context.librarySources) {
       LibraryElement library = context.computeLibraryElement(source);
```
eseidelGoogle: This and libraryName are no longer used in the case where the library doesn't match.  No clue if it's expensive or not.  (Sorry to keep commenting on your completed CL, was just playing around in this code and noticed.)  Thanks again for the fix! :)

keertip: We can always optimize! Though you will see a slowdown once flutter is back to generating *all* the docs, I saw ~20 sec difference on my machine between the two cases.

PR #1293, opened 2016-12-14
Add warning/error handling for --include-external

eseidelGoogle:
This makes --include-external a bit harder to use incorrectly (with surprising results).

Cases I would like to test (but don't see an easy way how would include):
1. `--external-library asldkfjsdf` hit the "warn no libraries found" case
2. `--external-library dart` hit the "multiple libraries found" case
3. `--external-library dartdoc.dart` hit the "library already found" case

I looked at adding tests to test/dartdoc_test.dart, but those do not seem set up to handle logging.  I looked briefly ad breaking this logic out into its own function, but it wasn't clear how to provide the necessary AnalysisContext with sources.

I've been able to trigger the first two test cases manually, but not yet seen the 3rd.

### DIFF
--- a/lib/dartdoc.dart
+++ b/lib/dartdoc.dart
@@ -254,16 +254,36 @@ class DartDoc {
     }
 
     // Use the includeExternals.
-    for (Source source in context.librarySources) {
+    bool matchesLibrary(String includeExternal, Source source) {
       LibraryElement library = context.computeLibraryElement(source);
       String libraryName = Library.getLibraryName(library);
-      var fullPath = source.fullName;
-      if (includeExternals.any((string) => fullPath.endsWith(string))) {
-        if (libraries.map(Library.getLibraryName).contains(libraryName)) {
-          continue;
-        }
-        libraries.add(library);
+      return source.fullName.endsWith(includeExternal);
+    }
+
+    for (String includeExternal in includeExternals) {
+      var quotedArgString = "'--include-external $includeExternal'";
+      List<Source> matchedSources = context.librarySources
+          .where((library) => matchesLibrary(includeExternal, library));
+      if (matchedSources.isEmpty) {
+        print("  warning: $quotedArgString -- " +
+            "no library found matching $includeExternal.");
+        continue;
       }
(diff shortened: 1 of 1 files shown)


lib/dartdoc.dart
```diff
@@ -254,16 +254,36 @@ class DartDoc {
...
-        libraries.add(library);
+      return source.fullName.endsWith(includeExternal);
+    }
+
+    for (String includeExternal in includeExternals) {
+      var quotedArgString = "'--include-external $includeExternal'";
+      List<Source> matchedSources = context.librarySources
+          .where((library) => matchesLibrary(includeExternal, library));
+      if (matchedSources.isEmpty) {
+        print("  warning: $quotedArgString -- " +
```
kevmoo: FYI: `+` not needed here. adjacent String literals auto-concat. Is nice.

lib/dartdoc.dart
```diff
@@ -254,16 +254,36 @@ class DartDoc {
...
-          continue;
-        }
-        libraries.add(library);
+      return source.fullName.endsWith(includeExternal);
+    }
+
+    for (String includeExternal in includeExternals) {
+      var quotedArgString = "'--include-external $includeExternal'";
+      List<Source> matchedSources = context.librarySources
+          .where((library) => matchesLibrary(includeExternal, library));
```
kevmoo: add `.toList()` here – make sure we don't re-iterate the iterable.

eseidelGoogle: What would it mean to re-iterate the iterable?  Is this a general pattern we should have the analyzer warn about?

lib/dartdoc.dart
eseidelGoogle: Neat!

eseidelGoogle:
Thanks for the comments!  I actually think I should drop this pull request.  I mostly was just in the code and wrote it up for funsies.  What @keertip provided in https://github.com/dart-lang/dartdoc/pull/1289 was sufficient for Flutter's needs and I'm likely not going to find time to take this to completion.

PR #1295, opened 2016-12-16
feat: support @example insertion of .md file fragments

chalin:
cc @filiph @kwalrath


chalin:
cc @matanlurey @ferhatb 

FYI, here is the internal documentation for the feature (so far):
> Replace `{@example ...}` in API comments with the content of named file.
> 
> Syntax:
> 
>     {@example PATH [region=NAME] [lang=NAME]}
> 
> where PATH and NAME are tokens _without_ whitespace; NAME can optionally be
> quoted (use of quotes is for backwards compatibility and discouraged).
> If PATH is `dir/file.ext` and region is `r` then we'll look for the file
> named `dir/file-r.ext.md`, relative to the project root directory (of the
> project for which the docs are being generated).
> 
> Examples:
> 
>     {@example examples/angular/quickstart/web/main.dart}
>     {@example abc/def/xyz_component.dart region=template lang=html}

PR #1296, opened 2016-12-16
don't use short links when generating for crossdart

devoncarew:
@jcollins-g, when we're in the 'generating for crossdart' mode, don't use the short form for hyperlinks. I didn't get to the root of the issue, but the short reference form and crossdart are at odds somehow.


devoncarew:
Hmm, nevermind, this is not a fix -

PR #1297, opened 2016-12-17
Add showing docs of overridden accessors [#1266]

### LINKED ISSUE #1266: ThemeData.hashCode has no documentation
I don't understand why this is blank:
https://docs.flutter.io/flutter/material/ThemeData/hashCode.html

Shouldn't it inherit it from Object.hashCode?
https://docs.flutter.io/flutter/dart-core/Object/hashCode.html

astashov:
For methods, in case there's a method, which overrides another method,
and it doesn't have docs, we show the docs from the overridden method.

@Hixie noticed we don't do the same thing for accessors. He and
@devoncarew think we should do the same for accessors as well.

This commit adds that functionality.

(fix #1266)


astashov:
Now appveyor fails, seems like because of network error. Could you please rerun it? I can't.

devoncarew:
Sorry for the delay! Many of us are out on vacation; re-running appveyor now.

devoncarew:
And it's green! Merging...

astashov:
> Sorry for the delay! Many of us are out on vacation; re-running appveyor now.

No problem, no rush here :)

astashov:
Thanks!

Issue #1158, opened 2016-05-04
In Flutter docs, classes are listed multiple times


Hixie:
When a library exports another library, but both libraries are being docced together, the result is that each class that is exported by both libraries is listed twice.

We should only list classes as coming from the package that they actually come from, and not mention them in packages that re-export them, if we are including that package in the docs.

See also:
- https://github.com/dart-lang/dartdoc/issues/1090 (about where the links to reexported symbols should point)
- https://github.com/dart-lang/dartdoc/issues/936 (about saying where types come from when reexported)
- https://github.com/dart-lang/dartdoc/issues/1159 (about automatically documenting even non-exported libraries)

cc @devoncarew

devoncarew:
also adding @keertip

abarth:
As an example, the BoxDecoration class has four different URLs:

https://docs.flutter.io/flutter/painting/BoxDecoration-class.html
https://docs.flutter.io/flutter/rendering/BoxDecoration-class.html
https://docs.flutter.io/flutter/widgets/BoxDecoration-class.html
https://docs.flutter.io/flutter/material/BoxDecoration-class.html

It's not obvious that these are the same class.  If there was a canonical URL that all the re-exports linked to, that would make that much clearer.

astashov:
I will look at it.

astashov:
> We should only list classes as coming from the package that they actually come from, and not mention them in packages that re-export them, if we are including that package in the docs.

Hmm, how do we know what's the "canonical" library? I.e. `library foo` and `library bar` both export `lib/src/blah.dart` with `foobar` function, what library docs will contain that `foobar` function? `foo` or `bar`?

> It's not obvious that these are the same class. If there was a canonical URL that all the re-exports linked to, that would make that much clearer.

Yeah, we can do that, the question is - how do we know which one is canonical? Just the first one encountered during parsing?

abarth:
That isn't what's happening here.  What's happening here is:

1. A defines Foo.
2. Foo doesn't appear in the docs for A because A is a private library, i.e., in `lib/src/../a.dart`
3. B exports Foo from A.
4. Foo appears in the docs for B.
5. C exports Foo from B.
6. Foo appears in the docs for C.

The issue is at step (6) above.  Specifically, Foo shouldn't appear in the docs for C (or at least should be de-emphasized in some way) because Foo was exported from B and B already had docs.

astashov:
Aha, I see, thanks.

[...]

PR #1298, opened 2016-12-17
Resolve non-imported symbols in comments [#1153]

### LINKED ISSUE #1153: handle dartdoc references to non-imported symbols (was: dartdoc-specific imports)
Sometimes, a library, like the Flutter rendering library, wants to refer to another library, like the Flutter widgets library, in the documentation, but doesn't want to introduce a dependency in the code. In this case, for example, we definitely do not want any sort of runtime relationship from the rendering library to the widgets library.

We can't use `import` for this because it risks code accidentally depending through the import.

One possible solution would be to have something like `/// @import` which would cause the dartdoc code to import it. Another would be for the dartdoc tool to be able to receive a list of imports to imply at the top of every file. I don't really mind how we solve this.

astashov:
@Hixie noticed, that sometimes, a library, like the Flutter rendering
library, wants to refer to another library, like the Flutter widgets
library, in the documentation, but doesn't want to introduce a
dependency in the code. Currently, there’s no mechanisms in dartdoc,
which allow that.

This commit adds that. You can use either a short name (e.g. [icon]) or
a fully qualified name (like, [material.Icon.icon]), in the HTML docs,
it always will be shown as a short name though (is that a desired
behavior?).  Dartdoc will go over all the libraries, classes, methods,
properties, constants of the package that is being documented, and will
try to resolve the symbol. If it’s successful, it will add the link to
it, same way as when the symbol is in scope.

Some outstanding questions:

* What to do in case of ambiguity here? Show a warning, crash with an
  error message, ignore?
* Do we want to show short name in case a fully qualified name is
  provided? I saw in the comments in the ticket
  (https://github.com/dart-lang/dartdoc/issues/1153#issuecomment-236012803) that @Hixie would want that.
* Anything else?

Testing: Unit tests, of course, also tried to generate Flutter docs with
that, and observed that the links for previously non-resolved symbols
work.

(fixes #1153)

### DIFF
--- a/lib/src/markdown_processor.dart
+++ b/lib/src/markdown_processor.dart
@@ -28,6 +28,12 @@ const List<String> _oneLinerSkipTags = const ["code", "pre"];
 
 final List<md.InlineSyntax> _markdown_syntaxes = [new _InlineCodeSyntax()];
 
+class MatchingLinkResult {
+  final ModelElement element;
+  final String label;
+  MatchingLinkResult(this.element, this.label);
+}
+
 // TODO: this is in the wrong place
 NodeList<CommentReference> _getCommentRefs(ModelElement modelElement) {
   if (modelElement == null) return null;
@@ -67,10 +73,10 @@ NodeList<CommentReference> _getCommentRefs(ModelElement modelElement) {
 }
 
 /// Returns null if element is a parameter.
-ModelElement _getMatchingLinkElement(
+MatchingLinkResult _getMatchingLinkElement(
     String codeRef, ModelElement element, List<CommentReference> commentRefs,
     {bool isConstructor: false}) {
-  if (commentRefs == null) return null;
+  if (commentRefs == null) return new MatchingLinkResult(null, null);
 
   Element refElement;
   bool isEnum = false;
@@ -87,7 +93,9 @@ ModelElement _getMatchingLinkElement(
   }
 
   // Did not find an element in scope
-  if (refElement == null) return null;
+  if (refElement == null) {
+    return _findRefElementInLibrary(codeRef, element, commentRefs);
+  }
 
   if (refElement is PropertyAccessorElement) {
     // yay we found an accessor that wraps a const, but we really
@@ -99,7 +107,7 @@ ModelElement _getMatchingLinkElement(
     }
   }
 
-  if (refElement is ParameterElement) return null;
+  if (refElement is ParameterElement) return new MatchingLinkResult(null, null);
 
   // bug! this can fail to find the right library name if the element's name
   // we're looking for is the same as a name that comes in from an imported
@@ -114,24 +122,66 @@ ModelElement _getMatchingLinkElement(
     // Is there a way to pull this from a registry of known elements?
     // Seems like we're creating too many objects this way.
     if (isEnum) {
-      return new EnumField(refElement, refLibrary);
+      return new MatchingLinkResult(new EnumField(refElement, refLibrary), null);
     }
-    return new ModelElement.from(refElement, refLibrary);
+    return new MatchingLinkResult(new ModelElement.from(refElement, refLibrary), null);
   }
-  return null;
+  return new MatchingLinkResult(null, null);
+}
+
+MatchingLinkResult _findRefElementInLibrary(String codeRef, ModelElement element, List<CommentReference> commentRefs) {
+  final package = element.library.package;
+  final Map<String, ModelElement> result = {};
+  package.libraries.forEach((library) {
+    final List<ModelElement> modelElements = []
+      ..addAll(library.allClasses)
+      ..addAll(library.constants)
+      ..addAll(library.enums)
+      ..addAll(library.functions)
+      ..addAll(library.properties)
+      ..addAll(library.typedefs);
+
+    library.allClasses.forEach((c) {
+      modelElements.addAll(c.allInstanceMethods);
+      modelElements.addAll(c.allInstanceProperties);
+      modelElements.addAll(c.allOperators);
+      modelElements.addAll(c.staticMethods);
+      modelElements.addAll(c.staticProperties);
+    });
+
+    modelElements.forEach((modelElement) {
+      if (codeRef == modelElement.name || codeRef == modelElement.fullyQualifiedName) {
+        result[modelElement.fullyQualifiedName] = modelElement;
+      }
+    });
+  });
+
+  if (result.isEmpty) {
+    return new MatchingLinkResult(null, null);
+  } else if (result.length == 1) {
+    return new MatchingLinkResult(result.values.first, result.values.first.name);
+  } else {
+    if (_emitWarning) {
+      print("Abiguous reference '${codeRef}' in '${element.fullyQualifiedName}', " +
+        "we found it in the following elements: ${result.keys.map((k) => "'${k}'").join(", ")}");
+    }
+    return new MatchingLinkResult(null, null);
+  }
 }
 
 String _linkDocReference(String reference, ModelElement element,
     NodeList<CommentReference> commentRefs) {
-  ModelElement linkedElement;
   // support for [new Constructor] and [new Class.namedCtr]
   var refs = reference.split(' ');
+  MatchingLinkResult result;
   if (refs.length == 2 && refs.first == 'new') {
-    linkedElement = _getMatchingLinkElement(refs[1], element, commentRefs,
+    result = _getMatchingLinkElement(refs[1], element, commentRefs,
         isConstructor: true);
   } else {
-    linkedElement = _getMatchingLinkElement(reference, element, commentRefs);
+    result = _getMatchingLinkElement(reference, element, commentRefs);
   }
+  final linkedElement = result.element;
+  final label = result.label ?? reference;
   if (linkedElement != null) {
     var classContent = '';
     if (linkedElement.isDeprecated) {
@@ -139,12 +189,12 @@ String _linkDocReference(String reference, ModelElement element,
     }
     // this would be linkedElement.linkedName, but link bodies are slightly
     // different for doc references. sigh.
-    return '<a ${classContent}href="${linkedElement.href}">$reference</a>';
+    return '<a ${classContent}href="${linkedElement.href}">$label</a>';
   } else {
     if (_emitWarning) {
       print("  warning: unresolved doc reference '$reference' (in $element)");
     }
-    return '<code>${HTML_ESCAPE.convert(reference)}</code>';
+    return '<code>${HTML_ESCAPE.convert(label)}</code>';
   }
 }
 
--- a/lib/src/markdown_processor.dart
+++ b/lib/src/markdown_processor.dart
@@ -9,13 +9,7 @@ import 'dart:convert';
 
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/element/element.dart'
-    show
-        LibraryElement,
-        Element,
-        ConstructorElement,
-        ClassElement,
-        ParameterElement,
-        PropertyAccessorElement;
+    show LibraryElement, Element, ConstructorElement, ClassElement, ParameterElement, PropertyAccessorElement;
 import 'package:html/parser.dart' show parse;
 import 'package:markdown/markdown.dart' as md;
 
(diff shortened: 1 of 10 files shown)


devoncarew:
Reviewing now -

> What to do in case of ambiguity here? Show a warning, crash with an error message, ignore?

I believe @Hixie mentioned fast-fail. We should definitely at least warn. Dartdoc does generate docs unless there are significant parse errors in the code - it's lenient towards somewhat bad source, so you can always generate docs.

The approach the analyzer takes is to allow clients to opt-into failing on warnings (`--fatal-warnings`). This is probably what we should do here as well. It's outside the scope of this PR however, so I would say warn, and open an issue to add a `--fatal-warnings` flag and associated behavior.

> Do we want to show short name in case a fully qualified name is provided? I saw in the comments in the ticket

Sounds like (from @Hixie) yes :)

devoncarew:
Re: `material.Icon.icon` as the fully qualified name - there are cases where this format is not unique (specifically, in the flutter package, there are two libraries with the name `material`). While that's a nice, clean fully specified format, we may need something more unique.

lib/src/markdown_processor.dart
```diff
@@ -114,37 +122,79 @@ ModelElement _getMatchingLinkElement(
...
+  MatchingLinkResult result;
   if (refs.length == 2 && refs.first == 'new') {
-    linkedElement = _getMatchingLinkElement(refs[1], element, commentRefs,
+    result = _getMatchingLinkElement(refs[1], element, commentRefs,
         isConstructor: true);
   } else {
-    linkedElement = _getMatchingLinkElement(reference, element, commentRefs);
+    result = _getMatchingLinkElement(reference, element, commentRefs);
   }
+  final linkedElement = result.element;
```
devoncarew: Please add the type here (and below).

lib/src/markdown_processor.dart
```diff
@@ -114,37 +122,79 @@ ModelElement _getMatchingLinkElement(
...
+    });
+  });
+
+  if (result.isEmpty) {
+    return new MatchingLinkResult(null, null);
+  } else if (result.length == 1) {
+    return new MatchingLinkResult(result.values.first, result.values.first.name);
+  } else {
+    if (_emitWarning) {
+      print("Abiguous reference '${codeRef}' in '${element.fullyQualifiedName}', " +
```
devoncarew: sp: ambiguous

Also, can you get the file and line number here in order to print them out?

devoncarew: Something like: `Ambiguous reference to [foo], lib/my_lib.dart, line 100`.

devoncarew:
OK, looks like `String get fullyQualifiedName` already exists on model elements. Let's try the existing `libName[.enclosingElement].symbolName` and see how that works.

astashov:
@devoncarew I addessed your feedback, thanks! PTAL

lib/src/markdown_processor.dart
```diff
@@ -39,27 +33,19 @@ NodeList<CommentReference> _getCommentRefs(ModelElement modelElement) {
...
     }
   }
   if (modelElement.element.computeNode() is AnnotatedNode) {
-    if ((modelElement.element.computeNode() as AnnotatedNode)
-            .documentationComment !=
-        null) {
-      return (modelElement.element.computeNode() as AnnotatedNode)
-          .documentationComment
-          .references;
+    if ((modelElement.element.computeNode() as AnnotatedNode).documentationComment != null) {
```
devoncarew: We're doing this cast twice inside this if statement (`modelElement.element.computeNode() as AnnotatedNode`) - perhaps create a new local variable to host the AnnotatedNode?

lib/src/markdown_processor.dart
```diff
@@ -130,58 +112,38 @@ MatchingLinkResult _getMatchingLinkElement(
...
-      modelElements.addAll(c.staticMethods);
-      modelElements.addAll(c.staticProperties);
-    });
-
-    modelElements.forEach((modelElement) {
-      if (codeRef == modelElement.name || codeRef == modelElement.fullyQualifiedName) {
-        result[modelElement.fullyQualifiedName] = modelElement;
-      }
-    });
+  package.allModelElements.forEach((modelElement) {
```
devoncarew: Last comment, I promise :)

If we use a Map here, in the event we have multiple matches, the entry in the map could be overridden. So that would hide the fact that the reference is not unique. Perhaps use a List instead?

And we'll be iterating over a lot of elements - I tend to avoid using a closure for that; a for loop will be more performant.

You might do something thuswise:

```dart
List<ModelElement> results = [];

for (ModelElement element in package.allModelElements) {
  if (codeRef.contains('.')) {
    // It's a fully qualified reference.
    if (element.fullyQualifiedName == codeRef) {
      results.add(element);
      break;
    }
  } else {
    if (element.name == codeRef) {
      results.add(element);
    }
  }
}
```

devoncarew:
@astashov, I think there were one or two issues w/ the lookup code - I added a proposal code snippet; everything else looks good.

astashov:
Hmm, it seems like there's one drawback with using `fullyQualifiedName` - it's the same for inherited methods/properties. E.g. tons of classes inherit from `Widget` in Flutter, so they all have `key` property, and all their fully qualified name are `material.Widget.key`. So, I see a lot of warnings in Flutter docs like this:

```
Ambiguous reference to [key] in 'material.State.didUpdateConfig' (/Users/anton/projects/flutter/packages/flutter/lib/src/widgets/framework.dart:785). We found matches to the following elements: 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.DataRow.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.MergeableMaterialItem.key', 'material.Widget.key', 'material.MergeableMaterialItem.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.MergeableMaterialItem.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.TableRow.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key', 'material.Widget.key'
```

Hixie:
Isn't that fine? They can all just point to Widget.key.

Hixie:
Having said that, I don't think we'd want `[key]` to match `material.Widget.key`. I think we just want things in the global scope of their respective libraries to be accessible without qualification.

astashov:
> Isn't that fine? They can all just point to Widget.key.

Oh right, that's actually fine. Sorry :)

> Having said that, I don't think we'd want [key] to match material.Widget.key. I think we just want things in the global scope of their respective libraries to be accessible without qualification.

Umm, now I'm lost :) Could you please elaborate? Why you don't want `key` to match `material.Widget.key`? Because it's a different library?

Hixie:
Because it's not in scope. Conceptually, the way I imagine this working is that you import all the libraries. But even if you did that, in code, `key` wouldn't resolve to `Widget.key`, right? You can only get to `Widget.key` if you qualify it via an instance of the class (or for statics via the class itself).

More practically, if we consider every member of every class to be in scope everywhere, we're going to have conflicts up the wazoo. :-)

astashov:
So, you don't want `[key]` to match `material.Widget.key`, you only want `key` to match `material.key` or something like that? I.e. the top level stuff only? And `Widget.key` to match `material.Widget.key`. Something like that?

Hixie:
That would be my recommendation, yeah.

devoncarew:
Cycling back to this, what @Hixie says makes sense. It sounds like `material.Widget.key` would be the fully qualified way to specify `key`, and `Widget.key` would be legal shorthand for it as long as there wasn't a `Widget.key` in another library you're generating docs for.

Similarly with `material.Widget` and `Widget`.

astashov:
Okay, I've updated the PR. So, from now on, if the reference was, for example, `[key]`, we won't match `material.Widget.key`, we will only match `material.key` (i.e. top-level `key` symbol). If you still want to match `Widget.key`, you should use either `[Widget.key]` or `[material.Widget.key]`.

That's how you want it to work, right?

Hixie:
How do I know what value to use for the library name? Is it (exclusively) the identifiers used with `library` keywords?

astashov:
Yeah, I think so. E.g. `Widget` is in `lib/src/widgets/framework.dart`, which is exported in `lib/widgets.dart`, and that file has `library widgets;` on top, so it will be `widgets.Widget`. The confusing thing that e.g. `lib/material.dart` also exports `lib/widgets.dart`, so `material.Widget` is also a valid reference. I believe there's an issue for that though (https://github.com/dart-lang/dartdoc/issues/1158)

Hixie:
SGTM

devoncarew:
lgtm! Landing...

PR #1299, opened 2016-12-19
adjust @example warning message

chalin:
Make “file missing” warning message more consistent with other warning messages, and add missing newline. Followup to #1295.

cc @filiph @kwalrath


devoncarew:
lgtm!

PR #1300, opened 2016-12-20
Fix bug when parameterized typedef props have wrong return type [#1263]

### LINKED ISSUE #1263: ValueChanged<T> callbacks don't document the T
(from https://github.com/flutter/flutter/issues/6475)

The doc generated for a ValueChanged valued callback doesn't reveal the type of its parameter.

For example see onChanged here: https://docs.flutter.io/flutter/material/Checkbox-class.html

astashov:
@HansMuller noticed, that for properties, that have parameterized
typedef (i.e. with generic parameter) as a return type - they don't show
that parameter in the docs in their return types.

It appears we don't show them, because in that case
`FunctionType#typeFormals` return an empty array. If we use
`FunctionType#typeArguments`, then everything works fine.

After further investigation I figured that if we run dartdoc on the
whole Flutter codebase, `FunctionType#typeFormals` always returns empty
array. So, not sure why it was introduced in the first place.

`git blame` has shown it was introduced here https://github.com/dart-lang/dartdoc/commit/9d96d8dbe24a15af8109a512bb463f235cfbebf5 by @keertip,
as part of upgrade to the latest analyzer (0.27.0).

Though after the changes in this PR all unit tests are still green, and
also I couldn't find anything broken in the Flutter docs after quick
glance, I'm still a bit nervous about this change, because:

* I don't quite understand what's the difference between `typeFormals`
  and `typeArguments`, and in what cases we need to use `typeFormals`.
* I don't understand why that `if` condition with `typeFormals` was
  initially introduced.

So, I'd like to get advice from @keertip in that case, I think.

(fixes #1263)


astashov:
Oooohh, I figured why those `typeFormals` were added - they return the type parameters in a typedef declaration. It **seems** like we can just combine `typeFormals` and `typeArguments`, and that will give us right type information in both cases (when we get the `nameWithGenerics` for a typedef, and also when we get `linkedReturnType`, which is a typedef). But again - not 100% sure. @bwilkerson could you please advice? Thanks!

astashov:
Not sure what to do with it? Maybe merge it? Tests pass, the test docs seem okay...

lib/src/element_type.dart
```diff
@@ -23,7 +23,11 @@ class ElementType {
 
   bool get isParameterizedType {
     if (_type is FunctionType) {
-      return (_type as FunctionType).typeFormals.isNotEmpty;
+      final FunctionType functionType = _type;
+      final Iterable<DartType> typeArguments = []
+          ..addAll(functionType.typeArguments)
+          ..addAll(functionType.typeFormals.map((f) => f.type));
+      return typeArguments.isNotEmpty;
     } else if (_type is ParameterizedType) {
```
keertip: @bwilkerson, do these changes look ok? As mentioned, this fixes the issue with generic info not showing up for typdef return types.

bwilkerson:
Sorry for the delay, for some reason I wasn't getting any notification when I was referenced.

Anyway, @jmesserly or @leafpetersen would be better able to answer this question.

jmesserly:
It depends what you're trying to accomplish. A generic function type, for example: `<T>(T) -> T` will have typeFormals. If we instantiate it with `<int>` we'll get `int -> int` which is a non-generic function, so its typeFormals are empty. But if it happened to have originated from a generic function, then it will have typeArguments that track the `<int>`

lib/src/element_type.dart
jmesserly: This getter (and the one below) looks incorrect to me, but I'm not 100% sure how this getter will be used.

It will break once we have typedefs for generic function types, and it will break for generic function typed parameters (e.g. `foo(T genericFn<T>(T x))`). Mixing typeArguments and typeFormals is never a good idea.

If the goal is to display instantiated typedefs using their `Name<PlusGenericArguments>` then it should be looking at the FunctionType's element and handling the FunctionTypeAliasElement case (for typedefs) and the FunctionTypedFormalParameter (for function typed formals).

jmesserly: More constructively, I would try to fix it like this:

```dart
    bool get isParameterizedType {
      var type = _type; // NOTE: type promotion now works
      if (type is FunctionType) {
        // NOTE: you could factor out a common function if the performance cost
        // of building the list is a concern. It would be my code below but skipping
        // the .map(...).toList() part. 
        return typeArguments.isNotEmpty;
      } // ...
    }
    // ...
    List<ElementType> get typeArguments {
      var type = _type;
      if (type is FunctionType) { 
        List<DartType> typeArguments;
        if (type.element is FunctionTypeAliasElement && type.typeFormals.isEmpty) {
          // TODO(jmesserly): simplify check above; we should have a way
          // to find instantiated typedefs without consulting the element.
          // Also, it will not work if we support typedefs declared inside classes.
          typeArguments = type.typeArguments;
        } else {
          typeArguments = type.typeFormals;
        }
        return typeArguments.map(_getElementTypeFrom).toList();
      } // ...
    }
```

astashov: Wow, thanks a lot for the extended answer, advice and the code snippet! I used it as is, and it works correctly. Thanks!

astashov:
@keertip PTAL. I applied the code snipped, as @jmesserly recommended, the tests pass, the Flutter docs also seem fixed.

jmesserly:
oh that's great news!

keertip:
@jmesserly Thanks!

keertip:
@astashov, looks good. Can you resolve the conflicts?

astashov:
@keertip Done!

PR #1303, opened 2016-12-22
Fix double props when inherited from parameterized class [#1228]

### LINKED ISSUE #1228: Duplicate entries in Flutter docs for RectTween members begin and end
This class shows "begin" and "end" twice:
https://docs.flutter.io/flutter/animation/RectTween-class.html

astashov:
Sometimes, when a subclass inherits a property from a parameterized
class, the property appears twice in the docs.

That happens, because (by some reason, I'm not sure why) the field
objects for those getters and setters have different references
(though they are the same - their names are the same, returning types
are the same, etc). Since `FieldElementImpl` doesn't override default
equality, after we added the field for getter into
`_inheritedProperties`, when handling setter, we compare fields of
getter and setter by reference there:

https://github.com/dart-lang/dartdoc/blob/38a7863ed87892461d6ac4c448a6a719fbe95463/lib/src/model.dart#L479

And usually these are the same object, so we won't see 2 properties in
the generated docs. But in that case by some reason they are 2 different
objects, so we add the field of the setter as well, and then we have 2
same props in the generated docs as a result.

It seems like we can safely compare their names (since there couldn't be
2 properties with the same name in a class anyway). That would fix the
problem.

I've added that in this commit, and also a unit test covering that case.

Fixes #1228


devoncarew:
Thanks! and lgtm

Hixie:
We really should figure out why we're getting two setters and getters. There could be an underlying more serious bug there.

astashov:
It's not 2 setters and getters, it's a setter and a getter have `variable` property, which usually returns the same object for both getter and setter, and the code relied on that. And in this edge case their `variable` property by some reason return different, but equal objects (they have different `hashCode`s).

PR #1307, opened 2016-12-26
Go to the first suggestion in the search field on Enter [#1149]

### LINKED ISSUE #1149: Hitting enter when searching in the docs should go to the top suggestion.  Right now does nothing.
_From @apwilson on April 13, 2016 21:15_
## Steps to Reproduce

Please tell us what you were doing and what went wrong.
## Flutter Doctor

Paste the output of running `flutter doctor` here.
## Logs and Crash Reports

Please attach any relevent logs or crash reports.

> For more information about diagnosing and reporting Flutter bugs, please see [https://flutter.io/bug-reports/](https://flutter.io/bug-reports/).

_Copied from original issue: flutter/flutter#3311_

astashov:
Unfortunately, typeahead.js doesn't give us any API for fetching current
suggesions, so we workaround that by adding `href` of a suggesion into
a data attribute of HTML for that suggestion, and then fetching
suggestions from DOM, choosing the first one, and using it's
`data-href`. Quite dirty though :( I'd prefer to do something like
`typeAheadElement.typeadead('suggestions')[0].href` instead.

Testing: Checked that functionality in latest Chrome, Firefox and Safari

Fixes #1149

### DIFF
--- a/lib/resources/script.js
+++ b/lib/resources/script.js
@@ -172,7 +172,7 @@ function initSearch() {
       templates: {
         suggestion: function(match) {
           return [
-            '<div>',
+            '<div data-href="' + match.href + '">',
               match.name,
               ' ',
               match.type.toLowerCase(),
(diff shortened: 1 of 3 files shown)


devoncarew:
lgtm

PR #1308, opened 2016-12-27
Rank parent class method higher than its overrides in search suggestions

### LINKED ISSUE #896: rank parent class method higher than its overrides
For example, search on "clone". This doesn't return Node.clone() but it does return a bunch of methods that override it. When B overrides A, we should rank A higher.

astashov:
We do that by saving `overriddenDepth` to index.json (which is just
how many overridden elements in the inheritance chain that element has),
and then reducing weight of elements in the suggestions according to
`overriddenDepth`.

Also, fixed `grind update-test-package-docs`, it didn't know where to
find examples for `@examples` directives.

Fixes #896


devoncarew:
Thanks! Looks great.

PR #1309, opened 2016-12-27
small tweak to the qualified reference resolution code

devoncarew:
A small tweak to the qualified reference resolution code - we only look for partially qualified matches if we didn't find a fully qualified one.

@astashov


astashov:
Okay, LGTM

PR #1313, opened 2016-12-29
Add warnings for "free-hanging" generics in docs [#1250]

### LINKED ISSUE #1250: Should recognize generic syntax in plain text in dart docs rather than treating it as HTML
The following comment:

```
/// This class specializes the interpolation of Tween<int> to be
```

...turns into the text "This class specializes the interpolation of Tween to be" because the `<int>` part is treated like HTML markup.

astashov:
@Hixie asked for warnings when the generics are "free-hanging", and not
wrapped into [] block (like `Apple<Cat>`, but not `[Apple<Cat>]`). This
commit adds those warnings. We try to find the tags, which are not
wrapped into square brackets, which are not from the whitelist of
possible HTML tags.

Also, now output all the warnings to the console:

* When there's a missing reference in the comments
* When there're ambiguous reference in the comments
* When there're "free-hanging" generics

But that generates TONS AND TONS of warnings when you e.g. generate Dart
SDK docs or Flutter docs. So, I've added the flag `--show-warnings`,
without it we don't show the warnings.

Fixes #1250

### DIFF
--- a/lib/src/markdown_processor.dart
+++ b/lib/src/markdown_processor.dart
@@ -18,15 +18,15 @@ import 'model.dart';
 import 'reporting.dart';
 
 const validHtmlTags = const [
-  "a", "abbr", "address", "area", "article", "aside", "audio", "b", "base",
-  "bdi", "bdo", "blockquote", "body", "br", "button", "canvas", "caption",
+  "a", "abbr", "address", "area", "article", "aside", "audio", "b",
+  "bdi", "bdo", "blockquote", "br", "button", "canvas", "caption",
   "cite", "code", "col", "colgroup", "data", "datalist", "dd", "del", "dfn",
-  "div", "dl", "dt", "em", "embed", "fieldset", "figcaption", "figure",
-  "footer", "form", "h1", "h2", "h3", "h4", "h5", "h6", "head", "header", "hr",
-  "html", "i", "iframe", "img", "input", "ins", "kbd", "keygen", "label",
+  "div", "dl", "dt", "em", "fieldset", "figcaption", "figure",
+  "footer", "form", "h1", "h2", "h3", "h4", "h5", "h6", "header", "hr",
+  "i", "iframe", "img", "input", "ins", "kbd", "keygen", "label",
   "legend", "li", "link", "main", "map", "mark", "meta", "meter", "nav",
   "noscript", "object", "ol", "optgroup", "option", "output", "p", "param",
-  "pre", "progress", "q", "rb", "rp", "rt", "rtc", "ruby", "s", "samp",
+  "pre", "progress", "q", "s", "samp",
   "script", "section", "select", "small", "source", "span", "strong", "style",
   "sub", "sup", "table", "tbody", "td", "template", "textarea", "tfoot", "th",
   "thead", "time", "title", "tr", "track", "u", "ul", "var", "video", "wbr"
@@ -154,7 +154,7 @@ MatchingLinkResult _findRefElementInLibrary(String codeRef, ModelElement element
     return new MatchingLinkResult(result.values.first, result.values.first.name);
   } else {
     warning(
-        "Ambiguous reference to [${codeRef}] in ${_elementSource(element)}. " +
+        "Ambiguous reference to [${codeRef}] in ${_elementLocation(element)}. " +
             "We found matches to the following elements: ${result.keys.map((k) => "'${k}'").join(", ")}");
     return new MatchingLinkResult(null, null);
   }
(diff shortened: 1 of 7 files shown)


lib/src/markdown_processor.dart
```diff
@@ -14,8 +15,23 @@ import 'package:html/parser.dart' show parse;
 import 'package:markdown/markdown.dart' as md;
 
 import 'model.dart';
-
-const bool _emitWarning = false;
+import 'reporting.dart';
+
+const validHtmlTags = const [
+  "a", "abbr", "address", "area", "article", "aside", "audio", "b", "base",
```
Hixie: we should probably comment out the following tags:
- base
- body
- embed
- head
- html
- rb
- rtc

Where did you get the list?

astashov: From there: http://www.htmldog.com/references/html/tags/

Hixie: https://html.spec.whatwg.org/#elements-3 would be a more authoritative source

lib/src/reporting.dart
```diff
@@ -0,0 +1,11 @@
+library reporting;
+
+import 'config.dart';
+
+void warning(String message) {
+  // TODO: Could handle fatal warnings here, or print to stderr, or remember
+  // that we had at least one warning, and exit with non-null exit code in this case.
+  if (config != null && config.showWarnings) {
+    print("warning: ${message}");
```
devoncarew: We should probably print to stderr here (`stderr.writeln()`).

lib/src/reporting.dart
```diff
@@ -0,0 +1,11 @@
+library reporting;
```
devoncarew: Please add a copyright header (available in the other files).

lib/src/reporting.dart
```diff
@@ -0,0 +1,11 @@
...
+
+import 'config.dart';
+
+void warning(String message) {
+  // TODO: Could handle fatal warnings here, or print to stderr, or remember
+  // that we had at least one warning, and exit with non-null exit code in this case.
+  if (config != null && config.showWarnings) {
+    print("warning: ${message}");
+  }
+}
```
devoncarew: nit: end in an eol

lib/src/markdown_processor.dart
```diff
@@ -137,9 +153,8 @@ MatchingLinkResult _findRefElementInLibrary(String codeRef, ModelElement element
   } else if (result.length == 1) {
     return new MatchingLinkResult(result.values.first, result.values.first.name);
   } else {
-    // TODO: add --fatal-warning, which would make the app crash in case of ambiguous references
-    print(
-        "Ambiguous reference to [${codeRef}] in '${element.fullyQualifiedName}' (${element.sourceFileName}:${element.lineNumber}). " +
+    warning(
```
devoncarew: 👍 to adding a `warning()` method

lib/src/markdown_processor.dart
```diff
@@ -165,23 +180,75 @@ String _linkDocReference(String reference, ModelElement element, NodeList<Commen
...
-    if (_emitWarning) {
-      // TODO: add --fatal-warning, which would make the app crash in case of ambiguous references
-      print("  warning: unresolved doc reference '$reference' (in $element)");
-    }
+    warning("unresolved doc reference '$reference' (in ${_elementSource(element)}");
     return '<code>${HTML_ESCAPE.convert(label)}</code>';
   }
 }
 
+String _elementSource(ModelElement element) {
```
devoncarew: _elementLocation?

lib/src/markdown_processor.dart
```diff
@@ -165,23 +180,75 @@ String _linkDocReference(String reference, ModelElement element, NodeList<Commen
...
 
+// Generics should be wrapped into `[]` blocks, to avoid handling them as HTML tags
+// (like, [Apple<int>]). @Hixie asked for a warning when there's something, that looks
+// like a non HTML tag (a generic?) outside of a `[]` block.
+// https://github.com/dart-lang/dartdoc/issues/1250#issuecomment-269257942
+void _showWarningsForGenericsOutsideSquareBracketsBlocks(String text, [ModelElement element]) {
+  List<int> tagPositions = findFreeHangingGenericsPositions(text);
+  if (tagPositions.isNotEmpty) {
+    tagPositions.forEach((int position) {
+      String errorMessage = "There's a generic type handled as HTML";
```
devoncarew: Perhaps: `Generic type handled as HTML `.

lib/src/markdown_processor.dart
```diff
@@ -165,23 +180,75 @@ String _linkDocReference(String reference, ModelElement element, NodeList<Commen
...
+    });
+  }
+}
+
+List<int> findFreeHangingGenericsPositions(String string) {
+  int currentPosition = 0;
+  int squareBracketsDepth = 0;
+  List<int> results = [];
+  while (true) {
+    final nextOpenBracket = string.indexOf("[", currentPosition);
```
devoncarew: please add a type annotation here (and below)

devoncarew:
Awesome! Some suggested changes (in addition to the one that @Hixie pointed out). Can you also include some sample output (just in a reply to the PR) of it finding warnings?

astashov:
Warnings:

```
warning: Generic type handled as HTML in 'material.AppBar.actions' (/Users/anton/projects/flutter/packages/flutter/lib/src/material/app_bar.dart:205) - 'rld'),
    actions: <Widget>[
      new '
warning: unresolved doc reference 'Scaffold' (in 'material.AppBar.bottomHeight' (/Users/anton/projects/flutter/packages/flutter/lib/src/material/app_bar.dart:null)
warning: Ambiguous reference to [WidgetsBinding.beginFrame] in 'widgets.debugPrintBuildScope' (/Users/anton/projects/flutter/packages/flutter/lib/src/widgets/debug.dart:35). We found matches to the following elements: 'widgets.WidgetsBinding.beginFrame', 'material.WidgetsBinding.beginFrame'
```

astashov:
@devoncarew PTAL

PR #1316, opened 2017-01-04
Fix failing angular2 docs generation [#1315]

### LINKED ISSUE #1315: Crash when building Angular docs
For details, see https://github.com/dart-lang/site-webdev/issues/296.

astashov:
It appears when we link the references in comments, sometimes
`ModelElement element` there could be null. So, need to check it before
showing its location in the warning.

Testing: Generated Angular 2 docs successfully

Fixes #1315


devoncarew:
lgtm, thanks!

PR #1319, opened 2017-01-08
Deemphasize reexported symbols [#1158]

### LINKED ISSUE #1158: In Flutter docs, classes are listed multiple times

astashov:
@Hixie noticed, that when a library exports another library, but both
libraries are being docced together, the result is that each class that
is exported by both libraries is listed twice.

We should still list classes and other symbols in the libraries that
reexport them (since they still part of public API of those libs), but
we only gonna show "canonical" symbols in Search, and also we won't
create duplicate HTML pages for reexported symbols.

We define the reexported symbols in the following way:

1. A defines Foo.
2. Foo doesn't appear in the docs for A because A is a private library,
   i.e., in lib/src/../a.dart
3. B exports Foo from A.
4. Foo appears in the docs for B.
5. C exports Foo from B.
6. Foo appears in the docs for C.

So, Foo of lib C shouldn't have its own page and also shouldn't appear
in Search, because Foo was exported from B and B already has docs. All
links to Foo from HTML docs of C should point to the Foo of B.

To solve that, we build directed "export graph", where we track what
libs export what libs. Then, to find the "canonical" lib, we go up the
graph until we find first documented lib - that's going to be the
canonical one.

Then, we filter out non-canonical libs when building the search index,
and also use canonical lib in `.href` getters, this way ensuring we
always link to symbols in canonical libs and don't create HTML files for
symbols from non-canonical libs.

Testing: Unit tests, also built test docs (`testing/test_package_docs`),
made sure we don't create files for non-canonical stuff (i.e. we create
`fake/Cool` files, but not `ex/Cool`). Also, built Flutter docs, ensured
we don't create `material/BoxDecoration`, `rendering/BoxDecoration` and
`widgets/BoxDecoration`, and only create `painting/BoxDecoration` files,
and all the links point to `painting/BoxDecoration`.

Fixes #1158


lib/src/export_graph.dart
```diff
@@ -0,0 +1,54 @@
+import 'package:analyzer/dart/element/element.dart';
+
```
keertip: Please add comments to methods to explain functionality.

astashov:
@keertip PTAL

PR #1320, opened 2017-01-10
Add macros support [#1264]

### LINKED ISSUE #1264: Ability to insert predefined blobs of text into dartdocs
This PR has a lot of duplication in the dartdocs:
https://github.com/flutter/flutter/pull/6473/files

It's very error-prone and likely to bit-rot. I wish we had a better way to do this.

We want the docs to include this information based on our experience in usability studies.

astashov:
This adds pretty primitive macros support.

Usage:

You define the template first:

    /// {@template foo}
    /// Some content
    /// {@endtemplate}

Then you use it like:

    /// Some stuff
    /// {@macro foo}
    /// More stuff

and it will be displayed as:

    /// Some stuff
    /// Some content
    /// More stuff

You can define the template pretty much in any documentable symbol
comments. So, after we build a `Package` object, we go through all the
comments documentation of all the model elements of the package, and
build the index of macros. It's still quite fast, and doesn't increase
the build times e.g. of the Flutter docs dramatically.

So, after that, when we actually build HTML pages, we already have the
index with templates in `Package`, so we can replace `{@macros` entries
with the corresponding contents from `{@template`

Fixes #1264


sethladd:
Did we document this in the readme? Apologies if I missed it.

astashov:
I didn't document it in README. Maybe we agree on syntax, and then I'll document it?

keertip:
Looks good. Can you document this in the README?

astashov:
@keertip added it to README, PTAL

PR #1324, opened 2017-01-12
Deemphasize and resort the inherited members [#641]

### LINKED ISSUE #641: Option: hide inherited fields / methods

astashov:
@sethladd asked to reorder the list of methods and properties in a way
that the inherited members go last, and also deemphasized in the output
(the names will be nonbold and italic).

This way it will be easier to grasp when skimming through the docs what
are the new methods/fields of a class, but you still can see all of them
and we don't introduce any new togglers/switchers, which would
complicate the UI/UX.

Testing: Generated Flutter docs, made sure the inherited members are
nonbold/italic and go last.

Fixes #641

### DIFF
--- a/lib/resources/styles.css
+++ b/lib/resources/styles.css
@@ -754,3 +754,16 @@ li.inherited a {
   opacity: 0.65;
   font-style: italic;
 }
+
+#instance-methods dt.inherited .name,
+#instance-properties dt.inherited .name,
+#operators dt.inherited .name {
+  font-weight: 300;
+  font-style: italic;
+}
+
+#instance-methods dt.inherited .signature,
+#instance-properties dt.inherited .signature,
+#operators dt.inherited .signature {
+  font-weight: 300;
+}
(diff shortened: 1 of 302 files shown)


sethladd:
From #641 :

Thanks! Looks good. We should also de-bold the return type on the inherited methods. In the example, I see that int is bolder than hashCode, for example. For hashCode → int

astashov:
Ok, done

sethladd:
look and feel lgtm

haven't performed a code review.

Thanks!

astashov:
@keertip PTAL

keertip:
@astashov, changes look good. 

A visual capture of the changes
screen shot 2017-01-12 at 10 50 58 am

PR #1329, opened 2017-01-19
Use only canonical elements when resolving unresolved refs in comments

### LINKED ISSUE #1328: We see ambiguous references when we think we shouldn't
From https://travis-ci.org/flutter/flutter/builds/191793939 (a Travis build for Flutter), using dartdoc 0.9.9

```
Ambiguous reference to [Flexible.fit] in 'widgets.Column' (/home/travis/build/flutter/flutter/packages/flutter/lib/src/widgets/basic.dart:2206). We found matches to the following elements: 'material.Flexible.fit', 'widgets.Flexible.fit'
Ambiguous reference to [Flexible.fit] in 'widgets.Column' (/home/travis/build/flutter/flutter/packages/flutter/lib/src/widgets/basic.dart:2206). We found matches to the following elements: 'material.Flexible.fit', 'widgets.Flexible.fit'
```

We would expect to not see these warnings, because `Flexible.fit` really comes from `widgets.Flexible.fit`

astashov:
If we don't do that, dartdoc will think there's ambiguity if several
libs export the same elements (like in Flutter, e.g. `widgets` exports
`Flexible` and `material` exports `Flexible`). And we won't get links
for these references and will get warnings about ambiguity in
stderr.

Testing: Generated Flutter docs, made sure `fit` from `Flexible` class
resolves correctly and doesn't produce a warning

Fixes #1328

### DIFF
--- a/lib/src/markdown_processor.dart
+++ b/lib/src/markdown_processor.dart
@@ -139,7 +139,7 @@ MatchingLinkResult _findRefElementInLibrary(String codeRef, ModelElement element
   final Package package = library.package;
   final Map<String, ModelElement> result = {};
 
-  for (final modelElement in package.allModelElements) {
+  for (final modelElement in package.allCanonicalModelElements) {
     if (codeRef == modelElement.fullyQualifiedName) {
       result[modelElement.fullyQualifiedName] = modelElement;
     }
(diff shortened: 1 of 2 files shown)


devoncarew:
Adding @keertip to review

brianegan:
Any chance we could bump the priority on this and get a deploy to dartdocs.org? Our docs for RxDart are broken as well due to this bug :(

keertip:
Will publish a version with this fix so that dartdocs.org can be updated. Sorry for the delay.

brianegan:
👍 Thanks for the help @keertip!!

PR #1335, opened 2017-01-31
Search box shouldn't always use the first suggestion

### LINKED ISSUE #1330: Search box doesn't always show me what I click on
1. Go to <https://docs.flutter.io/> and type "BoxPainter" in the search box
2. Press the down arrow three times.  You should have the "paint" method selected.
3. Now press "enter".

Notice that you don't go to the paint method but instead you go to the BoxPainter class.

abarth:
When you select an item from the search box dropdown, we were processing
the event twice: once in typeahead:select and once on keydown. The
second handler was overwriting the window.location generated by the
first and always sending the user to the first suggestion even if they
had selected a different suggestion.

Fixes #1330

### DIFF
--- a/lib/resources/script.js
+++ b/lib/resources/script.js
@@ -189,21 +189,24 @@ function initSearch() {
 
     var typeaheadElement = $('#search-box.typeahead');
     var typeaheadElementParent = typeaheadElement.parent();
+    var selectedSuggestion;
 
     typeaheadElement.on("keydown", function (e) {
       if (e.keyCode === 13) { // Enter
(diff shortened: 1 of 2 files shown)


abarth:
Not sure why Travis is mad.  Seems unrelated to this patch...

devoncarew:
Some of the sdk dartdocs get inlined into the generated docs for our sample project; our build compares the committed samples docs w/ the latest output.

I'll re-gen to fix. Thanks for the patch! Merging...

devoncarew:
`0.9.11` published -

abarth:
Thanks!

PR #1355, opened 2017-03-26
Cleanup and some new features

kevmoo:


kevmoo:
I can pull out the `dartfmt` change.

The test docs fix actually makes travis green – so I'd like to leave that in there.

...and the other test doc updates align w/ the other changes I made.

lib/src/options.dart
```diff
@@ -0,0 +1,17 @@
+class Options {
```
jcollins-g: This doesn't belong in a generic library as it is only applicable to HtmlGenerator.  So it should be defined in html_generator.dart (remember, cyclic imports are allowed in dart) since it is part of the interface to HtmlGenerator, and should be called something less generic, like HtmlGeneratorParameters.

However, I'd prefer we just not do this at all; while shortening the parameter list is a noble goal, we haven't actually achieved that in any meaningful sense.  We've just moved the construction of the argument list around a bit, and since this is the only place and the only object that uses this option field it is really just more readable to have an argument list.

pubspec.yaml
```diff
@@ -19,8 +19,8 @@ dependencies:
   mustache4dart: ^1.0.9
   package_config: '>=0.1.5 <2.0.0'
   path: ^1.3.0
-  quiver: '>=0.18.0 <0.23.0'
-  resource: ^1.1.0
+  quiver: '>=0.18.0 <0.25.0'
```
jcollins-g: you are not actually importing these?

tool/grind.dart
```diff
@@ -208,7 +208,8 @@ updateTestPackageDocs() {
         '../test_package_docs',
         '--example-path-prefix',
         'examples',
-        '--auto-include-dependencies'
+        '--auto-include-dependencies',
+        '--pretty-index-json'
```
jcollins-g: IIRC this should produce a noticeable difference in the index.json in the code review, here.  Why doesn't it?

pubspec.yaml
kevmoo: Both are imported...just ensure we get the latest version

lib/src/options.dart
kevmoo: I thought about the circular import thing – it annoys me, but I understand.

As far as cleanup, I point to the diff in `html_generator`. 25 lines removed.

We were referencing each value 3 times – and offering no value.

Options also allows us to centralize defaults is a much more clean way.

tool/grind.dart
kevmoo: This diff is *huge* so it's collapsed - https://github.com/dart-lang/dartdoc/pull/1355/files#diff-613b594c67e5864627539fcd158f63fc

jcollins-g: Ah.  Still getting used to github, OK.

pubspec.yaml
jcollins-g: quiver is not imported by this package, AFAICT.

pubspec.yaml
```diff
@@ -19,8 +19,8 @@ dependencies:
...
+  resource: ^2.1.2
```
jcollins-g: unless 1.1.0 no longer functions for us, better to leave the version constraint broad.

kevmoo:
lib/src/model.dart:26:import 'package:quiver/core.dart' show hash3;

pubspec.yaml
jcollins-g: Oh, right, sorry.  Was looking at my branch.  This is fine.

lib/src/html/html_generator.dart
```diff
@@ -50,33 +44,40 @@ class HtmlGenerator extends Generator {
...
-    return new HtmlGeneratorInstance(_toolVersion, _url, _templates, package,
-            out, _onFileCreated, _relCanonicalPrefix,
-            faviconPath: faviconPath, useCategories: useCategories)
+    return new HtmlGeneratorInstance(
+            _options, _templates, package, out, _onFileCreated)
         .generate();
   }
 }
+
+class HtmlGeneratorOptions {
```
jcollins-g: This looks better now, I can go with this; it does make things a bit neater in HtmlGenerator and its implementation.

kevmoo:
@jcollins-g can you update your review status?

pubspec.yaml
jcollins-g: Still looking for this to return to 1.1.0.

kevmoo: Sorry. I bumped the min constraint because

1) This is mostly used via the SDK, so I'm not too worried about folks getting a constraint issue here
2) Future updates would have to be careful about not using things that changed/broken w/ the v2 - branch. Trying to avoid the potential blowup...

jcollins-g:
I'd prefer not to speculatively upgrade here.  If resource makes a breaking change in the future, updating this constraint in this way doesn't actually protect us (you would want a ceiling on the version, not updating the base).  If they have already made a breaking change, then it should be clear where the ceiling needs to be and we can add one.  But for now, I think "no action" is correct.

kevmoo:
PTAL

pubspec.lock
```diff
@@ -324,6 +330,6 @@ packages:
       name: yaml
       url: "https://pub.dartlang.org"
     source: hosted
-    version: "2.1.10"
+    version: "2.1.12"
 sdks:
-  dart: ">=1.17.0-dev.6.2 <2.0.0"
+  dart: ">=1.23.0-dev.0.0 <2.0.0"
```
jcollins-g: we shouldn't move the SDK up this high as that's beyond any stable version currently existing.

kevmoo:
PTAL

PR #1422, opened 2017-05-12
update header styles

devoncarew:
- more work towards https://github.com/dart-lang/dartdoc/issues/1372
- borrow the header styles from dartlang.org
- two commits for easier reviewing

@jcollins-g 

screen shot 2017-05-12 at 3 23 58 pm


devoncarew:
(I'll need to add some changelog entries for this work)

lib/src/html/template_data.dart
```diff
@@ -56,12 +56,7 @@ abstract class TemplateData<T extends Documentable> {
 
   Iterable<Subnav> getSubNavItems() => <Subnav>[];
 
-  String _layoutTitle(String name, String kind, bool isDeprecated) {
-    if (kind.isEmpty) kind = '&nbsp;';
-    String str = '<span class="kind">$kind</span>';
-    if (!isDeprecated) return '${str} ${name}';
-    return '${str} <span class="deprecated">$name</span>';
-  }
+  String _layoutTitle(String name, String kind) => '${kind} ${name}';
```
jcollins-g: what does a page for a deprecated identifier look like now?

testing/test_package_docs/ex/Animal-class.html
```diff
@@ -30,7 +30,7 @@
           <ol class="breadcrumbs gt-separated dark hidden-xs">
             <li><a href="index.html">test_package</a></li>
             <li><a href="ex/ex-library.html">ex</a></li>
-            <li class="self-crumb">Animal</li>
+            <li class="self-crumb">enum Animal</li>
```
jcollins-g: I notice we used to use a span class for the kind but don't anymore, is that intentional?

devoncarew: Yeah, the style for this became a no-op.

devoncarew:
> It would be really helpful to have something browseable for this review so I can get a feel for how the new layout works in practice.

The easiest way might be to:
- git fetch origin
- git checkout css_header
- open open testing/test_package_docs/index.html

but I agree, it might be nice to have an easier way to have a staging preview server

lib/src/html/template_data.dart
devoncarew: updated the PR; here's the latest:

screen shot 2017-05-15 at 7 50 49 pm

jcollins-g:
lgtm

PR #1489, opened 2017-08-31
Use highlight js

kevmoo:


kevmoo:
FYI: just pushed an `oops' commit that adds swift, obj-c, and java (for Flutter interop packages)

Will cleanup/rebase once approved

kevmoo:
Thoughts @jcollins-g and @devoncarew ?

lib/resources/github.css
```diff
@@ -0,0 +1,99 @@
+/*
+
+github.com style (c) Vasily Polovnyov <vast@whiteants.net>
```
jcollins-g: Please indicate somewhere in LICENSE/README where this and other new third-party libraries included come from (including URLs) and their licensing.

lib/src/markdown_processor.dart
```diff
@@ -797,7 +797,6 @@ class MarkdownDocument extends md.Document {
       }
 
       bool specifiesLanguage = pre.classes.isNotEmpty;
-      pre.classes.add('prettyprint');
```
jcollins-g: Could you mention a bit more about how this is replaced?

tool/travis.sh
```diff
@@ -57,5 +57,5 @@ else
   (cd testing/test_package_small; dart -c ../../bin/dartdoc.dart)
 
   # Run the tests.
-  pub run grinder test
+  pub run test
```
jcollins-g: Why the change?

tool/grind.dart
```diff
@@ -181,9 +180,9 @@ publish() async {
 
 @Task('Run all the tests.')
 test() {
-  // this is 5 seconds faster than `pub run test`, so
-  // using straight-up VM here
-  return Dart.runAsync('test/all.dart', vmArgs: ['--checked']);
+  // `pub run test` is a bit slower than running an `test_all.dart` script
+  // But it provides more useful output in the case of failures.
```
jcollins-g: Could you be more specific?

testing/test_package/lib/fake.dart
```diff
@@ -496,7 +496,7 @@ class BaseForDocComments {
   ///
   /// Reference to a top-level const in another library [incorrectDocReferenceFromEx]
   ///
-  /// Reference to prefixed-name from another lib [css.theOnlyThingInTheLibrary] xx
+  /// Reference to prefixed-name from another lib [example_prefixed.incorrectDocReferenceFromEx]
```
jcollins-g: Does css.theOnlyThingInTheLibrary still fail?

lib/src/markdown_processor.dart
kevmoo: `prettyprint` was used by the pretty print library to find elements to style at runtime.

`highlight.js` looks at all code blocks with `language-XYZ` – no prettyprint needed

tool/travis.sh
kevmoo: See comments on this commit: https://github.com/dart-lang/dartdoc/pull/1489/commits/90a379d9d5845616cf8b8a050c15bc4fda5e4c89

testing/test_package/lib/fake.dart
kevmoo: That member was removed – but since this test was skipped, it didn't change output.

Just picked something that already existed.

lib/resources/github.css
kevmoo: Done

tool/grind.dart
kevmoo: See comments on this commit: 90a379d

kevmoo:
@jcollins-g here's the before/after

For the original bug, the new style is much more clean about comments vs ID selectors.

Before

screenshot_before

After

screenshot

testing/test_package/lib/fake.dart
jcollins-g: The member isn't actually gone, so this test should pass (assuming we didn't have the bug it is testing for).

kevmoo:
Sorry. Had to do a rebase fix – but went back to the original

test/model_test.dart
```diff
@@ -461,17 +461,10 @@ void main() {
...
-              'Wait for https://github.com/dart-lang/dartdoc/issues/767 to be fixed');
-
-      // remove this test when the above test is fixed. just here to
-      // track when the behavior changes
-      test('codeifies a prefixed top-level variable an imported lib', () {
         expect(
-            docsAsHtml, contains('<code>css.theOnlyThingInTheLibrary</code>'));
+            docsAsHtml,
+            contains(
+                '<p>Reference to prefixed-name from another lib <code>css.theOnlyThingInTheLibrary</code> xx</p>'));
```
jcollins-g: Now this has " xx" on the end?

kevmoo:
@jcollins-g – FYI. I removed the fixed test.

In the test case, the link is still broken....but in the full output, the link is there.

Zero clue why.

PR #1507, opened 2017-09-29
Enable strong-mode, test FutureOr.

### LINKED ISSUE #1482: FutureOr<T> gets interpreted as "dynamic"
Example:
https://www.dartdocs.org/documentation/io/0.2.0/io/isExecutable.html

matanlurey:
Closes #1482.

This is a breaking change, and has been documented as such. Specifically, it looks like we now are using strong-mode inference, so many `dynamic` elements have become non-dynamic. That's probably a good thing?

(`FutureOr` looks great after this)

/cc @kevmoo @sethladd - is enabling `strong-mode` OK here?


kevmoo:
I'm a fan!

matanlurey:
I see two travis failures - one for Flutter and one for the core SDK after this.

* Flutter [fails due to a single test][flutter]:

```
dartdoc:stdout: [error] The list literal type 'List<dynamic>' isn't of expected type 'List<int>'. The list's type can be changed with an explicit generic type argument or by changing the element types. at /home/travis/build/dart-lang/dartdoc/doc/flutter/bin/cache/pkg/sky_engine/lib/typed_data/typed_data.dart, line 396.
```

A link [to this SDK][sdk_typed_data] shows:

```dart
  static final Endianness HOST_ENDIAN =
      (new ByteData.view(new Uint16List.fromList([1]).buffer)).getInt8(0) == 1
          ? LITTLE_ENDIAN
          : BIG_ENDIAN;
```

... it looks kosher? I don't really understand this error.

[sdk_typed_data]: https://github.com/dart-lang/sdk/blob/master/sdk/lib/typed_data/typed_data.dart#L396

[flutter]: https://travis-ci.org/dart-lang/dartdoc/jobs/281156206

* Core SDK: A [failure during grinder][core_sdk]

[core_sdk]: https://travis-ci.org/dart-lang/dartdoc/jobs/281156207

```
Generation failed: 'package:dartdoc/src/model.dart': Failed assertion: line 3038 pos 12: '!name.isEmpty ||
        (this.element is TypeDefiningElement &&
            (this.element as TypeDefiningElement).type.name == "dynamic")': is not true.
dart:core                                                     _AssertionError._throwNew
```

I can fix this assertion to be fine with strong-mode, but I would like @jcollins-g's reaction first.

kevmoo:
🤷‍♂️

kevmoo:
FYI: I just tried running the code on master and got the same assert RE SDK.

jcollins-g:
The assert is fixed in #1506, however there are other problems in the flutter travis that seem to be related to pub that I'm currently debugging.  I'll finish reviewing this once the build is fixed but the general principle looks good.

matanlurey:
Thanks @jcollins-g - not a priority if you're working on something else, just trying to help!

kevmoo:
woo hoo!

matanlurey:
@jcollins-g I still see:

```
dartdoc:stdout: [error] The list literal type 'List<dynamic>' isn't of expected type 'List<int>'. The list's type can be changed with an explicit generic type argument or by changing the element types. at /home/travis/build/dart-lang/dartdoc/doc/flutter/bin/cache/pkg/sky_engine/lib/typed_data/typed_data.dart, line 396.
```

on travis. Anything special I should do? Maybe I'm not synced?

jcollins-g:
There is likely an analysis option that is missing but required to properly analyze Flutter with strong mode enabled.  Dartdoc doesn't obey analysis options files the same way analyzer does, so this change usually involves reading the affected code and any relevant analysis options files, and if possible, adding the option in the same place you just did.

matanlurey:
Sorry for the delay. I just re-merged against master, and tests pass locally.

Let's see if travis likes me 🙏

matanlurey:
:cry: Flutter bot still doesn't like me.

It looks like it is:
https://github.com/dart-lang/sdk/blob/master/sdk/lib/typed_data/typed_data.dart#L396

I don't _think_ it is an additional analysis option, it looks like it flags `INVALID_CAST_LITERAL_LIST`... unless this cause `-implicit-dynamic: false` to trigger (https://github.com/flutter/engine/blob/master/analysis_options.yaml#L12).

jcollins-g:
I've got this mostly working in a different branch now, however there is still one little wart in Flutter -- we generate a new broken link with strong-mode on.

jcollins-g:
This probably should be blocked on #1561.

matanlurey:
Will be closed/replaced by https://github.com/dart-lang/dartdoc/pull/1611!

PR #1518, opened 2017-10-17
Use the logging package for dartdoc output

kevmoo:


kevmoo:
Explicitly *not* using `pkg/logging` so I can send non-Strings to the listener

jcollins-g:
https://github.com/dart-lang/logging/blob/master/lib/logging.dart#L158

Why is this not sufficient?

kevmoo:
Because
https://github.com/dart-lang/logging/blob/50b32dc828d9ad4291d859c8cc6a80ed445cf5b2/lib/logging.dart#L173

jcollins-g:
I'll need more of an explanation than that.  Why do you need to be able to log non-string-convertible objects?

kevmoo:
The idea: push the "text or JSON" decision to the listener. When a component logs it either sends a String and/or something JSON-able.

kevmoo:
@jcollins-g – thoughts?

pubspec.yaml
```diff
@@ -1,6 +1,6 @@
 name: dartdoc
 # Also update the `version` field in lib/dartdoc.dart.
-version: 0.14.1
+version: 0.14.2-dev
```
jcollins-g: add entry to changelog if you like but I prefer to avoid updating the version number for non-breaking changes until release.

bin/dartdoc.dart
```diff
@@ -47,11 +48,9 @@ main(List<String> arguments) async {
 
   final bool sdkDocs = args['sdk-docs'];
 
-  if (args['show-progress']) {
-    _showProgress = true;
-  }
+  final showProgress = args['show-progress'] as bool;
 
-  var readme = args['sdk-readme'];
+  var readme = args['sdk-readme'] as String;
```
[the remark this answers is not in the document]
kevmoo: ack

kevmoo:
I'd have to write more code to extends pkg/logging classes (both Logger and LogRecord). We'd also end up with a leaky abstraction, since I'd also have to introduce my own stream of events that's separate from `Logger`. Or I'd have to implement Logger – and then implement all of the members...

The todo here is to make pkg/logging more extensible. In the mean time, I think this is the cleanest solution.

kevmoo:
@jcollins-g – further thoughts?

jcollins-g:
I still don't think this is urgent enough to develop an entirely new logging module here even as a stopgap -- and I don't see it taking "more code" to extend the logging module to enable JSON logging.  Whether you have a json object or not, the warnings will get transformed to a string and written out, which is what the existing logging module does.  If json objects do not have a "toString()" that does that it's easy enough to add as an extension to logger.

I see no reason to introduce a second stream of events or to reimplement Logger in its entirety.

kevmoo:
@jcollins-g – I understand logging a bit better now.

Trying again – with the logging package. PTAL

pubspec.yaml
jcollins-g: I still think this is poor CM practice but if it is standard for Dart package development, we can do this.  A better way might be to put this in a separate PR.

test/compare_output_test.dart
```diff
@@ -145,7 +145,7 @@ void main() {
       final sep = '.'; // We don't care what the path separator character is
       final firstUnfoundExample = new RegExp('warning: lib${sep}example.dart: '
           '@example file not found.*test_package${sep}dog${sep}food.md');
-      if (!result.stderr.contains(firstUnfoundExample)) {
+      if (!result.stdout.contains(firstUnfoundExample)) {
```
jcollins-g: Do we know why this is now stdout?  Warning prints should go to stderr.

bin/dartdoc.dart
```diff
@@ -123,6 +122,45 @@ main(List<String> arguments) async {
...
+  logging.Logger.root.level = progressLevel;
+  final stopwatch = new Stopwatch()..start();
+
+  // Used to track if we're printing `...` to show progress.
+  // Allows unified new-line tracking
+  var writingProgress = false;
+
+  logging.Logger.root.onRecord.listen((record) {
+    if (record.level == progressLevel) {
+      if (showProgress && stopwatch.elapsed.inMilliseconds > 250) {
```
jcollins-g: This definitely makes the progress ticker less annoying.

test/compare_output_test.dart
kevmoo: Disagree! if dartdoc has a error/crash that should go to stderr

See https://www.jstorimer.com/blogs/workingwithcode/7766119-when-to-use-stderr-instead-of-stdout

stderr should be for diagnostics.

If I pipe the output from dartdoc to a file, I want the hints and the warnings all together.

kevmoo:
> I still think this is poor CM practice but if it is standard for Dart package development, we can do this. A better way might be to put this in a separate PR.

It helps one understand if the package is in a "work in progress" state. `-dev` means between releases. I think it's super helpful.

kevmoo:
@jcollins-g PTAL

kevmoo:
Fixed travis oops – I think

test/compare_output_test.dart
jcollins-g: As discussed offline, this interpretation doesn't follow from the intended purpose of dartdoc, which is not to generate warnings, but documentation.  Given that, the warnings are diagnostic information, and should remain on stderr.

test/compare_output_test.dart
```diff
@@ -145,7 +145,7 @@ void main() {
       final sep = '.'; // We don't care what the path separator character is
       final firstUnfoundExample = new RegExp('warning: lib${sep}example.dart: '
           '@example file not found.*test_package${sep}dog${sep}food.md');
-      if (!result.stdout.contains(firstUnfoundExample)) {
+      if (!result.stderr.contains(firstUnfoundExample)) {
```
kevmoo: @jcollins-g – see here – that change got reverted...

test/compare_output_test.dart
kevmoo: See https://github.com/dart-lang/dartdoc/pull/1518/commits/d8e2a66e452e8425a181bee5d826e873c8c9e60e#r146899745

pubspec.yaml
jcollins-g: I'd like to see future -dev changes made in separate PRs to make reverts easier.

But if we're going to keep this one, the version number in lib/dartdoc.dart should also change.

kevmoo: Happy to do a rebase cleanup...if you have no other comments

kevmoo:
Look okay, @jcollins-g ?

PR #1521, opened 2017-10-18
Extract a FileWriter typedef and make it a parameter to HtmlGeneratorInstance

kevmoo:


kevmoo:
@jcollins-g – ready for review, I think

lib/src/html/html_generator_instance.dart
```diff
@@ -19,29 +19,17 @@ import 'resources.g.dart' as resources;
...
+class HtmlGeneratorInstance {
   final HtmlGeneratorOptions _options;
   final Templates _templates;
   final Package _package;
-  final String _outputDirectoryPath;
   final List<ModelElement> _documentedElements = <ModelElement>[];
-  final StreamController<File> _onFileCreated;
-
-  @override
-  String get relCanonicalPrefix => _options.relCanonicalPrefix;
```
jcollins-g: nice, I had vaguely remember wondering why we were doing this once, good to see it gone

kevmoo:
> This is more than just a cleanup; you've refactored HtmlGeneratorInstance to allow passing in a custom file writer. Can you update the title and description of the PR to describe that?

I was planning to land this as the 3 separate commits, so the history is not lost.

Cool?

lib/src/html/html_generator.dart
```diff
@@ -66,22 +69,61 @@ class HtmlGenerator extends Generator {
...
+        throw new StateError('`write` was called after `generate` completed.');
+      }
+      // If you see this assert, we're probably being called to build non-canonical
+      // docs somehow.  Check data.self.isCanonical and callers for bugs.
+      assert(allowOverwrite || !writtenFiles.contains(filePath));
+
+      var file = new File(p.join(outputDirectoryPath, filePath));
+      var parent = file.parent;
+      if (!parent.existsSync()) {
+        parent.createSync(recursive: true);
```
kevmoo: RE ensuring the directory is created: it also happens here 😄

jcollins-g:
Detailed PR descriptions are really helpful, because that's where I go to compile changelogs and other high level information.

lib/src/html/html_generator.dart
jcollins-g: Seems as though this is the only place it is really needed.

kevmoo:
> Detailed PR descriptions are really helpful, because that's where I go to compile changelogs and other high level information.

Are the commit descriptions sufficient? That's what you'll see once I land...

jcollins-g:
Your commit descriptions don't actually say what's relevant about the change -- besides we should be squashing and merging because that's what we do in this repository.   I'll say what I usually say to folks on how to do this:   In the PR description, write what you did as though six months from now you will be paged at 3am and must look at the description to quickly get a clear idea of why you did this.  Or, another way: Write it so that an engineer familiar with Dart in general, but not necessarily the details of Dartdoc or your plans for it, can understand the rationale for the change without reading the source code.

As an example, for this change, here is a good PR comment:

Title: "Extract a FileWriter typedef and make it a parameter to HtmlGeneratorInstance"

Body: "Make it possible to substitute out different methods for file writing so that we can later write them as rot13-encrypted ASCII protocol buffers.  Also includes some code cleanup."  (which reminds me, what exactly is the next step in what you're doing?)

kevmoo:
Got it! On it! Sorry!

kevmoo:
@jcollins-g – sorry for the confusion.

Here are two commits, with much better commit info – I hope.

I can split this into 2 PRs if you like – I plan to land them as two commits either way

> which reminds me, what exactly is the next step in what you're doing?

To enable a no-op mode so we can use pkg/dartdoc for package analysis/health

jcollins-g:
Thanks.  Re: landing as two commits: this is a squash and merge repository; if you feel strongly about splitting this up into two commits please separate them into two PRs.

jcollins-g:
Approved, don't forget to update the PR description (first comment)

kevmoo:
> if you feel strongly about splitting this up into two commits please separate them into two PRs.

Ah, 💩 – I did a local rebase and push to get both commits.

I'll just do one PR per commit in the future. Sorry...

PR #1523, opened 2017-10-18
Make all markdown content render at 700px

kevmoo:
Fixes https://github.com/dart-lang/dartdoc/issues/1522

Also add test content to validate 80-char code blocks

### DIFF
--- a/lib/resources/styles.css
+++ b/lib/resources/styles.css
@@ -79,10 +79,6 @@ main {
   padding: 20px 15px 0 30px;
 }
 
-.main-content p {
-  max-width: 700px;
-}
-
 .sidebar,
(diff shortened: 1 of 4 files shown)


lib/resources/styles.css
```diff
@@ -79,10 +79,6 @@ main {
   padding: 20px 15px 0 30px;
 }
 
-.main-content p {
```
jcollins-g: are there any consequences to removing this style?

devoncarew: I believe this just applies the max-width property, which is now applied using the `class="markdown desc"` selector below.

kevmoo:
This was my thought...

PR #1562, opened 2017-12-07
Remove unused field and no-op value set

kevmoo:
With the current version of pkg/analyzer (0.31.0-alpha.2),
enableAssertInitializer setter does nothing


jcollins-g:
The incomplete version number change seems to have broken travis.  Rebuild package docs?

kevmoo:
PTAL

PR #1573, opened 2017-12-15
Improve the rendering of inline <code>

kevmoo:
Taking inspiration from GitHub

Before
before

After
after


kevmoo:
Feel free to shash/merge as you like. Not sure what's up w/ appveyor

PR #1574, opened 2017-12-19
upgrade the version of grinder used

devoncarew:
- upgrade package:grinder to 0.8.1 (this version removes deps on a few packages, including package:unscripted, which was not strong and unlikely to ever be so
- tweak the output of the streaming process output in dartdoc's build (to indent the output so it lines up better with grinder's task output)

@jcollins-g


devoncarew:
(closing due to conflicts)

PR #1708, opened 2018-06-07
Add missing static resource.

gspencergoog:
This adds a missing asset to the resources manifest.
I didn't know I needed to run `pub run grinder index-resources` when I added it...

### DIFF
--- a/lib/src/html/resources.g.dart
+++ b/lib/src/html/resources.g.dart
@@ -10,6 +10,7 @@ const List<String> resource_names = const [
   'package:dartdoc/resources/favicon.png',
   'package:dartdoc/resources/github.css',
   'package:dartdoc/resources/highlight.pack.js',
+  'package:dartdoc/resources/play_button.svg',
   'package:dartdoc/resources/readme.md',
   'package:dartdoc/resources/script.js',
   'package:dartdoc/resources/sdk_footer_text.html',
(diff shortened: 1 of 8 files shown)


devoncarew:
@gspencergoog, is this timely? @jcollins-g is out until next wednesday

gspencergoog:
Well, without it, the play buttons will be missing on [animations on the Flutter doc site](https://master-docs-flutter-io.firebaseapp.com/flutter/dart-ui/StrokeJoin-class.html).  There aren't a lot of them yet, but I'd like to fix it.

devoncarew:
ok, gotcha.

I think to get travis green, you'll need to run `grind update-test-package-docs` (as per https://github.com/dart-lang/dartdoc/blob/master/CONTRIBUTING.md).

devoncarew:
> I didn't know I needed to run pub run grinder index-resources when I added it...

We should likely add a travis check for that, or in-line generation into something else, like `update-test-package-docs`. /cc @jcollins-g

gspencergoog:
Yeah, and a Travis check to make sure that the resource manifest doesn't change if you run `grind index-resources` would be good too.

gspencergoog:
Thanks for the review @devoncarew! (I don't have write access, so you'll have to commit it for me... :-))

PR #1730, opened 2018-07-25
Updated deps

kevmoo:


lib/dartdoc.dart
```diff
@@ -22,12 +22,12 @@ import 'package:dartdoc/src/html/html_generator.dart';
 import 'package:dartdoc/src/logging.dart';
 import 'package:dartdoc/src/model.dart';
 import 'package:dartdoc/src/package_meta.dart';
+import 'package:dartdoc/src/tuple.dart';
```
devoncarew: /cc @MichaelRFairhurst for in-lining uses of package:tuple

PR #1761, opened 2018-09-24
tweaks to the category css

devoncarew:
- tweaks to the category css

screen shot 2018-09-24 at 1 31 49 pm

We may also want to reduce the number of colors? Or, use the same color for all categories?

@jcollins-g


jcollins-g:
Hmmm, I tried to do something like this once but it tended to give extra space below the line, which looks a little funny in a mixed list of classes with and without categories.  I take it this version does not do that?  Can you give a screenshot of a single class in a big list so we can compare the line spacing?

jcollins-g:
Also, a screenshot of the class page for BaseForDocComments would be good, as that's a different context for this style that matters.

jcollins-g:
The point of using colors was to make the categories visually distinguishable in pages not necessarily sorted by category.  However, I'd be willing to dump colors entirely if we follow @nex3's suggestion in #1681.

There's sort of a tug of war between designs here:  are categories more like folders, or like tags?  I've implemented them almost entirely as though they were tags, but some sort of hybrid between the two may be more intuitive.

devoncarew:
screen shot 2018-09-24 at 2 35 23 pm

devoncarew:
I think it's harder using color here for different categories, when the color isn't meaningful to the category, but just randomly assigned. Sounds like there may be some iteration about whether categories should be more like tags or folders - whatever falls out from that discussion may inform whether colors should be used or not.

nex3:
I (obviously) would like to use them as folders. My primary use-case for categories is taking large lists of members that are sorted alphabetically by default and providing more semantic groupings for them. Alphabetical sorting is a reasonable default, but it's almost totally useless for locating the member you want--if you know the name, you're probably just going to use the search bar anyway.

devoncarew:
@jcollins-g, I'm having trouble getting the tests to pass; I ran `grind update-test-package-docs`; is there more to do re: golden output files to regen?

jcollins-g:
@devoncarew Greg's upcoming PR has an updated CONTRIBUTING.md with the instructions:

PR #1793, opened 2018-10-12
Add --inject-html flag to enable {@inject-html}, and update README.md

gspencergoog:
This adds a flag `--inject-html` that will enable the `{@inject-html}` directive. Without the flag, the directive is not enabled.

Also, updated the README.md to include information about inject-html.


gspencergoog:
OK, feel free to merge whenever you're ready.

PR #1805, opened 2018-10-17
Fix bug in warnings where not specifying --show-warnings caused dartdoc to succeed even if there were errors.

gspencergoog:
Also added some tests to make sure that external tool failures make dartdoc exit with a non-zero exit code.


gspencergoog:
Hmm. @jcollins-g: looks like this bug has been here a while... The tests are failing because dartdoc is now (correctly!) failing because of errors in the test package.

I think I'll try and move the ones that are supposed to fail into a separate package.

PR #1947, opened 2019-04-02
Adding youtube directive for embedding YouTube videos

goderbauer:
This PR adds an youtube directive that allows embedding YouTube videos. Flutter would like to use this to embed the "Widget of the Week" videos ([example](https://www.youtube.com/watch?v=eI43jkQkrvs)) right into the API docs.

The syntax for adding a YouTube video is:

```
{@youtube 560 315 https://www.youtube.com/watch?v=oHg5SJYRHA0}
```

The integers are the width and height of the video, followed by the YouTube URL of the video. The following YouTube URL format is supported: https://www.youtube.com/watch?v=oHg5SJYRHA0 (as shown in the browser's address bar).

### DIFF
--- a/lib/src/model.dart
+++ b/lib/src/model.dart
@@ -4083,7 +4083,7 @@ abstract class ModelElement extends Canonicalization
 
     // Matches YouTube IDs from supported YouTube URLs.
     final RegExp validYouTubeUrlRegExp =
-        new RegExp('https://www\.youtube\.com/watch\\?v=(.+)');
+        new RegExp('https://www\.youtube\.com/watch\\?v=([^&]+)\$');
 
     return rawDocs.replaceAllMapped(basicAnimationRegExp, (basicMatch) {
       final ArgParser parser = new ArgParser();
@@ -4131,7 +4131,7 @@ abstract class ModelElement extends Canonicalization
       return '''
 
 <p>
-  <iframe src="https://www.youtube.com/embed/$youTubeId"
+  <iframe src="https://www.youtube.com/embed/$youTubeId?rel=0"
           width="$width"
           height="$height"
           frameborder="0"
--- a/lib/src/model.dart
+++ b/lib/src/model.dart
@@ -3270,6 +3270,7 @@ abstract class ModelElement extends Canonicalization
       _rawDocs = documentationComment ?? '';
       _rawDocs = stripComments(_rawDocs) ?? '';
       _rawDocs = _injectExamples(_rawDocs);
+      _rawDocs = _injectYouTube(_rawDocs);
       _rawDocs = _injectAnimations(_rawDocs);
       _rawDocs = _stripHtmlAndAddToIndex(_rawDocs);
     }
@@ -3291,6 +3292,7 @@ abstract class ModelElement extends Canonicalization
       // Must evaluate tools first, in case they insert any other directives.
       _rawDocs = await _evaluateTools(_rawDocs);
       _rawDocs = _injectExamples(_rawDocs);
+      _rawDocs = _injectYouTube(_rawDocs);
       _rawDocs = _injectAnimations(_rawDocs);
       _rawDocs = _stripMacroTemplatesAndAddToIndex(_rawDocs);
       _rawDocs = _stripHtmlAndAddToIndex(_rawDocs);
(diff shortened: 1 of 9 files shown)


devoncarew:
Thanks for the PR! @gspencergoog is the best one to review this PR; in the meantime:

What is the expected behavior of the `{@youtube...` directive in IDEs? W/o any additional work, VS Code and IntelliJ will show the raw directive in tooltip hovers. I'd like to get a better sense of how we want this to roll out throughout the ecosystem. Do we want:

- to show raw `{@youtube 100 100 https://www.youtube.com/watch?v=oHg5SJYRHA0}` text in hovers?
- to completely remove anything between the curly braces?
- to specify a fallback for other, non dartdoc cli renderers? (so, something like `{@youtube 100 100 https://www.youtube.com/watch?v=oHg5SJYRHA0 alt="href src='..."}`)

goderbauer:
@devoncarew what do you currently do in IDEs for the existing `{@animation ...}` directive for embedding mp4 videos? I’d imagine you’d want to give the `{@youtube ...}` directive the same treatment.

Why would you need in `alt` attribute? Couldn’t you just use the video URL already provided in the directive and present it in an appropriate way (e.g. as a link people can click on to get to the video if embedding it is not an option)? What additional information would you need?

DanTup:
> what do you currently do in IDEs for the existing {@animation ...} directive for embedding mp4 videos? I’d imagine you’d want to give the {@youtube ...} directive the same treatment.

Currently we do nothing, though it's not something we're happy with right now - the docs in the IDEs are not as good as on the web (esp. with macros etc.).

I don't there's a good general solution, but I think in this particular instance showing some text with a link to the video could be better ("For an example see (youtubelink)").

goderbauer:
Where are the docs currently shown in the IDE? And what are the limitations there? Embedding the video directly would probably be the best choice. If that doesn’t work: we could probably add a title property to the YouTube directive. You could use the title to render a link (e.g. “Video: ClipRRect as Widget of the Week”) instead of the directive that opens the video when the user clicks on it.

DanTup:
For VS Code, they appear in hovers and also can appear in the help alongside the item in the completion list (though usually we trim this short so it's less likely there).

Screenshot 2019-04-02 at 10 25 52 am

We can display *limited* Markdown here, for ex we have code blocks and links (but we can't do videos or arbitrary HTML):

Screenshot 2019-04-02 at 10 26 00 am

The capabilities may vary across clients though. I don't think there's going to be a perfect solution, we might end up just having to strip this (or reduce to basic links, etc.) to avoid the in-editor docs starting to look like cryptic as the user of DartDoc increases.

lib/src/model.dart
```diff
@@ -4054,6 +4056,109 @@ abstract class ModelElement extends Canonicalization
...
+        warn(PackageWarning.invalidParameter,
+            message: 'Invalid @youtube directive, "${basicMatch[0]}"\n'
+                'YouTube directives must be of the form "{@youtube WIDTH '
+                'HEIGHT URL}"');
+        return '';
+      }
+
+      int width;
+      try {
+        width = int.parse(positionalArgs[0]);
```
gspencergoog: Maybe check width and height to make sure they're both greater than zero, and under 10,000 or so?

goderbauer:
I considered using just the video id. However, if you see

```
{@youtube 560 315 https://www.youtube.com/watch?v=oHg5SJYRHA0}
```

in an environment that doesn't process the directive nicely (e.g. the raw code on github or in the IDE) you know immediately where to go to see what video gets embedded. If it would just show

```
{@youtube 560 315 oHg5SJYRHA0}
```

you're on your own in figuring out how to turn turn that id into a link.

Long story short, I believe showing the full URL is more user friendly and doesn't really cost us anything to parse out the ID.

gspencergoog:
Yeah, I agree that it should be able to take the link, but I was thinking that adding the ability to also just take the ID would be OK.  You make a good point, though, so don't bother adding it.

lib/src/model.dart
```diff
@@ -4054,6 +4056,109 @@ abstract class ModelElement extends Canonicalization
     }
   }
 
+  /// Replace &#123;@youtube ...&#125; in API comments with some HTML to embed
```
Hixie: What is "`&#123;`"?

Hixie: Oh is that `{`? hm...
Does it work (not get interpreted as magic) if you put `{` in backticks?

lib/src/model.dart
```diff
@@ -4054,6 +4056,109 @@ abstract class ModelElement extends Canonicalization
...
+  /// The width and height must be integers specifying the dimensions of the
+  /// video in pixels.
+  ///
+  /// The following YouTube URL formats are supported:
+  ///
+  ///  * https://www.youtube.com/watch?v=oHg5SJYRHA0 (as shown in the browser's
+  ///    address bar)
+  ///  * https://youtu.be/oHg5SJYRHA0 (as shown in the "Share" dialog)
+  ///  * https://www.youtube.com/embed/oHg5SJYRHA0 (as shown in the "Share ->
+  ///    Embed" dialog)
```
Hixie: we should standardise on one to keep things consistent

lib/src/model.dart
gspencergoog: Open curly brace.  It needs that so that dartdoc doesn't try to interpret the dartdoc directives in its own docs.

eseidelGoogle:
YT even has shorter urls if needed: https://youtu.be/oyy_1CjNdBU

sfshaza2:
That's a link URL, Eric. Be we want embedded, right? And I'd like the version that won't suggest videos when it ends.

devoncarew:
Having `{@youtube 100 100 https://www.youtube.com/watch?v=oHg5SJYRHA0}` render as a link in IDE hover text may make the most sense.

I wonder if we can do this in a way where IDEs don't need to know about the `n` different dartdoc directives. If there's a directive they know about - line `@macro` they can process them. If there's a directive they don't know about, they could look for some fallback text for rendering. This would be an `alt` property which all directives could optionally support; or, and combination of an optional `text=` property and `url=` property if we wanted to render as a link.

lib/src/model.dart
goderbauer: Sure, you want to offer an opinion on what the one URL format should be that we support? I'd vote for either https://www.youtube.com/watch?v=oHg5SJYRHA0 or https://youtu.be/oHg5SJYRHA0. The former is easier accessible for people wanting to embed something (just copy the URL from the address bar), the second one looks cleaner.

lib/src/model.dart
goderbauer: Added a check for greater than zero. Imposing an upper limit seems really arbitrary, so I omitted that one.

lib/src/model.dart
goderbauer: Settled on https://www.youtube.com/watch?v=oHg5SJYRHA0.

goderbauer:
> Having `{@youtube 100 100 https://www.youtube.com/watch?v=oHg5SJYRHA0}` render as a link in IDE hover text may make the most sense.

Agreed.

lib/src/model.dart
```diff
@@ -4054,6 +4056,107 @@ abstract class ModelElement extends Canonicalization
...
+  ///
+  ///     &#123;@youtube 560 315 https://www.youtube.com/watch?v=oHg5SJYRHA0&#125;
+  ///
+  /// Which will embed a YouTube player into the page that plays the specified
+  /// video.
+  ///
+  /// The width and height must be positive integers specifying the dimensions
+  /// of the video in pixels. The height and width are used to calculate the
+  /// aspect ratio of the video; the video is always rendered to take up all
+  /// available horizontal space.
```
gspencergoog: You might mention why: "...because desktop and mobile require different sizes." (or some such)

goderbauer: Done.

goderbauer:
@devoncarew I verified that this change rolls cleanly into flutter/flutter. Whenever the tests are done running this should be ready for submission and publication. Thanks!

lib/src/model.dart
```diff
@@ -4054,6 +4056,108 @@ abstract class ModelElement extends Canonicalization
...
+  /// desktop and mobile.
+  ///
+  /// The video URL must have the following format:
+  /// https://www.youtube.com/watch?v=oHg5SJYRHA0. This format can usually be
+  /// found in the address bar of the browser when viewing a YouTube video.
+  String _injectYouTube(String rawDocs) {
+    // Matches all youtube directives (even some invalid ones). This is so
+    // we can give good error messages if the directive is malformed, instead of
+    // just silently emitting it as-is.
+    final RegExp basicAnimationRegExp = new RegExp(r'''{@youtube\s+([^}]+)}''');
```
devoncarew: I'm curious - are this (and `validYouTubeUrlRegExp` below)  constructed each time this method is entered? If so, we should move these to being static fields.

devoncarew:
> I verified that this change rolls cleanly into flutter/flutter. Whenever the tests are done running this should be ready for submission and publication. Thanks!

Great - I'll merge this, and publish a new version of the package tomorrow.

DanTup:
@devoncarew 

> [IDEs] could look for some fallback text for rendering. This could be an alt property which all directives could optionally support; or, a combination of optional text / url properties if we wanted to render as a link.

This sounds reasonable - you mean something like:

```dart
/// {@ youtube 100 100 url="https://www.youtube.com/watch?v=oHg5SJYRHA0"
///     text="Everyone's favourite YouTube video" }
```

?

We could (generally) regex these out in the editor and replace with something like:

```dart
/// See [Everyone's favourite YouTube Video](https://www.youtube.com/watch?v=oHg5SJYRHA0)
```

PR #1950, opened 2019-04-05
Fix iOS scroll physics

goderbauer:
`-webkit-overflow-scrolling` doesn't seem to be a pseudo selector. Instead, it has to be applied directly to the scrolling container (or inherited to it from a parent as in this fix).

Additionally, the `min-height: 100vh` is causing trouble on iOS because on this platform Safari is actually reporting a viewport size that's larger than what's visible. Despite what one may think, this statement therefore caused the dartdocs page to be bigger than what's visible, which is introducing extra awkward scrollbars on the whole page. Removing the statement fixes the issue without breaking anything else.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 0.28.3+1
+* Fix scroll physics and behavior for Safari on iOS.
+
 ## 0.28.3
 * Support a new `{@youtube}` directive in documentation comments to embed
   YouTube videos.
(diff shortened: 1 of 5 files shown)


devoncarew:
The changes look good. I restarted the travis job as the failure looked like trouble accessing pub.dartlang.org.

goderbauer:
Thanks. Looks like it's all green now! 🎉

PR #1953, opened 2019-04-09
Support the latest version of pkg:html

kevmoo:


kevmoo:
> lgtm. I'd thought that we checked the pubspec.lock file in for this repo, but looking at the repo source, apparently not.

...which for a publish package is the right thing to do!

PR #1954, opened 2019-04-16
Bump markdown dependency

srawlins:
This  markdown release properly escapes Fenced Code Block info strings.


srawlins:
Should I bump the version of dartdoc in the same PR?

devoncarew:
Looks like we have not yet published `0.28.3+2`, so just updating the changelog should be enough.

srawlins:
Done!

PR #1961, opened 2019-05-15
Fix inconsistent indentation for inheritance info - Fixes #1687

### LINKED ISSUE #1687: Inconsistent indentation on class inheritance/implements info
This may be intended, but the `Inheritance` list is indented further than `Implements` and `Implemented by`. Indentation is used sparingly on other elements, so I assumed the `Inheritance` list is supposed to match the others.

image

parlough:
Standardizes the indentation for inheritance info similar to other fields such as "Implementers" and "Mixed in types". Fixes https://github.com/dart-lang/dartdoc/issues/1687

If this is not desired design, that's fine. It just seems that elsewhere in the docs uses indentation like this sparingly, and it seems to just be because of forgetting to handle the first item in the list's padding separately.

Original:

Screen Shot 2019-05-15 at 12 56 59

Changed:

Screen Shot 2019-05-15 at 12 57 06


lib/resources/styles.css
```diff
@@ -563,6 +563,7 @@ footer a, footer a:hover {
   background-image: none;
   content: "";
   margin: 0;
+  padding: 0 0 0 0;
```
devoncarew: This can collapse to `padding: 0;`.

PR #1964, opened 2019-05-16
re-update highlight.js

devoncarew:
@kevmoo @keertip


lib/resources/readme.md
```diff
@@ -11,9 +11,7 @@ Included languages:
 * javascript
 * json
 * markdown
-* objectivec
```
kevmoo: Intentional? My understanding is that we had these so Flutter code snippets show up correctly...

devoncarew: I assumed this was unused. 

As per https://github.com/dart-lang/dartdoc/issues/1963, it would be nice if we had some automation / process around the upgrade.

devoncarew: Updated to include obj-c and kotlin.

kevmoo:
CC @jonasfj @isoos – RE syncing w/ pub site

Issue #1980, opened 2019-06-10
Travis failing for a week - failed assert


kevmoo:
See https://travis-ci.org/dart-lang/dartdoc/builds

CC @devoncarew

devoncarew:
Thanks! From talking locally, we think this is a change in the Flutter framework (that's breaking our flutter integration test).

cc @keertip

keertip:
Yes, am looking into it.

keertip:
This change in flutter is what caused the breakage - https://github.com/flutter/flutter/commit/90500a5dc4.

kevmoo:
Ah! Wonder if DartDoc is not happy w/ conditional imports.

keertip:
Yes, the conditional imports look to be the issue. The doc generation is working for the flutter folks at head, no complaints there. It's just the test. The test uses a sync method to build docs and has an assert to check for certain conditions - which are now failing.

Now that there is a likely reason, can work with a test case.

PR #1984, opened 2019-06-22
Support top level declarations with templates

keertip:


keertip:
fixes #1980

lib/src/model.dart
```diff
@@ -4911,6 +4911,10 @@ class PackageGraph {
         if (needsPrecacheRegExp.hasMatch(d.documentationComment) &&
             !precachedElements.contains(d)) {
           precachedElements.add(d);
+          if (m is TopLevelVariable) {
```
devoncarew: Can you add a comment explaining why `TopLevelVariable`s are special cased here?

devoncarew:
> fixes #1980

🎉 great to have the build fixed!

lib/src/model.dart
keertip: done

PR #1985, opened 2019-06-24
Sanitize HTML output of user-provided markdown content.

isoos:
Part of https://github.com/dart-lang/pub-dev/issues/1418

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 * Support the latest version of `package:analyzer`.
 * Fix hyperlinks to overriden methods (#1994).
 * Option in dartdoc_options.yaml to exclude version in footer info (#1982).
+* Sanitize HTML output of user-provided markdown content.
 
 ## 0.28.4
 * **Breaking change** Change the default for `allow-tools` command line flag to false.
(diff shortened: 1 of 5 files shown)


isoos:
Part of the CI failure was the version change, which I've reverted as I have no idea what needs to be done to do it.

jonasfj:
Looks more like we're failing:
> dart2-model_test.dart: 03:42 +8 -1: HTML Injection when allowed can inject HTML [E]
> dart2-model_test.dart: 03:42 +8 -2: HTML Injection when allowed can inject HTML from tool [E]

isoos:
When I run the test locally, I've got a simple `documentation` String in one of the failing tests, but when it is run on CI, it gets a different String. How does the testing of this work? Do I need to setup some kind of local environment upfront?

devoncarew:
Is this PR still active?

isoos:
@devoncarew I'd need more insight and help on why the testing fails, but I'd be happy to fix the failing tests if it is possible to do easily.

keertip:
@isoos , can you rebase your PR?

The test that fails is the inject HTML using tool test - https://github.com/dart-lang/dartdoc/blob/master/test/model_test.dart#L334., and the code under test is https://github.com/dart-lang/dartdoc/blob/master/testing/test_package/lib/example.dart#L601. What might be happening is that the html is sanitized  before the injection is done, and so loses the information needed to successfully inject the html.

isoos:
@keertip: rebased. Any idea how I should update the test to make them pass?

Issue #1940, opened 2019-03-05
allow-tools should default to false


sigurdm:
I think that having dartdoc run arbitrary code on a default invocation is risky and for me seems like a surprising behavior.

I suggest to change the default here:

https://github.com/dart-lang/dartdoc/blob/5a4308df5216707d5de75b00a27b470fbcf9bba8/lib/src/dartdoc_options.dart#L1380

devoncarew:
This seems like a reasonable change to me.

I believe that @gspencergoog added this feature; pinging him for an opinion -

gspencergoog:
No, it was @jcollins-g that added the flag, but I reviewed the PR.

It's not exactly "arbitrary code" though: you have to run Dartdoc with a configuration file that includes the path to the executable you want to run, and it will only run tools configured in the configuration file. The "name" of the tool to run is a key into this configuration file, not a path to (or even necessarily the name of) an executable.

We could make it default to false, but that doesn't actually give you any more security: the same configuration file could be used to set it to true, and also set the executable path.

sigurdm:
If you download a third-party package and run dartdoc on it, the dartdoc_options.yaml file of that library is used?
I would not expect people to inspect the dartdoc_options file before running dartdoc.

But the second point is really good. Changing the default is not enough (cc @jonasfj @isoos ) - it is (IMO) questionable if we should allow enabling this from the dartdoc_options file itself.

keertip:
How about if we make it such that allow-tools cannot be configured from the dartdoc_options file. allow-tools defaults to false and the only way to run it on is to explicitly pass it in on command line.

This will be a breaking change.  @jonasfj @isoos are there other packages apart from flutter using this? Do we know?

jonasfj:
I'm not familiar with all users of dartdoc... Maybe the dart-sdk uses it, athom might know if the Dart release process uses it.

We obviously don't use it on pub.dev :)

keertip:
The SDK does not use it. if pub.dev does not use it, then it's just flutter.

PR #1986, opened 2019-06-25
Change command line flag allow-tools default to false

keertip:


CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 0.28.4-dev
+* **Breaking change** The `allow-tools` command line flag defaults to false.
```
devoncarew: `Change the default for the '--allow-tools' command-line flag to false.`?

keertip:
Yes, will update the flutter script before publishing a new version. Flutter does pull in only published versions, right?

keertip:
#1940

devoncarew:
> Yes, will update the flutter script before publishing a new version. Flutter does pull in only published versions, right?

Yes, and I believe they use only explicit versions (the version number is referenced in the script).

sigurdm:
Does this also disallow enabling `--allow-tools` from the `analysis_options.yaml` file as discussed in #1940?

devoncarew:
It doesn't, but it should like that is something we'd like to do. I'll open an issue to track; PRs welcome for it as well.

keertip:
The only way to set `--allow-tools` is through command line. It cannot be set thru the dartdoc_options file.

sigurdm:
Ah - that was unclear to me. Thanks!

PR #1989, opened 2019-07-10
Prepare to publish 0.28.4

bwilkerson:


devoncarew:
Ah, hold on; I think you'll need to run a script locally to re-gen some files w/ the new version.

devoncarew:
If you run `dart tool/grind.dart build-test-package-docs`, do you see local git changes?

bwilkerson:
No, there were no changed files. But then I would hope there wouldn't be; I didn't change any semantics.

devoncarew:
I think the version number gets written into the files, and we have some generated test lib docs checked in.

From the CI results, try `grind build` instead.

bwilkerson:
Yep, I found that too, and now those two files are updated.

PR #1996, opened 2019-08-01
Add configuration option for output format

jdkoren:
Adds a configuartion option to specify the output format for generated
docs, defaulting to 'html'. Hidden from help usage for the time being.

### DIFF
--- a/lib/src/dartdoc_options.dart
+++ b/lib/src/dartdoc_options.dart
@@ -1422,6 +1422,8 @@ class DartdocOptionContext extends DartdocOptionContextBase
 
   bool isPackageExcluded(String name) =>
       excludePackages.any((pattern) => name == pattern);
+
+  String get format => optionSet['format'].valueAt(context);
 }
 
 /// Instantiate dartdoc's configuration file and options parser with the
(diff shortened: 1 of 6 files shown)


devoncarew:
Can you elaborate on this PR a bit? Or, describe anticipated future output formats? I see you've wired up both `html` and `md`.

Adding @keertip for review and thoughts.

devoncarew:
Ah, I see https://github.com/dart-lang/dartdoc/pull/1997 is a follow-up PR.

lib/src/dartdoc_options.dart
```diff
@@ -1624,6 +1626,8 @@ Future<List<DartdocOption>> createDartdocOptions() async {
             'exist. Executables for different platforms are specified by '
             'giving the platform name as a key, and a list of strings as the '
             'command.'),
+    // TODO(jdkoren): unhide when this feature is in working order.
+    new DartdocOptionArgFile<String>('format', 'html', hide: true)
```
keertip: Does this have to be both command line and options file? Just have a command line flag only?

jdkoren: Done

PR #2002, opened 2019-08-19
Use a smaller set of github extensions for markdown

keertip:


lib/src/markdown_processor.dart
```diff
@@ -142,7 +142,7 @@ final HtmlEscape htmlEscape = const HtmlEscape(HtmlEscapeMode.element);
 final List<md.InlineSyntax> _markdown_syntaxes = [
   new _InlineCodeSyntax(),
   new _AutolinkWithoutScheme()
-]..addAll(md.ExtensionSet.gitHubWeb.inlineSyntaxes);
+]..addAll(md.ExtensionSet.gitHubFlavored.inlineSyntaxes);
```
devoncarew: I'm curious, what is the list of `gitHubFlavored`  extensions, and, should we consider in-lining the list here, so that we don't accidentally get a broader than intended supported set with a future upgrade of the markdown package?

keertip: Changed, inlined the list.

PR #2004, opened 2019-08-21
Lint fixes

kevmoo:


keertip:
failed: The following generated files needed to be rebuilt:
    lib\src\html\resources.g.dart
  Rebuild them with "grind build" and check the results in.

kevmoo:
Fixed the builder!

PR #2006, opened 2019-08-21
Support custom templates directory

jdkoren:
Adds a (hidden) command-line flag to supply a directory with custom
html templates to use. If not specified, dartdoc defaults to using the
packaged templates.

Change-Id: I4148ff388947b1e0e24aa63b598be00cb01ae220

### DIFF
--- a/lib/src/dartdoc_options.dart
+++ b/lib/src/dartdoc_options.dart
@@ -1417,6 +1417,9 @@ class DartdocOptionContext extends DartdocOptionContextBase
 
   bool isPackageExcluded(String name) =>
       excludePackages.any((pattern) => name == pattern);
+
+  String get templatesDir =>
+      optionSet['templatesDir'].valueAt(context);
 }
 
 /// Instantiate dartdoc's configuration file and options parser with the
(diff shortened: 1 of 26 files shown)


lib/src/dartdoc_options.dart
```diff
@@ -1617,6 +1620,13 @@ Future<List<DartdocOption>> createDartdocOptions() async {
             'exist. Executables for different platforms are specified by '
             'giving the platform name as a key, and a list of strings as the '
             'command.'),
+    DartdocOptionArgOnly<String>("templatesDir", null, isDir: true, mustExist: true, hide: true,
+        help: 'Path to a directory containing templates to use instead of the default ones. '
+            'Directory must contain an html file for each of the following: 404error, category, '
+            'class, constant, constructor, enum, function, index, library, method, mixin, '
+            'property, top_level_constant, top_level_property, typedef. Partial templates are '
+            'supported; they must begin with an underscore, and references to them must omit the '
+            'leading underscore (e.g. use {{>foo}} to reference the partial template _foo.html).'),
```
keertip: It would be good to have a check to see if all the expected templates are in the directory, and fail with the a message that lets the user know what's missing. The template files can change (working on one such change now), and it will be helpful to flag this with a check rather than a failure/stack trace.

jdkoren: Added a check (and test), but let me know if you want me to go about it in a different way.

lib/src/html/templates.dart
```diff
@@ -36,50 +37,86 @@ const _partials = <String>[
...
+  var partials = await templatesLoader.loadPartials();
 
-    if (templatePath.contains('_head')) {
-      String headerValue =
-          headerPaths.map((path) => File(path).readAsStringSync()).join('\n');
-      template = template.replaceAll(headerPlaceholder, headerValue);
+  void replacePlaceholder(String key, String placeholder, List<String> paths) {
+    var template = partials[key];
+    if (template != null && paths != null && paths.isNotEmpty) {
+      String replacement = paths.map((path) => File(path).readAsStringSync())
```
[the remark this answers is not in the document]
jdkoren: Done

Issue #1982, opened 2019-06-20
Make emitting version in footer optional


dnfield:
Flutter builds dartdocs on every merge commit, which are then uploaded to our hosting service.  The service is smart enough to deal with files that have not changed differently and not require us to upload them - however, because the version code is in every HTML file, the all (nearly 35k of them) get marked as needing an update, which results in lots of additional network traffic and CI time/flakiness.

I'd like to have an option on dartdoc that allows me to omit this so I can add it in some other way (e.g. by including a JavaScript file that does it).  e.g., we an option to guard the whole `<footer>` element in `_footer.html`, or at least to guard the inclusion of the version string in it.

devoncarew:
The current Flutter footer looks like:

```
Flutter 1.5.4-hotfix.2 • 2019-05-07 18:27 • 7a4c33425d • stable
```

Flutter does customize the footer in order to insert most of that info:

- https://github.com/flutter/flutter/blob/master/dev/tools/dartdoc.dart#L142
- https://github.com/flutter/flutter/blob/master/dev/tools/dartdoc.dart#L267

Stabilizing the files between commits may be as simple as removing the custom footer? Or, removing the date and commit hash at least.

dnfield:
The footer by default will look like

```
Flutter 1.5.4-hotfix.2
```

Even if we specify no options.  That's done here:

https://github.com/dart-lang/dartdoc/blob/master/lib/templates/_footer.html#L5

devoncarew:
Yes, agreed. Will that version (`1.5.4-hotfix.2`) change significantly between commits? I'm assuming that it'll be relatively stable.

dnfield:
On master it changes every commit.  See https://master-api.flutter.dev

keertip:
@dnfield, just to be clear, you would like an option to not include the default footer text (at the least the version no) in the generated docs?

dnfield:
Yes.  Right now I'm cheating it with javascript.

The problem this causes for us is that *every* file gets marked as dirty when we try to update docs on every commit, when we really just want to update the docs for things that actually have changed.

PR #2010, opened 2019-08-22
 Add an option to exclude version info from footer

keertip:
- fix https://github.com/dart-lang/dartdoc/issues/1982


devoncarew:
@dnfield, I assume that this solution - a flag to have dartdoc exclude version info - would address your use case?

dnfield:
Yes, this should work.  Thanks!

PR #2025, opened 2019-09-16
Don't create model if library cannot be found

keertip:


devoncarew:
Is this to address a bug - can you link to that for context?

keertip:
Fixes exception in #2023

keertip:
Closing out this PR, these are not the right checks to catch the exception.

Issue #2017, opened 2019-09-04
Latest analyzer in Dart SDK breaks doc generation


keertip:
analyzer no longer has elements for FunctionType (sometimes), so dartdoc should not use elements when getting info for function types.

PR #2026, opened 2019-09-18
Enable using summary2 for analysis

keertip:
Refactor to use summary2 as sometimes FunctionType's are not backed by elements, we have use the type to get the information.

### DIFF
--- a/lib/src/element_type.dart
+++ b/lib/src/element_type.dart
@@ -334,16 +334,17 @@ abstract class CallableElementTypeMixin implements ParameterizedElementType {
   // TODO(jcollins-g): Rewrite this and improve object model so this doesn't
   // require type checking everywhere.
   Iterable<ElementType> get typeArguments {
-    DefinedElementType elementType = returnedFrom as DefinedElementType;
     if (_typeArguments == null) {
       Iterable<DartType> dartTypeArguments;
       if (type.typeFormals.isEmpty &&
           element is! ModelFunctionAnonymous &&
-          elementType?.element is! ModelFunctionAnonymous) {
+          (returnedFrom is DefinedElementType &&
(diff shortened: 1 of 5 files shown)


lib/src/model.dart
```diff
@@ -6956,7 +6873,7 @@ class PackageBuilder {
       //                   many AnalysisDrivers
       // TODO(jcollins-g): make use of DartProject isApi()
       // TODO(keertip): Use summary 2 after fixing #2017
-      AnalysisDriver.useSummary2 = false;
+      AnalysisDriver.useSummary2 = true;
```
scheglov: Remove this?
It is `true` by default, and in general you should not access private APIs :-)

keertip: done.

lib/src/element_type.dart
```diff
@@ -5,23 +5,30 @@
 /// The models used to represent Dart code.
 library dartdoc.element_type;
 
+import 'dart:collection';
```
devoncarew: organize imports?

lib/src/element_type.dart
```diff
@@ -83,31 +93,104 @@ abstract class ElementType extends Privacy {
...
+    }
     return name;
   }
 
   @override
-  String get nameWithGenerics => name;
+  String get name => type.name ?? '';
+}
+
+// A FunctionType that does not have an underpinning Element.
```
devoncarew: convert to a dartdoc comment?

lib/src/element_type.dart
keertip: done.

lib/src/element_type.dart
keertip: done.

keertip:
Fixes #2017

PR #2085, opened 2019-12-03
Move extendedDocLink creation to ModelElementRenderer

jdkoren:
Also separate it from oneLineDoc and update the templates accordingly.

### DIFF
--- a/lib/src/model/model_element.dart
+++ b/lib/src/model/model_element.dart
@@ -797,7 +797,7 @@ abstract class ModelElement extends Canonicalization
   /// does not exist.
   String get extendedDocLink {
     if (hasExtendedDocumentation) {
-      return '<a href="${href}">[...]</a>';
+      return ModelElementRendererHtml().renderExtendedDocLink(this);
     }
     return '';
   }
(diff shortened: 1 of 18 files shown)


lib/src/model/model_element.dart
```diff
@@ -977,19 +977,8 @@ abstract class ModelElement extends Canonicalization
...
-  String _oneLineDoc;
-
   @override
-  String get oneLineDoc {
-    if (_oneLineDoc == null) {
-      _oneLineDoc = computeOneLineDoc();
-    }
-    return _oneLineDoc;
-  }
+  String get oneLineDoc => _documentation.asOneLiner;
```
jcollins-g: glad the no-super workaround is gone

test/model_test.dart
```diff
@@ -2757,10 +2760,9 @@ String topLevelFunction(int param1, bool param2, Cool coolBeans,
 
     test('has extended documentation', () {
       expect(lengthX.hasExtendedDocumentation, isTrue);
-      expect(
-          lengthX.oneLineDoc,
-          equals(
-              'Returns a length. <a href="fake/WithGetterAndSetter/lengthX.html">[...]</a>'));
+      expect(lengthX.oneLineDoc, equals('Returns a length.'));
+      expect(lengthX.extendedDocLink,
+          equals(ModelElementRendererHtml().renderExtendedDocLink(lengthX)));
```
jcollins-g: I think we should have at least one literal matching string here for the Html renderer test, especially since it is very targeted -- otherwise we're simply validating that extendedDocLink returns whatever the renderer does.

jcollins-g:
@jdkoren Also, whenever I touch the templates I like to upload sample screenshots.  It helps make sure there's not some subtle rendering problem I introduced, as the unit tests don't always cover how the browser shows something.  Could you upload a sample screenshot or two of pages with the changed templates?

test/model_test.dart
jdkoren: Done

jdkoren:
Screenshots before and after of a library:
master_ex_ex-library html
test_ex_ex-library html

Screenshots before and after of a class:
master_fake_AClassWithFancyProperties-class html
test_fake_AClassWithFancyProperties-class html

I had to add a minor fix to prevent enum ordinal members from showing an extended doc link.
Screenshots before and after of an enum:
master_fake_Color-class html
test_fake_Color-class html

PR #2096, opened 2019-12-16
Decouple href and from a model's filename

jdkoren:
Currently a model's href is also used to determine the filename where Dartdoc generates documentation. This change separates the two usages so that hrefs could be relative and include parts that should not be in a file path, such a `..`.

See #2090

### DIFF
--- a/lib/src/model/class.dart
+++ b/lib/src/model/class.dart
@@ -252,7 +252,7 @@ class Class extends Container
     }
     assert(canonicalLibrary != null);
     assert(canonicalLibrary == library);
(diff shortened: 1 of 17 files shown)


lib/src/model/class.dart
```diff
@@ -201,6 +201,16 @@ class Class extends Container
   @override
   String get fileName => "${name}-class.html";
 
+  @override
+  String get filePath {
+    if (!identical(canonicalModelElement, this)) {
```
jdkoren: I basically copied `href`'s implementation in all these files and removed the `${package.baseHref}`, but I'm not sure this is the right approach.

PR #2125, opened 2020-01-10
Fix the search widget on non-root pages

### LINKED ISSUE #2124: Clicking search result while on non-root page goes to page not found
On `v0.30.0+1`. If you are on a non-root page and use the search bar, clicking a search result goes to a nonexistent page. Steps to reproduce:

1. `grind serve-test-package-docs`
1. Open a browser window and navigate to the served pages on localhost.
1. Enter some text in the search bar and click on a result.
    * On the root page, this should work just fine.
    * On any other page, it will most likely go to a page that doesn't exist. [BUG]

Similar problem as #2117, but it's the search result items rather than the script itself. The script constructs a list where each item uses an incomplete href, expecting base href to take care of the rest. [Source](https://github.com/dart-lang/dartdoc/blob/master/lib/resources/script.js#L204)

jdkoren:
Fixes #2124 
Also simplifies the fix for #2117

### DIFF
--- a/lib/resources/script.js
+++ b/lib/resources/script.js
@@ -100,6 +100,8 @@ function initSearch(name) {
     'constructor' : 4
   };
 
+  var baseHref = $('body').data('base-href');
+
   function findMatches(q) {
     var allMatches = []; // list of matches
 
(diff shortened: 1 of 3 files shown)


jdkoren:
I wonder if we need the same change [here](https://github.com/dart-lang/dartdoc/blob/master/lib/resources/script.js#L177)

PR #2137, opened 2020-01-21
Fix a crash when using --no-generate-docs

jdkoren:
Fixes a crash when accessing `documentationAsHtml` directly. Most usages first check `hasDocumentation`, but nothing enforces this. Notably, `EmptyGenerator` accesses this property without checking first (you can reproduce the crash by passing `--no-generate-docs`).


jdkoren:
> I think a better solution would be to move this to `_parseDocumentation` and return ([], false) as the value of the tuple. The renderers should be able to handle that as input, and there's less special casing that way.

This works, too. In that case I added a check in the renderer so it skips creating further objects.

PR #2145, opened 2020-02-12
Create some initial templates for md

jdkoren:
These are ~~very likely~~ most definitely not perfect, but gets us moving on wiring together a backend for them. They are currently not loaded yet as `--format md` still uses the EmptyGenerator.


jdkoren:
> I don't see a problem with this as written, but it's going to be challenging to know what's potentially wrong until more gets wired up and we can look at some sample output as you suggest in the PR. Approved.

It's mainly just to have a place for the fields to go. I expect the first round of actual output will look really bad :upside_down_face:

PR #2152, opened 2020-02-26
Add markdown renderers and renderer factory

jdkoren:


lib/src/render/renderer_factory.dart
```diff
@@ -86,3 +97,50 @@ class HtmlRenderFactory extends RendererFactory {
...
+
+class MdRenderFactory extends RendererFactory {
+  @override
+  TemplateRenderer get templateRenderer => MdTemplateRenderer();
+
+  @override
+  CategoryRenderer get categoryRenderer => CategoryRendererMd();
+
+  // We render documentation as HTML for now.
+  // TODO(jdkoren): explore using documentation directly in the output file.
```
jcollins-g: This will be a challenge, particularly when it comes to tools.  I'm also curious as to whether tool execution works correctly end-to-end here.

jdkoren: Indeed, `@tool` was one of the reasons I thought to leave documentation as html inside the markdown page. Same for `@video`.

We _could_ in theory run the tools as they are and expect that whoever is running Dartdoc has taken the output format into account when they added the tools...

jcollins-g: Well... that's not completely unreasonable _if_ we provide a way to tell the tool what sort of output to generate.  Keeping things HTML for now though makes a lot of sense as if you want any sort of compatibility with what Flutter is doing we'll either have to extend those tools or make it safe for them to continue outputting HTML.

jdkoren:
Index
_tmp_dartdoc_test_index md

Apple class
_tmp_dartdoc_test_ex_Apple-class md

isGreaterThan method
_tmp_dartdoc_test_ex_Apple_isGreaterThan md

PR #2162, opened 2020-03-13
Markdown template tweaks

jdkoren:
Lots of little tweaks that increase readability of the generated markdown. Some notable issues remain:
- Angle brackets are tricky. We need to escape some so that they aren't interpreted as HTML tags (e.g. `<B>`), however we also end up escaping brackets in source code snippets.
- Special directives like `@tool` seem to not work

### DIFF
--- a/lib/templates/md/top_level_property.md
+++ b/lib/templates/md/top_level_property.md
@@ -7,7 +7,7 @@
 {{>categorization}}
 
 {{#hasNoGetterSetter}}
-{{{ linkedReturnType }}} {{>name_summary}}
+{{{ linkedReturnType }}} {{>name_summary}}  {{!two spaces intentional}}
 {{>features}}
 
 {{>documentation}}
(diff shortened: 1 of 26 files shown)


jdkoren:
Index.md
_tmp_dartdoc_test_md_index md

ex_library.md
_tmp_dartdoc_test_md_ex_ex-library md

Klass-class.md
_tmp_dartdoc_test_md_ex_Klass-class md

invokeTool.md
_tmp_dartdoc_test_md_ex_ToolUser_invokeTool md

lib/templates/md/top_level_property.md
```diff
@@ -7,10 +7,11 @@
 {{>categorization}}
 
 {{#hasNoGetterSetter}}
-{{{ linkedReturnType }}} {{>name_summary}}
+{{{ linkedReturnType }}} {{>name_summary}}  {{!two spaces intentional}}
```
jcollins-g: What is the purpose of "two spaces intentional"?

jdkoren: Two spaces at the end of a line renders a line break, but not a paragraph break.

Normally a line break just renders a space (unless there's other formatting involved, like headings). Two line breaks renders a paragraph break, separating the blocks with extra vertical space that I didn't want.

PR #2194, opened 2020-04-25
Fix reported avoid_types_as_parameter_names lints.

scheglov:


scheglov:
Ah, no, it was just a random PR, without an intention to work on dartdoc seriously :-)

PR #2238, opened 2020-06-16
Allow DartDoc to execute a callback after generateDocs

mehmetf:


mehmetf:
@jcollins-g I assume you are OK with this since you already reviewed the change in g3.

PR #2246, opened 2020-06-24
Use &lt; instead of < in markdown

rokob:
There are several places in the generated markdown where generics get rendered with <types>. This changes those to instead use &lt;types>. Some markdown renderers don't like <types> because it looks like embedded html.


mehmetf:
Should the `>` be changed to `&gt;` for symmetry? :)

rokob:
I was coping the style of the rest of the codebase, for example https://github.com/dart-lang/dartdoc/pull/2162 only does the open bracket. You only really need to do the open bracket as the > doesn't raise any red flags to parsers. I'm happy to go either way.

PR #2248, opened 2020-07-01
Depend on analyzer 0.39.11, use overrideKnownFeatures() instead of bogus_ features.

scheglov:


test/experiment_options_test.dart
```diff
@@ -5,81 +5,123 @@
 /// Unit tests for lib/src/experiment_options.dart.
 library dartdoc.experiment_options_test;
 
+import 'dart:cli';
```
devoncarew, line 8: Do we need to depend on this library? If there's a straightforward way to write around the use of `waitFor` we probably should. We probably need to re-think dart:cli, and having fewer deps on it would be helpful.

scheglov: Yeah, I had to use it because `overrideKnownFeatures` works only for synchronous code, and options contributors of DartDoc are asynchronous. So, this is the only way to reconcile them.

Alternatively I could add `overrideKnownFeaturesAsync` or make `overrideKnownFeatures` async by default. This would require publishing a new version of analyzer. It is not hard though. Let me know if you think that this would be better. I also read that `dart:cli` is considered not nice, so I'm in doubt.

devoncarew: I'd add a commit to the use site of `waitFor` to briefly describe why we're using it. Perhaps a future resolution would be to change the dartdoc options contributors to be sync? I'm not familiar with why they're currently async.

PR #2253, opened 2020-07-07
Don't HTML escape source code blocks in Markdown

### LINKED ISSUE #2252: Code blocks in Markdown output don't need HTML escapes
Dartdoc generates fenced blocks for code, which should be rendered as-is. Using HTML escapes within these fenced blocks results in code that can be diffcult to read.

Screenshot from 2020-07-07 10-12-13

jdkoren:
Adds a renderer in between to control whether or not we apply HTML
escapes to source code blocks.

Fixes #2252


lib/src/render/source_code_renderer.dart
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
```
srawlins, line 1: This should be 2020.

lib/src/render/source_code_renderer.dart
```diff
@@ -0,0 +1,21 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:convert';
+
+abstract class SourceCodeRenderer {
```
srawlins, line 7: I think the implementations or the abstract class should have a doc comment.

lib/src/render/source_code_renderer.dart
jdkoren: Fixed

lib/src/render/source_code_renderer.dart
jdkoren: Thanks, added some in next commit

lib/src/render/source_code_renderer.dart
```diff
@@ -0,0 +1,23 @@
...
+  String renderSourceCode(String source) => source;
+}
+
+/// [SourceCodeRenderer] that escapes characters for HTML.
+class SourceCodeRendererHtml extends SourceCodeRenderer {
+  @override
+  String renderSourceCode(String source) {
+    return (const HtmlEscape()).convert(source);
+  }
+}
```
devoncarew, line 23: As a drive by comment, I'd `dartfmt` this file.

jdkoren: Indeed, there were a few files flagged by dartfmt.

PR #2255, opened 2020-07-08
Improve performance by 22% via memoization.

srawlins:
I tested with the camera flutter plugin. Before this change it took
67 seconds. Multiple fixes brought down the duration:

* memoize Package.fileType => 53 seconds (79% duration)
* memoize ModelElement.name => 52 seconds (78% duration)

### DIFF
--- a/lib/src/model/model_element.dart
+++ b/lib/src/model/model_element.dart
@@ -710,12 +710,12 @@ abstract class ModelElement extends Canonicalization
             var confidence = highestScore - secondHighestScore;
             var message =
                 '${candidateLibraries.map((l) => l.name)} -> ${candidateLibraries.last.name} (confidence ${confidence.toStringAsPrecision(4)})';
-            var debugLines = <String>[];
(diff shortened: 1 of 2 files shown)


lib/src/model/model_element.dart
```diff
@@ -710,12 +710,12 @@ abstract class ModelElement extends Canonicalization
...
                 '${candidateLibraries.map((l) => l.name)} -> ${candidateLibraries.last.name} (confidence ${confidence.toStringAsPrecision(4)})';
-            var debugLines = <String>[];
-            debugLines.addAll(scoredCandidates.map((s) => '${s.toString()}'));
 
             if (confidence < config.ambiguousReexportScorerMinConfidence) {
               warnable.warn(PackageWarning.ambiguousReexport,
-                  message: message, extendedDebug: debugLines);
+                  message: message,
+                  extendedDebug:
+                      scoredCandidates.map((s) => '${s.toString()}'));
```
devoncarew, line 718: `'${s.toString()}'` ==> `'$s'`

srawlins: done

PR #2259, opened 2020-07-09
Performance improvements and refactoring

srawlins:
When documenting the camera Flutter plugin, these changes reduce the time to document from 52 seconds to 51 seconds.

Much of this speedup likely came from memoizing Package.documentedWhere, which has a few expensive calls like `linkToRemote` and `linkToUrl`. Additionally ModelElement.locationPieces is now memoized.

The rest is a few categorical refactorings:

* ModelElement.canonicalLibrary:
    * Refactor most of the complexity of ModelElement.canonicalLibrary into a new private method.
    * Don't compute topLevelElement until you need it.
    * Short circuit if `definingLibrary.exportedInLibraries` is null.
    * Reduce nesting
    * Create warning message only if it will be used.
* Prefer for loops over Iterable.forEach.

### DIFF
--- a/lib/dartdoc.dart
+++ b/lib/dartdoc.dart
@@ -290,10 +290,10 @@ class Dartdoc {
     var indexJson = path.joinAll([normalOrigin, 'index.json']);
     var foundIndexJson = false;
     for (var f in Directory(normalOrigin).listSync(recursive: true)) {
-      var fullPath = path.normalize(f.path);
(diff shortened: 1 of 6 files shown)


srawlins:
Yeah I was trying a bunch of little things, cleaning code, and none of it made it _worse_, and I think the code is more idiomatic and maintainable so, might as well cut and commit.

PR #2285, opened 2020-08-06
add more types

devoncarew:
- add more types

Add some more explicit typing. This turns on the `type_annotate_public_apis` lint, and addresses a few items found with `implicit-casts: false` and `implicit-dynamic: false` (those last two weren't enabled as there are many more failures).


devoncarew:
I'm not sure why the tests / CIs aren't completing. I'll try restarting them once more.

PR #2286, opened 2020-08-07
Fix paragraph spacing in enum values.

### LINKED ISSUE #2268: Paragraph breaks don't render correctly for enum dartdocs
e.g. https://master-api.flutter.dev/flutter/rendering/CrossAxisAlignment-class.html

The source code is:

```dart
  /// Place the children along the cross axis such that their baselines match.
  ///
  /// Because baselines are always horizontal, this alignment is intended for
  /// horizontal main axes. If the main axis is vertical, then this value is
  /// treated like [start].
  ///
  /// For horizontal main axes, if the minimum height constraint passed to the
  /// flex layout exceeds the intrinsic height of the cross axis, children will
  /// be aligned as close to the top as they can be while honoring the baseline
  /// alignment. In other words, the extra space will be below all the children.
  ///
  /// Children who report no baseline will be top-aligned.
```

The resulting HTML looks like this:

Screen Shot 2020-07-15 at 1 27 10 PM

It should look like this:

Screen Shot 2020-07-15 at 1 27 10 PM

srawlins:
Since enum values listed on the enum class's page are, I think, the
only case of multiple paragraphs inside a dd, they deserve suome
special styling.

Fixes #2268

### DIFF
--- a/lib/resources/styles.css
+++ b/lib/resources/styles.css
@@ -341,6 +341,12 @@ dd p {
   margin-bottom: 0;
 }
 
+/* Enum values do not have their own pages; their full docs are presented on the
+ * enum class's page. */
+dt.constant + dd p {
+  margin-bottom: 1em;
+}
+
 /* indents wrapped lines */
 section.summary dt {
   margin-left: 24px;


srawlins:
I'm going to side-step the flutter timeout and merge. It's happening on all PRs.

PR #2289, opened 2020-08-08
Preserve newline following `{@endtemplate}`.

### LINKED ISSUE #1848: Line breaks after template are lost
The following:

```
///    using the [AutomaticKeepAliveClientMixin], then implementing the
///    [wantKeepAlive] getter and calling [updateKeepAlive].
/// {@endtemplate}
///
/// See also:
///
///  * [SliverChildBuilderDelegate], which is a delegate that uses a builder
///    callback to construct the children.
```

...renders like this:

> using the AutomaticKeepAliveClientMixin, then implementing the wantKeepAlive getter and calling updateKeepAlive. See also:
> * SliverChildBuilderDelegate, which is a delegate that uses a builder callback to construct the children.
> * SliverChildListDelegate, which is a delegate that has an explicit list of children.

Notice how the "See also" isn't a separate paragraph.

See e.g. https://master-docs-flutter-io.firebaseapp.com/flutter/widgets/SliverChildDelegate-class.html

srawlins:
Additionally introduce (some of?) the first unit tests. These live
in test/unit/ for now. This includes a new dev dependency on mockito.

Fixes #1848

### DIFF
--- a/test/unit/comment_processable_test.dart
+++ b/test/unit/comment_processable_test.dart
@@ -1,3 +1,7 @@
+// Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
 import 'dart:io' show Directory;
 
 import 'package:dartdoc/src/dartdoc_options.dart';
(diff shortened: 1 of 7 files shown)


test/unit/comment_processable_test.dart
```diff
@@ -0,0 +1,191 @@
+import 'dart:io' show Directory;
```
devoncarew, line 1: nit: copyright

srawlins:
Side-stepping the flutter bot timeout. This is affecting all PRs.

PR #2292, opened 2020-08-12
Privatize some of the interface of Class

srawlins:
* Deprecate the setters `Class.mixins` and `Class.supertype` and method `Class.isInheritedFrom`.
* Small clean-ups in Class.


srawlins:
Side-stepping timing-out flutter bot.

PR #2300, opened 2020-08-16
Only sky_engine can have _embedder.yaml file.

scheglov:
And analyzer [supports](https://github.com/dart-lang/sdk/blob/ebd5b622ba0228b7a90cff7177430ed161be7b51/pkg/analyzer/lib/src/context/builder.dart#L244) only `sky_engine`. So, there is no reason to test, or expect that it will work, with any other package.


test/dartdoc_test.dart
```diff
@@ -367,7 +366,7 @@ void main() {
       expect(
           dart_bear.allClasses.map((cls) => cls.name).contains('Bear'), isTrue);
       expect(p.packageMap['Dart'].publicLibraries, hasLength(3));
-    });
+    }, solo: true);
```
srawlins, line 369: Remove?

scheglov: Done.

PR #2301, opened 2020-08-16
Add testing/analysis_options.yaml

scheglov:
Packages in `testing/` should not be analyzed with the `analysis_options.yaml` from the root of the `dartdoc` package, for example because this file completely excludes `testing/**` from analysis.


devoncarew:
I believe you can ignore the failure in the flutter bot (though, we should investigate the failure; I'll open an issue for that).

devoncarew:
I filed https://github.com/dart-lang/dartdoc/issues/2302 for the CI issue.

PR #2287, opened 2020-08-07
remove dartdoc's dep on package:quiver

devoncarew:
- remove dartdoc's dep on package:quiver
- add some import prefixes to make more clear which packages we import but that we only use a method or two from
- expand the supported version range for `package:cli_util`

@srawlins - this PR is a bit speculative. In dartdoc we're importing package:quiver - a fairly large package - but only using a very small bit of its API. I'd definitely welcome your thoughts here -

PR #2305, opened 2020-08-17
remove dartdoc's dep on package:quiver

devoncarew:
Redux of https://github.com/dart-lang/dartdoc/pull/2287.

- remove dartdoc's dep on package:quiver
- add some import prefixes to make more clear which packages we import but that we only use a method or two from
- expand the supported version range for package:cli_util


devoncarew:
For some reasons my PRs were failing when I was using branches on the repo. I switched to creating branches on my fork - looks like the bots like that more.

PR #2314, opened 2020-08-19
Remove dependency on deprecated resource package.

srawlins:
Copy implementation for loading bytes from a 'package:' URI.


srawlins:
Yeah I can see us doing that longterm. There are a lot of templates so it would perhaps be quite a task to set this up with the build package...

PR #2315, opened 2020-08-20
Update all (approximately) i/o operations to use ResourceProvider

### LINKED ISSUE #2297: Use a file system abstraction
This is required for making the integration tests smaller, and will encourage better testing. Currently, the dartdoc repository exclusively uses real file system APIs like `Directory()` from dart:io. It should instead use an abstraction, like `package:file`, or ResourceProvider from analyzer. I vote ResourceProvider. Then PhysicalResourceProvider is used in real code, and MemoryResourceProvider is used in tests.

srawlins:
Fixes #2297 

All i/o operations now pass through a ResourceProvider, an abstraction which allows disk-free testing. In this change, almost all ResourceProviders used are the PhysicalResourceProvider; no tests are converted from being filesystem-dependent to using MemoryResourceProvider.

dartdoc uses the filesystem extensively, so this change is large. Often top-level or static elements become more complicated, depending on a ResourceProvider variable.

Some extra dart:io functionality not found in ResourceProvider is added via an extension, ResourceProviderExtensions.

Many classes have a new `ResourceProvider resourceProvider` field: DartdocFileWriter, SnapshotCache, DartToolDefinition, ToolConfiguration, DartdocOption, PackageMetaProvider, PackageMeta, ToolTempFileTracker,

Each of SnapshotCache and ToolTempFileTracker has a static `instance` field which took no arguments; now that they need a ResourceProvider, it was unwieldy to pass a ResourceProvider to get the instance each time, so a new method, `createInstance` creates the instance.


lib/dartdoc.dart
```diff
@@ -289,35 +296,45 @@ class Dartdoc {
...
-      } else {
-        // Error messages are orphaned by design and do not appear in the search
-        // index.
-        if (<String>['__404error.html', 'categories.json'].contains(fullPath)) {
-          _warn(packageGraph, PackageWarning.orphanedFile, fullPath,
-              normalOrigin);
+
+    void checkDirectory(Folder dir) {
+      for (var f in dir.getChildren()) {
+        //if (path.basename(f.path) == '.' || path.basename(f.path) == '..') {
```
devoncarew, line 302: Do you intend this to be commented out?

lib/src/io_utils.dart
```diff
@@ -19,15 +25,73 @@ String resolveTildePath(String originalPath) {
...
+    } else {
+      return getFolder('TODO(srawlins)');
+    }
+  }
+
+  String get resolvedExecutable {
+    if (this is PhysicalResourceProvider) {
+      return io.Platform.resolvedExecutable;
+    } else {
+      return 'TODO(srawlins)';
```
devoncarew, line 63: I assume this PR is still a WIP?

lib/dartdoc.dart
srawlins: Oops, fixed.

lib/src/io_utils.dart
srawlins: Sorry, that part is not implemented yet. It will be the part used in tests with MemoryResourceProvider. Moved TODO to comment; return null.

srawlins:
Same TargetKind crash on analyzer at HEAD. Side-stepping.

srawlins:
Side-stepping analyzer 0.40.1 breakage.

PR #2316, opened 2020-08-24
Improve error when FLUTTER_ROOT is missing.

### LINKED ISSUE #2279: Invalid argument(s): join(null, "bin", "cache", "dart-sdk"): part 0 was null, but part 1 was not.
while we generating doc this error will come.And I set dart-SDK path ,flutter-SDK path but is not working

srawlins:
This is a re-do, sort of, of #1714, which was broken in #2190. Fixes #2279

### DIFF
--- a/lib/src/dartdoc_options.dart
+++ b/lib/src/dartdoc_options.dart
@@ -1603,6 +1603,11 @@ Future<List<DartdocOption<Object>>> createDartdocOptions(
           (option.root['topLevelPackageMeta'].valueAt(dir) as PackageMeta)
               .requiresFlutter) {
         String flutterRoot = option.root['flutterRoot'].valueAt(dir);
+        if (flutterRoot == null) {
+          // For now, return null. An error is reported in
+          // [PackageBuilder.buildPackageGraph].
+          return null;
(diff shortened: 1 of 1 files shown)


devoncarew:
Interesting - when does dartdoc run `pub get` / ` flutter pub get`?

srawlins:
It looks like it only runs it when `.dart_tool/package_config.json` is not found... Not sure why we voluntarily do it, but it's been in there for at least two years.

devoncarew:
> It looks like it only runs it when `.dart_tool/package_config.json` is not found... Not sure why we voluntarily do it, but it's been in there for at least two years.

Interesting. We should likely change this to hard fail if some set of conditions aren't met, instead of falling back on running pub for the user.

srawlins:
Looks like analyzer is broken at HEAD (~0.40.1-dev)

scheglov:
@bwilkerson There might be an issue with `TargetKind` in the analyzer.

srawlins:
@scheglov Travis turned green.

srawlins:
@scheglov This is ready for review.

PR #2330, opened 2020-09-03
Rename some fields which erroneously reference "default" constructors.

srawlins:
A "default constructor" is an unnamed, zero-arg constructor. But the
constructors being referenced in these fields are just "unnamed."

* Deprecated `Class.defaultConstructor`: use `Class.unnamedConstructor`.
* Deprecated `Constructor.isDefaultConstructor`: use
  `Constructor.isUnnamedConstructor`.


stereotype441 (approved):
lgtm

PR #2336, opened 2020-09-04
enable the await_only_futures lint

devoncarew:
- enable the await_only_futures lint


devoncarew:
Hmm, this PR may be changing behavior somehow...

PR #2337, opened 2020-09-04
remove a use of resource loading

devoncarew:
- remove a use of resource loading

I believe this is only used when generating the docs for the Dart sdk. I'll prep a CL for the drt-lang/sdk repo; it would need to land in the same commit (or soon after) the one that rolls the next version of dartdoc into the sdk.


devoncarew:
This is the CL we'd need to land in the SDK repo: https://dart-review.googlesource.com/c/sdk/+/161880.

srawlins (approved):

devoncarew:
@srawlins - ping me when we next want to roll dartdoc into the sdk repo - we'll need to land my CL at the same time (or put the changes into the roll CL).

PR #2379, opened 2020-10-07
Remove packageGraph from TemplateData.

### LINKED ISSUE #2377: Provide more getter helpers for templates; use fewer extraneous objects
I think it is a smell in a template to reach out of a ModelElement into strange places like `.packageGraph`. The desire to access information on the `packageGraph` field is not bad, but it should be more accessible so that templates don't have to jump through a UML diagram to get at what they need.

In the standard templates, there are actually only 28 instances of using nested properties at all:

```
git grep '{{\s*[a-zA-Z_]\+\.' lib/templates/ |wc -l
```

Many of these are perfectly innocent. I don't see a reason to change `{{self.name}}` or `{{parent.kind}}`. As a matter of fact, when filtering out `{{parent` and `{{self`, the only other case is using `{{packageGraph`.

srawlins:
Fixes #2377 

**Breaking change** for any custom templates which reference `packageGraph`.

This opens up too many possibilities for templates, and requires knowledge of
the PackageGraph class's interface.

Instead, offer `localPackages` iterator and `defaultPackage` field on
TemplateData.

### DIFF
--- a/lib/src/generator/template_data.dart
+++ b/lib/src/generator/template_data.dart
@@ -11,10 +11,10 @@ abstract class TemplateOptions {
 }
 
 abstract class TemplateData<T extends Documentable> {
-  final PackageGraph packageGraph;
+  final PackageGraph _packageGraph;
   final TemplateOptions htmlOptions;
 
-  TemplateData(this.htmlOptions, this.packageGraph);
+  TemplateData(this.htmlOptions, this._packageGraph);
 
   String get title;
   String get layoutTitle;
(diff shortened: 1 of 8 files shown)


srawlins:
@jcollins-g what do you think of this change? Maybe #2377 is just an opinion of mine, and referencing `packageGraph` in templates is fine... Like narrowing the API of what you can access in templates is an unending can of worms, but I dunno. What do you think?

jcollins-g (commented):
I agree with the general idea of restricting the surface in this way, constraining to getters on Template classes where possible.  Does this work with the markdown templates included with Dartdoc?  They too reference packageGraph.

srawlins:
For some reason only the index.md template referenced packageGraph.

jcollins-g (approved):
OK.  As long as what's referenced here works, the breaking change notice we'll drop in the changelog seems good enough.

PR #2434, opened 2020-11-18
Mustachio: Add property maps for each renderer class

srawlins:
These property maps provide functions for each valid property of a given type.
Currently the maps only reveal how to get a value (for a variable node) and
resolve a bool (for a conditional section). Still needed are Iterable functions
and value functions, for repeated and value sections, respectively.


srawlins:
Not sure what to do about the out of date `templates.renderers.dart` file. `grind build` results in an empty `git diff`.

jcollins-g (approved):

srawlins:
Thanks!

PR #2459, opened 2020-12-12
Switch from Source Code Pro to Roboto Mono

parlough:
The previous monospace font is a little blurry and many characters feel slightly squished. To standardize with other Dart web tools and improve clarity I suggest switching to Roboto Mono which is licensed under Apache License, Version 2.0.

This also removes Source Sans Pro which was not used at all as far as I could tell.

The rendering result on Chrome Dev (Linux): 

**Previously(Source Code Pro):**
image

**Suggested(Roboto Mono):**
image

SO much better <3

### DIFF
--- a/lib/resources/styles.css
+++ b/lib/resources/styles.css
@@ -245,7 +245,7 @@ a:hover {
 }
 
 pre.prettyprint {
-  font-family: 'Source Code Pro', Menlo, monospace;
+  font-family: 'Roboto Mono', Menlo, monospace;
   color: black;
   border-radius: 0;
   font-size: 15px;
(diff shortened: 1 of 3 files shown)


parlough:
I believe this test is failing due to _fe_analyzer_shared migrating to null safety. I think the following CL should fix it, but I'm not certain:

jcollins-g (changes requested):
You're right this looks better.  Could you instead give Roboto Mono a try, as used in code examples on the [dart.dev](https://dart.dev) site, such as at https://dart.dev/samples?  In general when we make changes we try to align to the larger Dart web style where possible.

Also it looks like Travis won't stop trying to run sdk-analyzer against stable unless you resync your PRs, so if you can do that it'll clear the error.

parlough:
@jcollins-g This looks like an improvement as well, good to go on my end :)

jcollins-g (approved):
This looks good, code bits are much more readable when I tried this locally.

PR #2460, opened 2020-12-13
Remove unnecessary bootstrap files and classes

parlough:
Bootstrap hasn't been used/included since https://github.com/dart-lang/dartdoc/pull/1452 and the remaining bootstrap classes present on the HTML simply weren't doing anything.

**Potentially breaking change:** Removes outdated bootstrap css files downstream template users may be using.


parlough:
I believe this test failed due to _fe_analyzer_shared migrating to null safety. I think the following CL which landed last night should have fixed it.

jcollins-g (approved):
Assuming tests pass when you merge the updated master into this PR, clear to land.  Thanks for cleaning this up!

Oh, and if you can, this qualifies as a breaking change, so note that in the PR description/first comment.  Just because we aren't using bootstrap doesn't mean there might not be some downstream custom template user who is depending on it.

PR #2461, opened 2020-12-13
Fix minor focus spacing

parlough:
On Chrome, the spacing in the sidebars cut off pieces of the focus outlines:

**Previously:**
image
image

**After change:**
image
image


parlough:
I believe this test failed due to _fe_analyzer_shared migrating to null safety. I think the following CL which landed last night should have fixed it.

jcollins-g (approved):
Approved, assuming things turn green after updating PR from head.

PR #2474, opened 2020-12-31
Update Highlight.js to 10.5.0

parlough:
There's been a lot of fixes and improvements since the previous version.

Relating to Dart, the primary improvements are:

> - enh(dart) Fix empty block-comments from breaking highlighting (#2898) Jan Pilzer
> - enh(dart) Fix empty doc-comment eating next line Jan Pilzer
> - **enh(dart) Add late and required keywords, the Never built-in type, and nullable built-in types (#2550) Sam Rawlins**

You can refer to the entire changelog here: https://github.com/highlightjs/highlight.js/blob/master/CHANGES.md

As for the increased file length, that is how they distribute their minified bundle now. The languages included are the same as listed in the README.

**Corresponding pub-dev PR:** https://github.com/dart-lang/pub-dev/pull/4387


jcollins-g (approved):
Reviewed by checking a pack downloaded from highlightjs.org on 1/6/21 is byte for byte identical when the given languages are selected, and that highlighting still seems to work.

PR #2481, opened 2021-01-11
Move templates.dart to use ResourceProvider.

srawlins:
* The top-level functions in resource_loader.dart were wrapped in a class, ResourceLoader, to account for ResourceProvider.
* **Breaking Change**: Templates.fromDirectory is made private and now requires
  a ResourceLoader.
* Templates.createDefault is made `@visibleForTesting` and now requires a
  ResourceLoader.

This includes a few minor cleanups as well.

### DIFF
--- a/lib/dartdoc.dart
+++ b/lib/dartdoc.dart
@@ -50,7 +50,8 @@ class DartdocGeneratorOptionContext extends DartdocOptionContext
 
 class DartdocFileWriter implements FileWriter {
   final String outputDir;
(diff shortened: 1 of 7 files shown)


srawlins:
I think the new ResourceLoader class could actually just be a set of ~3 extension methods on ResourceProvider. I'm happy to explore that in this CL or in a follow-up.

jcollins-g (approved):

parlough (approved):

PR #2487, opened 2021-01-26
Remove raw path package usage from a handful of files; use ResourceProvider

srawlins:


lib/src/model/package_builder.dart
```diff
@@ -424,11 +423,11 @@ class PubPackageBuilder implements PackageBuilder {
     var resources = dir.getChildren();
 
     var pubspec = resources.firstWhere(
-        (e) => e is File && path.basename(e.path) == 'pubspec.yaml',
+        (e) => e is File && _pathContext.basename(e.path) == 'pubspec.yaml',
         orElse: () => null);
 
     var libDir = resources.firstWhere(
-        (e) => e is Folder && path.basename(e.path) == 'lib',
+        (e) => e is Folder && _pathContext.basename(e.path) == 'lib',
```
parlough, line 430: This is causing the build to fail due to the method `_packageDirList` being static and not having access to the `_pathContext` getter. What's strange though is the analysis server is not identifying this as being the case.

parlough (approved):

lib/src/model/package_builder.dart
srawlins: Thanks for filing https://github.com/dart-lang/sdk/issues/44780. Fixed.

srawlins:
The model_special_cases_test.dart passes for me locally.

jcollins-g (approved):

PR #2488, opened 2021-01-26
Change ResourceLoader from a thin wrapper around ResourceProvider …

srawlins:
into an extension.

I'm not super married to this change, but I do think it is an improvement. ResourceLoader had turned into this this wrapper around ResourceProvider, and I decided it was silly for classes to be carrying around either a reference to a ResourceProvider, or a ResourceLoader, or both. This change simplifies things.

### DIFF
--- a/test/resource_loader_test.dart
+++ b/test/resource_loader_test.dart
@@ -4,26 +4,28 @@
 
 library dartdoc.resource_loader_test;
 
+import 'package:analyzer/file_system/file_system.dart';
 import 'package:analyzer/file_system/physical_file_system.dart';
 import 'package:dartdoc/src/generator/resource_loader.dart';
 import 'package:test/test.dart';
 
 void main() {
   group('Resource Loader', () {
-    ResourceLoader loader;
+    ResourceProvider resourceProvider;
 
     setUp(() {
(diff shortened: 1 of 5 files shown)


test/resource_loader_test.dart
```diff
@@ -4,26 +4,28 @@
...
-      var contents = await loader
-          .loadAsString('package:dartdoc/templates/html/index.html');
+      var contents = await resourceProvider
+          .loadResourceAsString('package:dartdoc/templates/html/index.html');
       expect(contents, isNotNull);
     });
 
     test('throws if non-package', () async {
-      expect(loader.loadAsString('wefoij:something'), throwsArgumentError);
+      expect(resourceProvider.loadResourceAsString('wefoij:something'),
```
parlough, line 27: Ah yes wefoij, my best friend!

parlough (approved):

jcollins-g (approved):
I can see having mixed feelings on it, though this is a good candidate for converting to an extension as you basically are just adding methods to ResourceProvider.

test/resource_loader_test.dart
srawlins: Hehe same here.

PR #2507, opened 2021-02-07
Convert originalMember to an optional parameter of ModelElement

parlough:
We're now in a world with https://github.com/dart-lang/sdk/issues/15101 and https://github.com/dart-lang/sdk/issues/31543 fixed, so this change should be fine.


jcollins-g (approved):
Very nice!

PR #2512, opened 2021-02-08
Mustachio: reduce generated code with some base helpers

srawlins:
These two helpers, which apply to rendering a variable (`{{foo.bar}}`) really reduce the amount of nearly identical code which was being repeatedly stamped out in the generated output.

Additionally, it fixes a bug in the error message being thrown when a variable cannot be resolved.

### DIFF
--- a/lib/src/mustachio/renderer_base.dart
+++ b/lib/src/mustachio/renderer_base.dart
@@ -4,6 +4,7 @@
 
 import 'dart:collection';
 
+import 'package:analyzer/dart/element/type.dart';
 import 'package:analyzer/file_system/file_system.dart';
 import 'package:meta/meta.dart';
 import 'parser.dart';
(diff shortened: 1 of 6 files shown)


lib/src/mustachio/renderer_base.dart
```diff
@@ -4,6 +4,7 @@
 
 import 'dart:collection';
 
+import 'package:analyzer/dart/element/type.dart';
```
parlough, line 7: I don't think you use this anywhere.

parlough (approved):
Looks good to me, a nice simplification.

lib/src/mustachio/renderer_base.dart
srawlins: Correct! Thanks

jcollins-g (approved):
Very nice, it makes the generated code much more readable.

PR #2525, opened 2021-02-18
Introduce TemplateDataWithLibrary and TemplateDataWithContainer.

srawlins:
These two interfaces are necessary in order to render certain TemplateData
subtypes into _sidebar_for_library.html and _sidebar_for_container.html,
respectively.

Since _sidebar_for_library.html references `{{#library}}`, the static type
rendered into this template needs a `library` getter. Same for the other
sidebar.

Additionally, _sidebar_for_container.html accesses `hasPublicConstructors` and
`publicConstructorsSorted` on Container, so that type needs those getters.

### DIFF
--- a/lib/src/generator/template_data.dart
+++ b/lib/src/generator/template_data.dart
@@ -4,8 +4,9 @@
 
 import 'package:dartdoc/src/model/model.dart';
 
-typedef ContainerSidebar = String Function(Container, TemplateData);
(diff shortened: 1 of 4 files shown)


jcollins-g (approved):
Ah yes.  Types are a thing, and mirrors are much more forgiving here.

srawlins:
Hahaha yes. Alas.

PR #2533, opened 2021-02-24
Remove unnecessary use of list view

parlough:
We don't need to use a list view in these cases as we are creating the list as needed so the caller won't share it. I've also made them not `growable` as they don't need to be modified afterwards.

### DIFF
--- a/lib/src/element_type.dart
+++ b/lib/src/element_type.dart
@@ -5,8 +5,6 @@
 /// The models used to represent Dart code.
 library dartdoc.element_type;
 
-import 'dart:collection';
-
 import 'package:analyzer/dart/element/element.dart';
 import 'package:analyzer/dart/element/nullability_suffix.dart';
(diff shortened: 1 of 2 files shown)


lib/src/element_type.dart
```diff
@@ -293,11 +286,13 @@ abstract class DefinedElementType extends ElementType {
   String createLinkedReturnTypeName() => returnType.linkedName;
 
   Iterable<ElementType> _typeArguments;
+
+  /// An unmodifiable list of this element type's parameters.
   Iterable<ElementType> get typeArguments {
     _typeArguments ??= (type as ParameterizedType)
```
kevmoo, line 292: This could be ` => _typeArguments ??=...`

parlough: Good point, I've switched to an expression body. Thanks.

jcollins-g (approved):

PR #2537, opened 2021-02-24
Minor category renderer cleanup

parlough:
Cleans up the category renderer a bit:

- Switch rendering HTML to use StringBuffer throughout as we already used it for part. Some minor perf benefits here since this logic is hit more often.


parlough:
The mac test has been intermittently failing, if you resubmit the checks it _should_ be fine.

jcollins-g (approved):

PR #2538, opened 2021-02-24
Minor typedef renderer cleanup

parlough:
Cleans up the typedef renderer a bit:

- Cache type parameters as a local variable to avoid running checks from its getter twice
- Switch rendering to use `StringBuffer`. Some minor performance benefits here since this logic is hit more often.


jcollins-g (approved):
No objection here; the StringBuffer makes little difference and may even improve the legibility.  Thanks!

PR #2539, opened 2021-02-24
Cleanup language feature usage and renderer

parlough:
Cleans up the `LanguageFeature` class and renderer a bit:

- Switch HTML rendering to `StringBuffer` for more consistency and slightly improved performance since this logic is hit more often.
- Pulls out usage of `feature.featureUrl` to a local variable to enable non-null type inference in the future


jcollins-g (approved):
A little harder to read, but as it's in a commonly used renderer, probably worth it.  Approved!

PR #2546, opened 2021-02-25
Mustachio: avoid calling Iterable getter multiple times

srawlins:
Take this template, where `widgets` is an Iterable property:

    Text {{#widgets}} {{name}} {{/widgets}}.

A mustache renderer should assume that the `widgets` getter may have side
effects and may be expensive to compute, and it should only be called one
time. Currently, Mustachio will call this getter 2 or 3 times, as the behavior
differs if the section is inverted, and if the Iterable is empty.

This change is actually quite readable, and simplifies the code. The magic
sauce is in the lazy Iterable returned by `renderIterable`. If Mustachio does
not wish to render the elements, then the lazy map will not be evaluated.

### DIFF
--- a/lib/src/mustachio/renderer_base.dart
+++ b/lib/src/mustachio/renderer_base.dart
@@ -236,10 +236,7 @@ abstract class RendererBase<T> {
         // An inverted section is rendered with the current context.
         renderBlock(node.children);
       } else if (!node.invert && renderedIterable.isNotEmpty) {
-        var buffer = StringBuffer();
-        for (var renderedElement in renderedIterable) {
-          buffer.write(renderedElement);
-        }
+        var buffer = StringBuffer()..writeAll(renderedIterable);
         write(buffer.toString());
       }
       // Otherwise, render nothing.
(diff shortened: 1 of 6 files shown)


lib/src/mustachio/renderer_base.dart
```diff
@@ -230,13 +230,20 @@ abstract class RendererBase<T> {
...
+      var renderedIterable =
+          property.renderIterable(context, this, node.children);
+      if (node.invert && renderedIterable.isEmpty) {
+        // An inverted section is rendered with the current context.
         renderBlock(node.children);
+      } else if (!node.invert && renderedIterable.isNotEmpty) {
+        var buffer = StringBuffer();
+        for (var renderedElement in renderedIterable) {
+          buffer.write(renderedElement);
+        }
```
parlough, line 242: Can this be simplified to the following?

```dart
buffer.writeAll(renderedIterable);
```

srawlins: I love it. Done.

parlough (approved):

parlough:
@srawlins Can you merge master into your PRs to get the tests passing?

jcollins-g (approved):
LGTM

PR #2547, opened 2021-02-26
Mustachio: Trim whitespace preceding a tag

srawlins:
This brings Mustachio much closer to the output that the mustache package writes.


parlough (approved):

srawlins:
Test failure looks transient.

jcollins-g (approved):

PR #2554, opened 2021-03-02
Refactor header and footer support to use mustache templating.

srawlins:
* Remove `mixin GeneratorContext` and move contained code into
  DartdocGeneratorOptionContext, the only class which mixed in this mixin.
* **Breaking change**: Move DartdocGeneratorOptionContext from
  `lib/dartdoc.dart` to `lib/options.dart`, which is more directly related to
  the class, and is the only library in which this class is constructed or
  extended.
* Add `header`, `footer`, and `footerText` getters to
  DartdocGeneratorOptionContext.
* Add `customHeaderContent`, `customFooterContent`, and `customInnerFooterText`
  to DartdocGneratorBackendOptions.
* **Breaking change**: Privatize the DartdocGeneratorBackendOptions into a new
  private constructor which takes no parameters.
* Change all header/footer/footer-text placeholder comments in templates to
  instead be mustache interpolations.

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -6,7 +6,6 @@ analyzer:
   errors:
     unused_import: warning
     unused_shown_name: warning
-    todo: ignore
   exclude:
     - 'doc/**'
     - 'lib/src/third_party/pkg/**'
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -6,6 +6,7 @@ analyzer:
   errors:
     unused_import: warning
     unused_shown_name: warning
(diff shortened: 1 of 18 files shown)


parlough (approved):

jcollins-g (approved):
This looks good, and is way simpler than what I thought it would look like!

Approved, assuming you have tested this with Flutter and the custom headers/footers look reasonable there.

srawlins:
Header and footers look good to me. The use `--header` and `--footer-text`, but not `--footer`.

PR #2560, opened 2021-03-07
Update highlight.js and add support for plain text and C

parlough:
This update doesn't have too many improvements for us, but we are lacking support for plain text and C so to add support for that using their optimized builds, needed to update at the same time.

You can verify the minified `highlight.js` pack if desired by going to https://highlightjs.org/download/ and checking all supported languages(found in `readme.md`).

You can find this highlight.js update's changelog here: https://github.com/highlightjs/highlight.js/releases/tag/10.6.0

The main desire is to get support for `plaintext` and `c` but there's also small improvements in the `swift` and `xml` grammars as well as the more reliable `highlightAll` API.

**Some dart examples using `plaintext` instead:**
image


jcollins-g:
Add a screenshot or two of docs with the the new highlight.js active?

parlough:
@jcollins-g I've added a screenshot of the main goal, adding support for plaintext. Dart support should be the same as always.

parlough:
I'm converting this to a draft because as I was going through SDK docs I realized we should likely include support for C as well due to ffi becoming stable.

parlough:
@jcollins-g This is ready for review. I was preparing a CL tagging all code blocks in the SDK and I found not only `plaintext` needed in quite a few locations, but `c` as well now that `ffi` is stable and there are a few C code blocks.

jcollins-g (changes requested):
Hmm.  I'm not able to verify the highlight pack.  The diff is challenging to read, but I think the one you uploaded includes Kotlin which is not in the list of languages from the readme.

parlough:
@jcollins-g Strangely for me, re-downloading all of the selected languages does not result in a change of the file. The following are the languages I have selected according to the README(including the additions of **C** and **Plaintext**). Did you account for adding support for C? If it still doesn't match after that, feel free to download your own and commit to this branch, you should have access: `parlough:misc/update-highlightjs`

image

jcollins-g (approved):
I was on the wrong branch in my clone of your fork for this PR.

This update process needs to be added to the grinder so the verification is not manual every time; too many opportunities to make an error.

parlough:
@jcollins-g Next time I work with highlighting I'll get something figured out for automatic updating and/or verification. Thanks for sticking with me!

PR #2566, opened 2021-03-10
Do not perform hint or lint analysis while resolving files

srawlins:
Dartdoc previously skipped hints and lints, but this behavior was lost in the transition to ContextCollections.

This saves 10s of 210s in resolving Flutter's files while compiling Flutter's docs.


srawlins:
Oops. CI failing left and right. Maybe in an unreleased version of analyzer...

jcollins-g (approved):

PR #2572, opened 2021-03-12
Add avoid_dynamic_calls lint rule

srawlins:
It appears dartdoc is already clean with this lint 🥳 but I thought it would be nice to enforce it as well.


parlough (approved):

srawlins:
The singular head-scratcher is sort of curious to me...

jcollins-g (approved):

PR #2579, opened 2021-03-16
Update dev_dependencies to latest

kevmoo:


jcollins-g (changes requested):
Test failures here appear to be real.

PR #2623, opened 2021-04-23
Introduce a README for mustachio.

srawlins:
I had to choose between `lib/src/mustachio/` and `tool/mustachio` and decided to put the README in the latter, and put references to it in the former, since the latter directory has the actual generators.

Just a start, but I have most of the rest of the content sitting in a Google doc, so you might have faith that I will finish writing it 😁 .


jcollins-g (approved):
Looks good!  Glad to have the documentation here.

parlough (approved):
This seems like a great introduction from both a dartdoc user and dartdoc developer perspective, amazing job. 

I would personallysay landing this as is would be worth it and other improvements and missing pieces could follow :)

PR #2628, opened 2021-04-28
Breaking: Make containerDesc private and final in TemplateData implementations

srawlins:
PropertyTemplateData and MethodTemplateData each had a public non-final `containerDesc`. Each one could be made final by assigning in constructor initializers, and each could be made private; neither was ever accessed outside the class.


jcollins-g (approved):
The PR title might be better rephrased as "Make containerDesc private and final in TemplateData implementations".  Otherwise, LGTM!

PR #2664, opened 2021-06-02
Add end-to-end tests for Mustachio AOT compiler

srawlins:
These tests, in aot_compiler_render_test.dart, are a little irregular. The goal
is to write short test cases in which we specify the text of a Mustache
template, and the expected rendered text, visually near each other, which is a
challenge, as there are several steps to get from Mustache template text to
rendered output, including running the builders, and executing freshly
generated Dart code.

The solution here is a test system which takes a Mustache template text, runs
the mustachio AOT compiler to generate a Dart script which renders a context
object into that template, then executes the  generated Dart script, and
asserts on the output.

Writing these tests revealed a few bugs that are fixed concurrently:

* imports in the generated script should be accurate; this requires using
  code_builder, which requires a sizeable change to the AOT compiler code.
  code_builder allows you to "reference" symbols like types or top-level
  functions and the URL where they may be found.
* handle partial template paths with dots and slashes.

Additionally, fix some nits in runtime_renderer_builder_test.dart and
runtime_renderer_render_test.dart.

### DIFF
--- a/test/mustachio/render_tests_test.dart
+++ b/test/mustachio/render_tests_test.dart
@@ -15,8 +15,8 @@ void main() {
     var testCasePattern = RegExp(r'test\(.*,');
     var dartdocLibUri = await Isolate.resolvePackageUri(
         Uri.parse('package:dartdoc/dartdoc.dart'));
-    var dartdocPath =
-        p.dirname(p.dirname(p.joinAll(dartdocLibUri.pathSegments)));
+    var dartdocPath = p.dirname(
+        p.dirname(p.separator + p.joinAll(dartdocLibUri.pathSegments)));
     var runtimeRendererRenderTest = File(p.join(dartdocPath, 'test',
             'mustachio', 'runtime_renderer_render_test.dart'))
         .readAsLinesSync();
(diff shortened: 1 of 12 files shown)


jcollins-g (approved):
I like the direction this has taken!   Something seems to be wrong with windows paths, though.  Assuming that gets fixed, approved.

kevmoo:
https://github.com/dart-lang/dartdoc/pull/2664/checks?check_run_id=2737504814#step:7:3992

Looks like we need to use pkg:path!

test/mustachio/render_tests_test.dart
```diff
@@ -18,6 +16,10 @@ void main() {
     var dartdocLibUri = await Isolate.resolvePackageUri(
         Uri.parse('package:dartdoc/dartdoc.dart'));
     var dartdocPath = p.dirname(p.dirname(dartdocLibUri.path));
+    // Correct Windows issue path coming out of [Isolate.resolvePackageUri].
+    if (dartdocPath.startsWith(p.windows.separator)) {
+      dartdocPath = dartdocPath.substring(1).replaceAll('/', p.separator);
```
kevmoo, line 21: so weird that `pkg:path` doesn't "do the right thing" here!

srawlins: Yeah it is curious. I think the issue is that Isolate.resolvePackageUri gives a Uri like `file:///C:/one/two`, which maybe isn't incorrect because URIs should always have `/`, even on Windows? And then I ask for the `.path` out of that, which gives `/C:/one/two` and now we're in trouble.

srawlins:
I'm landing this before it breaks again.

PR #2669, opened 2021-06-07
Refactor lib/dartdoc.dart and dartdoc_test.dart

srawlins:
I was poking around these files and noticed a few things I could tidy up:

* `DartdocFileWriter.outputDir` can be private.
* `Dartdoc.writtenFiles`, `Dartdoc.outputDir`, `Dartdoc.validateLinks` can be private.
* `generateDocsBase` is only used in tests; mark is used as `@visibleForTesting` and move its dartdoc.
* There were two ResourceProviders in the test, among some other unused elements.


jcollins-g (approved):
Yep, and as far as I can tell these aren't used by the pub site, great!

PR #2676, opened 2021-06-10
Escape variable renders in runtime renderer

srawlins:
A bit shocking that this bug slipped through, but this corrects escaping behavior for the runtime renderers.


jcollins-g (approved):
I think it makes some sense -- the default templates don't usually _depend_ on us escaping.  It is more likely we depend on it not escaping, which this bug wouldn't impact.

PR #2677, opened 2021-06-10
Use Mustachio AOT compiler for default templates

srawlins:
This is a pretty big change; here goes:

* Contains the content of https://github.com/dart-lang/dartdoc/pull/2674 and https://github.com/dart-lang/dartdoc/pull/2676
* SidebarGenerator is much simplified; only takes a render function, not a Template object, which is only used in runtime renderers.
* Add blank lines in generated AOT renderer files which correct the whitespace of the output. 800 newlines in one file!
* `templates.dart` is viciously refactored; the diff is not really meaningful, maybe just new content is meaningful:
  * explicit list of partials is no longer needed; partials are eagerly parsed in both the Runtime renderers and the AOT renderers.
  * TemplatesLoader classes also not needed.
  * Templates class wildly refactored; instead of containing Template objects (only needed by the Runtime renderers; in AOT, the AOT render Dart code is in effect the parsed Template itself), it contains a dozen or so render functions.
  * Templates.fromContext now picks an subtype of Templates based on format ('html', 'md') and on presence of a templatesDir option.
* `Dartdoc.generator` is made not final purely so that we can test two generators on one Dartdoc; this saves ~20 seconds in testing.
* New renderers_output_test.dart generates HTML docs for a package using the AOT renderers (default) and the runtime renderers (forced), then walks the output trees of each result, comparing them, and making sure they generate the exact same bytes.


jcollins-g (approved):
This is great!  Also appreciate the precision in the comparison test, makes me much more comfortable switching this on!

PR #2686, opened 2021-06-17
Move ToolDefinition and related classes out of dartdoc_options

### LINKED ISSUE #2344: If first invocation of a Dart script, as a @tool, crashes, the snapshot is not made, but dartdoc does not stop.
This was discovered testing 0.33.0 on flutter.

srawlins:
ToolDefinition, DartToolDefinition, Snapshot, and SnapshotCache (~250 lines) are all only lightly related to dartdoc options. These four classes are easily extracted to their own file.

This change also replaces the Tuple2 usage with ToolStateForArgs, and refactors `modifyArgsToCreateSnapshotIfNeeded` to a new method which exists on both ToolDefinition and on DartToolDefinition.

These changes will help a new change to fix #2344 by emitting a warning from the new file.


jcollins-g (approved):
Yep, straightforward enough.

PR #2691, opened 2021-06-19
Refactor MultiFutureTracker to not be generic

srawlins:
The generic is never used; the Futures are always `Future<void>`.


jcollins-g (approved):
Don't even remember why it wasn't like this to begin with, thanks!

srawlins:
Dropping in favor of #2730.

PR #2806, opened 2021-09-22
Reduce broad imports of dartdoc.dart

srawlins:
This also involves moving `DartdocFailure` to its own file.

The idea here is that `dartdoc.dart` exports everything, so we really want to minimize imports of that library _within the package._ This should help with the null safety migration, because it simplifies the import graph, and in particular reduces cycles.


jcollins-g (approved):
Yes, yes, a hundred times yes

PR #2810, opened 2021-09-27
Migrate most mustachio tests to null safety

srawlins:
Migrate all of the tests that don't depend on generated code or html_generator.dart (neither are migrated yet).


jcollins-g (approved):
looks good!

PR #2855, opened 2021-11-16
Use "font-display: swap" for icon fonts, similar to monospace the line above.

isoos:


isoos:
The PR is more of a request for comments and a suggestion to also include the settings here, unless there is a reason that it doesn't work the same way as for monospace fonts.
/cc @jcollins-g

isoos:
Based on the additional info on https://www.zachleat.com/web/font-display-icon-fonts/ maybe it is better to not do that swap setting.

PR #2857, opened 2021-11-29
Add a resources-dir option

srawlins:
This adds support for an option to be provided at runtime which specifies where dartdoc's resources are located.

Currently resources are always loaded from `package:dartdoc/resources`, but this will not work for dartdoc compiled AOT.

CC @mit-mit 

Work towards https://github.com/dart-lang/sdk/issues/44610

### DIFF
--- a/lib/options.dart
+++ b/lib/options.dart
@@ -51,7 +51,7 @@ class DartdocGeneratorOptionContext extends DartdocOptionContext {
   String get relCanonicalPrefix =>
       optionSet['relCanonicalPrefix'].valueAt(context);
 
-  /// The 'templatesDir' Dartdoc option if one was specified; otherwise `null`.
+  /// The 'templatesDir' dartdoc option if one was specified; otherwise `null`.
   String get templatesDir => optionSet['templatesDir'].valueAt(context);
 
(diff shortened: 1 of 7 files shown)


srawlins:
@mit-mit let me know if you'd like another reviewer.

I've tested this in the SDK with a modified form of your https://dart-review.googlesource.com/c/sdk/+/220744 (just changing the flag name)

lib/options.dart
```diff
@@ -57,6 +57,9 @@ class DartdocGeneratorOptionContext extends DartdocOptionContext {
   // TODO(jdkoren): duplicated temporarily so that GeneratorContext is enough for configuration.
   @override
   bool get useBaseHref => optionSet['useBaseHref'].valueAt(context);
+
+  /// The 'resourcesDir' Dartdoc option if one was specified; otherwise `null`.
```
mit-mit, line 61: nit: don't we usually write dartdoc without a leading cap?

mit-mit (approved):
Looks great, thanks!

lib/options.dart
srawlins: Done

PR #2867, opened 2021-12-30
Update nnbd branch for latest changes and new resources-dir option

parlough:
Makes necessary changes to migrate the additions from the new resources-dir option.


srawlins (approved):
Thaaaaank you much!

PR #2868, opened 2021-12-30
Utilize new Object.hash function instead of quiver code

parlough:


pubspec.yaml
```diff
@@ -4,7 +4,7 @@ version: 4.1.0
 description: A non-interactive HTML documentation generator for Dart source code.
 homepage: https://github.com/dart-lang/dartdoc
 environment:
-  sdk: '>=2.12.0 <3.0.0'
+  sdk: '>=2.15.0 <3.0.0'
```
srawlins, line 7: Looks like these functions are available in 2.14.0. Let's go with that.

srawlins (approved):
I love this. Thanks!

PR #2870, opened 2021-12-31
Migrate tool/grind.dart to null safety

srawlins:
Net fewer lines 😁 .

Mostly late initializing top-level variables, a few other standard migration changes.

### DIFF
--- a/tool/grind.dart
+++ b/tool/grind.dart
@@ -105,16 +105,16 @@ Future<FlutterRepo> get cleanFlutterRepo async {
   return repoCompleter.future;
 }
 
-late final String _dartdocDocsPath = createTempSync('dartdoc').path;
+final String _dartdocDocsPath = createTempSync('dartdoc').path;
 
-late final Directory _sdkDocsDir = createTempSync('sdkdocs');
+final Directory _sdkDocsDir = createTempSync('sdkdocs').absolute;
 
 Directory cleanFlutterDir = Directory(path.join(
     path.context.resolveTildePath('~/.dartdoc_grinder'), 'cleanFlutter'));
 
-late final Directory _flutterDir = createTempSync('flutter');
+final Directory _flutterDir = createTempSync('flutter');
 
-late final Directory _languageTestPackageDir =
+final Directory _languageTestPackageDir =
     createTempSync('languageTestPackageDir');
 
 Directory get testPackage =>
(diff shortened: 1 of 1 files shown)


tool/grind.dart
```diff
@@ -1023,11 +1000,11 @@ Future<String> _buildPubPackageDocs(
     'Build an arbitrary pub package based on PACKAGE_NAME and PACKAGE_VERSION environment variables')
 Future<String> buildPubPackage() async {
   assert(Platform.environment.containsKey('PACKAGE_NAME'));
```
scheglov, line 1002: We probably don't need this, it will fail anyway.

tool/grind.dart
```diff
@@ -138,49 +126,33 @@ Directory get testPackageExperiments =>
...
 
-Directory get pluginPackageDocsDir =>
-    _pluginPackageDocsDir ??= createTempSync('test_package_flutter_plugin');
+late final String _pluginPackageDocsPath =
+    createTempSync('test_package_flutter_plugin').path;
 
 /// Version of dartdoc we should use when making comparisons.
 String get dartdocOriginalBranch {
   var branch = 'master';
   if (Platform.environment.containsKey('DARTDOC_ORIGINAL')) {
```
scheglov, line 140: We could extract `Platform.environment['DARTDOC_ORIGINAL']` as a local variable, and check it for null instead of using `containsKey`.

tool/grind.dart
```diff
@@ -138,49 +126,33 @@ Directory get testPackageExperiments =>
...
-    if (Platform.environment.containsKey('DARTDOC_PARAMS')) {
-      _extraDartdocParams
-          .addAll(Platform.environment['DARTDOC_PARAMS'].split(whitespace));
-    }
-  }
-  return _extraDartdocParams;
-}
+late final List<String> _extraDartdocParameters = [
+  if (Platform.environment.containsKey('DARTDOC_PARAMS'))
+    ...Platform.environment['DARTDOC_PARAMS']!.split(_whitespacePattern),
```
scheglov, line 151: Can we use `...?` here?

tool/grind.dart
```diff
@@ -355,23 +327,23 @@ class WarningsCollection {
   String _toKey(String text) {
     var key = text.replaceAll(tempDir, kTempDirReplacement);
     if (pubCachePath != null) {
-      key = key.replaceAll(pubCachePath, kPubCachePathReplacement);
+      key = key.replaceAll(pubCachePath!, kPubCachePathReplacement);
```
scheglov, line 330: Maybe declare `final pubCachePath = this.pubCachePath;` to avoid null check.

tool/grind.dart
```diff
@@ -355,23 +327,23 @@ class WarningsCollection {
...
+      key = key.replaceAll(kPubCachePathReplacement, pubCachePath!);
     }
     return key;
   }
 
   void add(String text) {
     var key = _toKey(text);
     warningKeyCounts.putIfAbsent(key, () => 0);
-    warningKeyCounts[key]++;
+    warningKeyCounts[key] = warningKeyCounts[key]! + 1;
```
scheglov, line 346: Consider using `Map.update(..., ifAbsent: () => 1)`.

tool/grind.dart
```diff
@@ -462,11 +433,11 @@ Future<void> compareSdkWarnings() async {
       Directory.systemTemp.createTempSync('dartdoc-comparison-sdkdocs');
   var originalDartdoc = createComparisonDartdoc();
   var currentDartdocSdkBuild = _buildSdkDocs(
-      sdkDocsDir.path, Future.value(Directory.current.path), 'current');
+      _sdkDocsDir.path, Future.value(Directory.current.path), 'current');
   var originalDartdocSdkBuild =
       _buildSdkDocs(originalDartdocSdkDocs.path, originalDartdoc, 'original');
   var currentDartdocWarnings = jsonMessageIterableToWarnings(
-      await currentDartdocSdkBuild, sdkDocsDir.absolute.path, null, 'HEAD');
+      await currentDartdocSdkBuild, _sdkDocsDir.absolute.path, null, 'HEAD');
```
scheglov, line 440: It would be nice to make all these getters return absolute objects. Even if we sometimes print relative, it should be an explicit operation - print something relative.

scheglov (approved):

tool/grind.dart
```diff
@@ -60,74 +55,67 @@ Directory createTempSync(String prefix) =>
...
 
-Directory _dartdocDocsDir;
-
-Directory get dartdocDocsDir => _dartdocDocsDir ??= createTempSync('dartdoc');
-
-Directory _dartdocDocsDirRemote;
-
-Directory get dartdocDocsDirRemote =>
-    _dartdocDocsDirRemote ??= createTempSync('dartdoc_remote');
+late final String _dartdocDocsPath = createTempSync('dartdoc').path;
```
parlough, line 108: Since these are top-level final variables, they are already evaluated lazily. You can remove the `late` here and in similar locations.

PR #2872, opened 2021-12-31
Minor cleanup to lints and some top level files

parlough:


lib/src/comment_references/parser.dart
```diff
@@ -65,7 +66,7 @@ class StringTrie {
   }
 }
 
-late final StringTrie operatorParseTrie = () {
```
srawlins, line 68: This changes the semantics, yes? It is no longer lazily evaluated, e.g. in

```
class C {
  final f = () {
    print('yay');
    return 7;
  }();
}

void main() {
  C();
  print('boo');
  C().f;
}
```

we get "yay" "boo" "yay" (rather than "yay" "boo"). Is this trie used unconditionally?

lib/src/special_elements.dart
```diff
@@ -99,12 +99,12 @@ class SpecialClasses {
 
   /// Add a class object that could be special.
   void addSpecial(Class aClass) {
-    if (_specialClassDefinitions.containsKey(aClass.name)) {
-      var d = _specialClassDefinitions[aClass.name]!;
-      if (d.matchesClass(aClass)) {
-        assert(!_specialClass.containsKey(d.specialClass) ||
-            _specialClass[d.specialClass] == aClass);
-        _specialClass[d.specialClass] = aClass;
+    var def = _specialClassDefinitions[aClass.name];
```
srawlins, line 102: Let's go ahead and call this 'definition'.

https://dart.dev/guides/language/effective-dart/documentation#avoid-abbreviations-and-acronyms-unless-they-are-obvious

srawlins (commented):
All good stuff, thanks!

lib/src/comment_references/parser.dart
parlough: Since this is a top-level field with an initializer it is already evaluated lazily. I added [`unnecessary_late`](https://dart-lang.github.io/linter/lints/unnecessary_late.html) to catch this but that won't be included until 2.16.

From the spec:
> Static variable declarations with an initializing expression are initialized
lazily (8.1).

srawlins: Ah I see. It looked indented by two spaces but I think its just the diff `- ` text.

srawlins (approved):
Thanks!

PR #2874, opened 2021-12-31
Migrate the generated AOT renderers to null safety

srawlins:


parlough (approved):
Seems straightforward enough to understand and good to me :)

Bye bye extra lines

scheglov (approved):

PR #2875, opened 2021-12-31
Cleanup categorization to simplify template generation migration

parlough:
Clean up model/categorization migration to simplify eventual template generation migration by changing `displayedCategories` to not contain nullable `Category` objects, as it was already assuming there weren't any `null` anyway.

\cc @srawlins This may slightly simplify the template generation as it seems to be making this assumption already anyway.

### DIFF
--- a/lib/src/model/categorization.dart
+++ b/lib/src/model/categorization.dart
@@ -55,11 +55,11 @@ abstract class Categorization implements ModelElement {
 
   /// Either a set of strings containing all declared subcategories for this symbol,
   /// or 'null' if none were declared.
-  late final List<String>? subCategoryNames = () {
+  List<String>? get subCategoryNames {
     // TODO(jcollins-g): avoid side-effect dependency
     if (_subCategoryNames == null) documentationLocal;
     return _subCategoryNames;
-  }();
+  }
 
   @override
   bool get hasCategoryNames => categoryNames?.isNotEmpty ?? false;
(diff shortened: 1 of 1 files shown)


lib/src/model/categorization.dart
```diff
@@ -49,23 +50,23 @@ abstract class Categorization implements ModelElement {
     return rawDocs;
   }
 
-  bool get hasSubCategoryNames => subCategoryNames!.isNotEmpty;
+  bool get hasSubCategoryNames => subCategoryNames?.isNotEmpty ?? false;
   List<String>? _subCategoryNames;
 
   /// Either a set of strings containing all declared subcategories for this symbol,
```
srawlins, line 56: Hmm, returns a List.... 🤷 ; to fix later I suppose.

lib/src/model/categorization.dart
```diff
@@ -75,40 +76,42 @@ abstract class Categorization implements ModelElement {
...
+  /// Either a URI to documentation with samples,
+  /// or 'null' if one was not declared.
   String? get samples {
     // TODO(jcollins-g): avoid side-effect dependency
     if (_samples == null) documentationLocal;
     return _samples;
   }
 
-  Iterable<Category?>? _categories;
+  late final Iterable<Category> categories = () {
```
srawlins, line 98: 💯 Love it

srawlins (approved):

PR #2879, opened 2022-01-05
Correct some bad mustachio migration; migrate mustachio tests

srawlins:


lib/src/mustachio/renderer_base.dart
```diff
@@ -342,7 +342,7 @@ class SimpleRenderer extends RendererBase<Object?> {
 /// An individual property of objects of type [T], including functions for
 /// rendering various types of Mustache nodes.
 @immutable
-class Property<T> {
+class Property<T extends Object?> {
```
scheglov, line 345: Do you want it explicit? AFAIK `Object?` is the default.

scheglov (approved):

lib/src/mustachio/renderer_base.dart
srawlins: I think `dynamic` is still the default, and I was having nullability questions here. I like to just make it explicit with `Object?` so that dynamic does not sneak in.

PR #2902, opened 2022-01-17
Misc library null safety cleanup

parlough:
I also renamed some local variables which began with an underscore(`_`) in preparation for the [eventual inclusion](https://github.com/dart-lang/lints/issues/58) of `no_leading_underscores_for_local_identifiers` in the `recommended` lint set.


lib/src/model/library.dart
```diff
@@ -469,53 +452,40 @@ class Library extends ModelElement with Categorization, TopLevelContainer {
...
-      // ambiguous.  dart-lang/dartdoc#2683.
-      for (var prefixEntry in prefixToLibrary.entries) {
-        if (!_referenceChildren!.containsKey(prefixEntry.key)) {
-          _referenceChildren![prefixEntry.key] = prefixEntry.value.first;
-        }
-      }
-    }
-    return _referenceChildren!;
-  }
+  Map<String, CommentReferable> get referenceChildren => _referenceChildren;
```
srawlins, line 488: Can this be removed, and the private field be made public?

srawlins (approved):

PR #2908, opened 2022-01-20
Remove unused optional name parameter

parlough:
This was causing a CI failure on new dev builds. The diagnostic must have recently expanded its reach.


lib/src/special_elements.dart
```diff
@@ -41,8 +44,8 @@ class _SpecialClassDefinition {
   /// can be found.
   final String libraryName;
 
-  /// The package name in which this [ModelElement] can be found.
-  final String packageName;
+  // /// The package name in which this [ModelElement] can be found.
+  // final String packageName;
```
srawlins, line 48: Remove?

PR #2913, opened 2022-01-21
Misc CI cleanup

parlough:


srawlins (approved):
dart pub dart pub dart pub dart pub dart pub

lgtm :)

PR #2914, opened 2022-01-21
Some minor List to Set improvements

parlough:


srawlins (commented):
Can you confirm real quick for each of these that they are better as Sets, e.g. because we just call `contains` on them or something?

parlough:
@srawlins Yeah we only call `contains` or `isEmpty` on these, or had converted them to sets to calculate the difference with other sets.

srawlins (approved):

PR #2916, opened 2022-01-21
Misc documentable and documentation cleanup

parlough:


srawlins:
This one I want to review more carefully; I think it was intentional that documentation be nullable (null means "doesn't have any" but I need to look. This will take a few days.

parlough:
@srawlins I thought about this too, but couldn't find a situation where `null` could propagate through. I may have missed a case though. It seems we instead we used an empty string in a lot of places. With null safety and Mustachio able to handle `null` values, perhaps I should restructure this to return `null` with lack of documentation. 

I'll take another look and see what makes the most sense, perhaps splitting pieces into smaller PRs.

PR #2920, opened 2022-01-25
Fix PropertyAccess crash in comment reference

srawlins:


lib/src/comment_references/model_comment_reference.dart
```diff
@@ -123,3 +123,18 @@ class _ModelCommentReferenceImpl implements ModelCommentReference {
   late final List<CommentReferenceNode> parsed =
       CommentReferenceParser(codeRef).parse();
 }
+
+extension on CommentReferableExpression {
+  Element? get element {
+    var expression = this;
+    if (expression is PropertyAccess) {
```
scheglov, line 130: Might be better to sort these `if is X`.

lib/src/comment_references/model_comment_reference.dart
```diff
@@ -123,3 +123,18 @@ class _ModelCommentReferenceImpl implements ModelCommentReference {
   late final List<CommentReferenceNode> parsed =
       CommentReferenceParser(codeRef).parse();
 }
+
+extension on CommentReferableExpression {
+  Element? get element {
+    var expression = this;
```
scheglov, line 129: This name is fine, just I just note that in the analyzer we tend to name it `final self = this;`.

scheglov (approved):

lib/src/comment_references/model_comment_reference.dart
srawlins: done

lib/src/comment_references/model_comment_reference.dart
srawlins: self is better; some of these sub-types have an expression field.

Issue #2928, opened 2022-02-02
Remove `[...]` generated links


kevmoo:
They add a lot of noise to the output, without much benefit.

PR #2929, opened 2022-02-02
Drop the [...] link on index pages

kevmoo:
Fixes https://github.com/dart-lang/dartdoc/issues/2928

### DIFF
--- a/lib/src/markdown_processor.dart
+++ b/lib/src/markdown_processor.dart
@@ -357,14 +357,14 @@ class MarkdownDocument extends md.Document {
             linkResolver: linkResolver,
             imageLinkResolver: imageLinkResolver);
 
+  /// Parses markdown text, collecting the first [md.Node] or all of them
+  /// if [processFullText] is `true`.
   List<md.Node> parseMarkdownText(String text, bool processFullText) {
     var lines = LineSplitter.split(text).toList();
     md.Node? firstNode;
     var nodes = <md.Node>[];
     for (var node in _IterableBlockParser(lines, this).parseLinesGenerator()) {
-      if (firstNode != null) {
-        if (!processFullText) break;
-      }
+      if (firstNode != null && !processFullText) break;
       firstNode ??= node;
       nodes.add(node);
     }
--- a/lib/src/markdown_processor.dart
+++ b/lib/src/markdown_processor.dart
@@ -357,25 +357,19 @@ class MarkdownDocument extends md.Document {
             linkResolver: linkResolver,
             imageLinkResolver: imageLinkResolver);
 
-  /// Parses markdown text, collecting the first [md.Node] or all of them
-  /// if [processFullText] is `true`. If more than one node is present,
-  /// then [DocumentationParseResult.hasExtendedDocs] will be set to `true`.
-  DocumentationParseResult parseMarkdownText(
-      String text, bool processFullText) {
-    var hasExtendedContent = false;
+  List<md.Node> parseMarkdownText(String text, bool processFullText) {
     var lines = LineSplitter.split(text).toList();
     md.Node? firstNode;
     var nodes = <md.Node>[];
     for (var node in _IterableBlockParser(lines, this).parseLinesGenerator()) {
       if (firstNode != null) {
-        hasExtendedContent = true;
         if (!processFullText) break;
       }
       firstNode ??= node;
       nodes.add(node);
     }
     _parseInlineContent(nodes);
-    return DocumentationParseResult(nodes, hasExtendedContent);
+    return nodes;
   }
 
   // From package:markdown/src/document.dart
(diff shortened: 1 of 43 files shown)


kevmoo:
Feel free to squash/merge and delete the branch if you're happy

srawlins:
What does a chopped off doc comment look like with out `[...]`?

kevmoo:
It's almost ALWAYS a full sentence. See https://api.dart.dev/stable/2.16.0/index.html

I'd love to see an example where it isn't....

parlough:
I thought it just shows the first line then shows this if there are more lines?

https://github.com/dart-lang/dartdoc/blob/master/lib/src/markdown_processor.dart#L371

kevmoo:
> I thought it just shows the first block (which I believe can be multiple source lines?) then showed this if there are more blocks?
> 
> https://github.com/dart-lang/dartdoc/blob/master/lib/src/markdown_processor.dart#L371

Godo catch, @parlough – did some more cleanup!

lib/src/markdown_processor.dart
```diff
@@ -357,25 +357,19 @@ class MarkdownDocument extends md.Document {
             linkResolver: linkResolver,
             imageLinkResolver: imageLinkResolver);
 
-  /// Parses markdown text, collecting the first [md.Node] or all of them
-  /// if [processFullText] is `true`. If more than one node is present,
-  /// then [DocumentationParseResult.hasExtendedDocs] will be set to `true`.
-  DocumentationParseResult parseMarkdownText(
-      String text, bool processFullText) {
-    var hasExtendedContent = false;
```
parlough, line 365: Can you add the docs back?

```dart
/// Parses markdown text, collecting the first [md.Node] 
/// or all of them if [processFullText] is `true`.
```

lib/src/markdown_processor.dart
```diff
@@ -357,25 +357,19 @@ class MarkdownDocument extends md.Document {
...
+  List<md.Node> parseMarkdownText(String text, bool processFullText) {
     var lines = LineSplitter.split(text).toList();
     md.Node? firstNode;
     var nodes = <md.Node>[];
     for (var node in _IterableBlockParser(lines, this).parseLinesGenerator()) {
       if (firstNode != null) {
-        hasExtendedContent = true;
         if (!processFullText) break;
       }
```
parlough, line 367:
```dart
if (firstNode != null & !processFullText) {
  break;
}

parlough (commented):
I love the extra cleanup! I think you'll need to rerun `dart run grinder build` again.

lib/src/markdown_processor.dart
kevmoo: ack

lib/src/markdown_processor.dart
kevmoo: ack

kevmoo:
Thoughts @srawlins ?

srawlins:
Ok it does look like its always a full sentence because its the first line. What is the indication that there is more doc comment beyond what is shown?

kevmoo:
> Ok it does look like its always a full sentence because its the first line. What is the indication that there is more doc comment beyond what is shown?

I don't think it's necessary. If folks want to know more, they will click on the entry. I don't see this affordance in any other API docs I've reviewed (and I've looked at pretty much all of them 😄 )

parlough:
I took a look at the SDK docs without the [...] and I have to say the pages look less cluttered and my attention is drawn to each member name instead. I think clicking on the member name for more information is pretty intuitive as it already presents as a link and in most cases the first line is sufficient anyway.

srawlins:
Yeah I'd definitely prefer a review from a UX expert.

kevmoo:
> Yeah I'd definitely prefer a review from a UX expert.

Understood, @srawlins . I hope to engage UX here for bigger items. In the meantime, look at Microsoft, Android, Apple, etc – no one uses the `[...]` anywhere. For places where we're clearly being unnecessarily different than other thought leaders here, we should just get in line – unless there's a good reason not to.

srawlins:
I agree.

Can you show examples which parallel our style of printing a 1-line summary of a doc, and a complete doc on a different page? I couldn't find any examples in Android, but I see that Rust uses the words, "Read more": https://doc.rust-lang.org/core/primitive.array.html#method.borrow_mut

kevmoo:
Microsoft: https://docs.microsoft.com/en-us/dotnet/api/system?view=net-6.0

The output has a "read more" header section, which seems gratuitous.

Android:
https://developer.android.com/reference/androidx/test/espresso/accessibility/package-summary

Shows the first line of a description. Clicking on AccessibilityChecks
<https://developer.android.com/reference/androidx/test/espresso/accessibility/AccessibilityChecks>
takes
you to the full description.

srawlins (approved):
OK that's good enough for me. Mind re-basing and regenerating those files, Kevin?

kevmoo:
@srawlins – note version change here, too!

PR #2938, opened 2022-02-07
add runtime perf stats; remove some async

devoncarew:
This was work done during a very casual look at the runtime perf of dartdoc; this PR does not contain any actual perf improvements.

- add some simple runtime perf stats, to how much time we're spending in various parts of doc generation
- remove some uses of async (where we could just collapse to simple sync calls, and then iteratively remove async from callers and APIs; this makes the perf a little easier to trace in things like DevTools
- fix a few minor issues in dartdoc's own API comments

Here's some sample perf stats (from documenting dartdoc itself):

```
generateDocs                     31,323ms
  buildPackageGraph              2,3106ms
    _calculatePackageMap             18ms
    getLibraries                 16,787ms
    initializePackageGraph        6,299ms
  generator.generate              1,865ms
  validateLinks                   6,343ms
```

### DIFF
--- a/bin/dartdoc.dart
+++ b/bin/dartdoc.dart
@@ -12,7 +12,7 @@ import 'package:dartdoc/options.dart';
 /// Analyzes Dart files and generates a representation of included libraries,
 /// classes, and members. Uses the current directory to look for libraries.
 Future<void> main(List<String> arguments) async {
-  var config = await parseOptions(pubPackageMetaProvider, arguments);
+  var config = parseOptions(pubPackageMetaProvider, arguments);
   if (config == null) {
     // There was an error while parsing options.
     return;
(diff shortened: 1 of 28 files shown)


srawlins (approved):
Wow, very cool; love removing all the unnecessary async.

How did you track those down? I'd love to prevent it in the future. Can you think of a way to track unnecessary async statically? Maybe I can bake it into whole-world analysis.

kevmoo:
There's a lint for await only Future, right?

devoncarew:
> How did you track those down? I'd love to prevent it in the future. Can you think of a way to track unnecessary async statically? Maybe I can bake it into whole-world analysis.

I didn't do anything fancy - just remove async from simple methods and APIs, trying to fix up all the resulting static analysis issues, and backing a change out if it bottoms out somewhere that really does need to be async (package:package_config was one of those places - there are some APIs there where we could easily offer sync versions).

PR #2968, opened 2022-02-25
Parse-and-resolve fewer files

srawlins:
Two features, plus small code changes:

* `_parseLibraries` is called exactly two times. Retain the "known parts" between the two calls.
* Paths are added to the `files` variable within the for loop. The for loop is inside a do/while. Track processed libraries to prevent parsing multiple times.

In a unit test I found that this reduced the total number of resolution requests from 225 to 130. In documenting flutter, I found that this reduced the total number of resolution requests from 2946 to 2543.

In the two flutter runs, the time to document was reduced by 20 seconds, but I did not do multiple trials.


srawlins:
I definitely feel there is more to do in package_builder.dart. I find it extremely difficult to reason about `_parseLibraries` and the difference between regular files and special files, and the various loops.

scheglov (approved):

PR #2981, opened 2022-03-05
Enum constructors

srawlins:
This adds support for enhanced enum constructors. A few small refactors are done in Container. Made `Container.extraReferenceChildren` abstract, so each class must intentionally declare some implementation.

Enum constructors are displayed, have good info, and can be referenced.

### DIFF
--- a/test/enum_test.dart
+++ b/test/enum_test.dart
@@ -388,6 +388,47 @@ enum E {
       expect(method1.documentationComment, '/// Doc comment.');
     });
 
+    test("an enhanced enum's constructors are documented", () async {
+      var library = await bootPackageWithLibrary('''
+enum E {
+  one.named(1),
+  two.named(2);
+
+  final int x;
+
+  /// A named constructor.
+  const E.named(this.x);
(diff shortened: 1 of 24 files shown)


test/enum_test.dart
```diff
@@ -388,6 +386,45 @@ enum E {
...
+''');
+      var namedConstructor =
+          library.enums.named('E').constructors.named('E.named');
+
+      expect(namedConstructor.isFactory, false);
+      expect(namedConstructor.fullyQualifiedName, 'enums.E.named');
+      expect(namedConstructor.nameWithGenerics, 'E.named');
+      expect(namedConstructor.documentationComment, '/// A named constructor.');
+    });
+    test("has an 'index' getter which can be referenced", () async {
```
scheglov, line 409: The test name does not look to correspond to the content.

scheglov (approved):

test/enum_test.dart
srawlins: Thanks; done.

PR #2994, opened 2022-03-24
Update analyzer to v3.4.0 + fix warnings/deprecations

DanTup:
I recently landed changes in `package:analyzer` that made `lineInfo` non-nullable in a few places. Since then, when running some tests from source in the SDK repo I see repeated warnings printed because dartdoc uses `?` in places that are no longer required.

I believe the fix is to upgrade the `analyzer` package here (and then update the `dartdoc` package in the SDK).

The tests all pass locally (running `dart pub global run grinder test`), although I'm not sure whether whether forcing consumers of this package onto this version is reasonable yet. If not, perhaps some `// ignore`'s could be added for the `?.lineInfo` calls in the meantime?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## 5.1.0-dev
 * Removed superfluous `[...]` links. (#2928)
+* Move to analyzer 3.4.0.
 
 ## 5.0.1
 * Add support for new VM service message. (#2931)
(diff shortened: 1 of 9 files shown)


srawlins:
Thanks!

Looks like you need to rebuild the generated code (with something like `dart run build_runner build`.

DanTup:
Oops, done!

srawlins (approved):
Thanks!

PR #3003, opened 2022-03-28
Add RuntimeStats for validation file count and file write count

srawlins:
* Move RuntimeStats to its own library
* Add stats for each time a file is written, per element type
* Add stats for each time a file is read for validation
* Alphabetize the `generate*` methods


lib/src/generator/dartdoc_generator_backend.dart
```diff
@@ -204,12 +202,26 @@ abstract class DartdocGeneratorBackend implements GeneratorBackend {
         sidebarForContainer.getRenderFor);
     var content = templates.renderMethod(data);
     write(writer, method.filePath, data, content);
+    runtimeStats.accumulators.update('writtenMethodFileCount', (c) => c + 1);
```
scheglov, line 205: We could add a method `incrementAccumulator` to `_RuntimeStats`, and maybe make the `Map accumulators` itself private.

lib/src/generator/generator_frontend.dart
```diff
@@ -34,6 +35,19 @@ class GeneratorFrontEnd implements Generator {
   /// Traverses the package graph and generates documentation for all contained
   /// elements.
   List<Indexable> _generateDocs(PackageGraph packageGraph, FileWriter writer) {
+    runtimeStats.accumulators['writtenCategoryFileCount'] = 0;
```
scheglov, line 38: This is a lot of repetitive code. We could add a method `resetAccumulators(Set<String> names)` in `_RuntimeStats`.

lib/src/generator/dartdoc_generator_backend.dart
srawlins: Done

lib/src/generator/generator_frontend.dart
srawlins: Done

lib/src/generator/generator_frontend.dart
```diff
@@ -34,6 +35,21 @@ class GeneratorFrontEnd implements Generator {
   /// Traverses the package graph and generates documentation for all contained
   /// elements.
   List<Indexable> _generateDocs(PackageGraph packageGraph, FileWriter writer) {
+    runtimeStats.resetAccumulators([
```
scheglov, line 38: It would be slightly better to use constant `Set`, because then the compiler will complain if there are duplicates :-)

scheglov (approved):

lib/src/generator/generator_frontend.dart
srawlins: Good point! Done

PR #3008, opened 2022-04-06
Do not store the AstNode on ModelNode; eagerly grab the source indexes

srawlins:
I think it is probably good to keep `sourceCode` as late final, because I suspect we generate many ModelNodes whose sources are never fetched, and this code ** ** **reads the file back from disk** ** **!!

In any case, keeping the AstNode on as a field has to be a bad idea. Just do a little logic to store the `sourceEnd` and `sourceOffset`.

### DIFF
--- a/lib/src/model/model_node.dart
+++ b/lib/src/model/model_node.dart
@@ -14,12 +14,33 @@ class ModelNode {
   final List<ModelCommentReference> commentRefs;
   final Element element;
   final ResourceProvider resourceProvider;
+  final int _sourceEnd;
+  final int _sourceOffset;
 
-  final AstNode? _sourceNode;
+  factory ModelNode(
+      AstNode? sourceNode, Element element, ResourceProvider resourceProvider) {
+    var commentRefs = _commentRefsFor(sourceNode, resourceProvider);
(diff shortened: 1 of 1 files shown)


lib/src/model/model_node.dart
```diff
@@ -36,26 +57,13 @@ class ModelNode {
...
-    if (enclosingSourceNode is VariableDeclaration) {
-      enclosingSourceNode = enclosingSourceNode.parent!.parent;
-      assert(enclosingSourceNode is FieldDeclaration ||
-          enclosingSourceNode is TopLevelVariableDeclaration);
-    }
-
-    var sourceEnd = enclosingSourceNode!.end;
-    var sourceOffset = enclosingSourceNode.offset;
-
     var contents = model_utils.getFileContentsFor(element, resourceProvider)!;
```
scheglov, line 64: You should not do this, the analyzer already gives `ResolvedUnitResult.content`.

scheglov (approved):

lib/src/model/model_node.dart
srawlins: I agree 100%. This is shocking that we re-read everything.

PR #3009, opened 2022-04-06
Move many static declarations to top-level; avoid RegExp

srawlins:
We were HTML-escaping static strings like "///" and "/**" repeatedly in a loop, for every single line of every single element's source code.

So, static-ify that.

Also, to determine if a URI was a package's implementation, we unnecessarily use a RegExp for `/\`

but backslashes are not allowed...


lib/src/model_utils.dart
```diff
@@ -108,49 +109,67 @@ bool hasPrivateName(Element e) {
...
   if (e is LibraryElement) {
+    if (e.identifier.startsWith('dart:_') ||
+        e.identifier.startsWith('dart:nativewrappers/') ||
+        'dart:nativewrappers' == e.identifier) {
+      return true;
+    }
     var elementLocation = e.location;
     if (elementLocation != null) {
-      var locationParts = elementLocation.components[0].split(slashes);
+      var elementUri = elementLocation.components[0];
```
scheglov, line 120: Use `e.source.uri` instead.

lib/src/model_utils.dart
```diff
@@ -108,49 +109,67 @@ bool hasPrivateName(Element e) {
...
+        return true;
+      }
     }
   }
   return false;
 }
 
+/// Returns whether [uri] is a package's implementation, a URI starting with
+/// "package:", and with a first path segment of "src".
+bool _uriIsPackageImplementation(String uri) {
```
scheglov, line 132: And use `Uri` here, and `Uri.pathSegments`. If we can use standard URI implementation, we should, unless there are good reasons.

lib/src/model_utils.dart
srawlins: Done

lib/src/model_utils.dart
srawlins: Done

lib/src/model_utils.dart
```diff
@@ -108,20 +109,16 @@ bool hasPrivateName(Element e) {
...
-          locationParts[0].startsWith('package:') &&
-          locationParts[1] == 'src') return true;
+    if (e.identifier.startsWith('dart:_') ||
+        e.identifier.startsWith('dart:nativewrappers/') ||
+        'dart:nativewrappers' == e.identifier) {
+      return true;
+    }
+    var elementUri = e.source.uri;
+    // TODO(jcollins-g): Implement real cross package detection
+    if (elementUri.scheme == 'package' && elementUri.pathSegments[1] == 'src') {
```
scheglov, line 120: Practically, any `UriResolver` that understands `package:` URIs will return `null` if there are less than `2` segments, so this code should be fine. But purely from `Uri` point of view, it is possible to parse `package:foo`. So, if we could make it safer... Something like `elementAtOrNull`, similar to `firstOrNull` from `IterableExtension` from `package:collection`. Unfortunately it does not exist yet :-)

Anyway, just some random notes.

scheglov (approved):

lib/src/model_utils.dart
srawlins: Yeah I decided the simplicity was worth ignoring `package:foo`, haha. We'll see if it ever causes a problem.

PR #3011, opened 2022-04-07
Add a web frontend

devoncarew:
- infra for compiling and including a dart based web front end for dartdoc
- commit the compiled frontend

@srawlins - this is a first step of what we discussed today; a follow-up step would be to convert some or all of the js library into dart

Done as 2 commits for easier review.


srawlins (approved):

srawlins:
I have a web server in-progress which could use this web frontend.

PR #3016, opened 2022-04-08
convert lib/resources/script.js to Dart

devoncarew:
Convert lib/resources/script.js to Dart; this PR:
- removes the scroll position save / restore code
- moves the highlight.js initialization to Dart code
- converts the navbar event handling code to dart
- converts the search box handling code to dart

cc @srawlins 

This is more a less a straight 1:1 conversion from the JS code. Some follow up work would be to make the code follow more of a Dart style; later work may change the behavior (improvements to the search box? programmatically building the left nav?).


devoncarew:
Hmm, it looks like the `web/docs.dart` signature code has some issues on windows.

devoncarew:
OK, the sig on windows is incorrect, even after removing line endings. I don't think there are any path char issues - not sure why the platform difference.

```
  current files: 4862D84995B6C7E2B74C025C713BBF21
  cached sig   : 05FD8D099278ECC2307AAFD453A1E02E
  failed: The web frontend (web/docs.dart) needs to be recompiled; rebuild it with "grind build-web" or "grind build".
```

srawlins:
It wouldn't be bad if the signature stuff just wasn't run on Windows in CI. It would sort of limit our abilities to accept PRs from a person using Windows, maybe...

devoncarew:
> It wouldn't be bad if the signature stuff just wasn't run on Windows in CI. It would sort of limit our abilities to accept PRs from a person using Windows, maybe...

I think I fixed it; the file listing was in a different order :) I normalize it by sorting the list of files.

devoncarew:
@srawlins - I believe this is fully functional for what's been converted over, and I've removed the JS code that's been converted. I.e., the functionality is consistent, just some is now in Dart.

Do you want to review the PR now, or, wait until the search stuff is converted?

srawlins:
I think if it's sound, even if in-progress, that sounds great. I can review.

devoncarew:
I think it's ready for review now (was was able to convert over the search code as well).

srawlins:
Do we know that this is runnable at pub.dev? Meaning if I cut a release with this PR, and pub.dev used it, the Dart web app is all good to go and run on pub.dev?

I don't have a specific reason why it wouldn't. Since we're already serving up a bit of JS at pub.dev.

web/search.dart
```diff
@@ -0,0 +1,445 @@
...
+    suggestionElements = [];
+    presentation.text = '';
+
+    if (suggestions.length < minLength) {
+      setHint(null);
+      hideSuggestions();
+      return;
+    }
+
+    for (var i = 0; i < suggestions.length; i++) {
```
srawlins, line 289: Can this be a for-in loop?

web/sidenav.dart
```diff
@@ -0,0 +1,22 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:html';
+
+/// Initialize the sidenav event handlers.
+void init() {
+  final document = window.document;
+
+  var leftNavToggle = document.getElementById('sidenav-left-toggle');
```
srawlins, line 11: Is this functionality new? I can't find other references to this class.

srawlins:
Actually, the other question might be the more important one: Are dartdocs still visible locally? I am unfamiliar with the sidenav navigation JS.

devoncarew:
> Do we know that this is runnable at pub.dev? Meaning if I cut a release with this PR, and pub.dev used it, the Dart web app is all good to go and run on pub.dev?

I would say we're good to go, as long as the pub.dev server doesn't depend on specific JS resources existing in specific places.

cc'ing in @jonasfj and @isoos to comment on whether these changes are compatible w/ pub docs serving. The TLDR from the POV of serving the docs is that the file `doc/api/static-assets/script.js` will no longer be generated, and instead we'll generate `doc/api/static-assets/docs.dart.js`. An additional file - `docs.dart.js.map` will also exist in that assets directory, but it's not strictly speaking necessary to serve that file (definitely ok to thought). The generated html files have been updated to refer to the correct resources.

devoncarew:
> Actually, the other question might be the more important one: Are dartdocs still visible locally? I am unfamiliar with the sidenav navigation JS.

Yup, the docs are visible and working locally. There's no real functionality change with this PR (modulo removing the code to save and restore the scroll positions).

web/sidenav.dart
devoncarew: It was in and around here: https://github.com/dart-lang/dartdoc/blob/master/lib/resources/script.js#L2 - just made a bit more terse.

web/search.dart
devoncarew: Happy to, however, I was planning to make a pass through in a follow-on PR to make the code much more Dart like (I avoided doing too much style conversion in this PR in order break the work into smaller chunks).

srawlins: That sounds good.

srawlins (approved):

isoos:
> he TLDR from the POV of serving the docs is that the file doc/api/static-assets/script.js will no longer be generated, and instead we'll generate doc/api/static-assets/docs.dart.js. An additional file - docs.dart.js.map will also exist in that assets directory, but it's not strictly speaking necessary to serve that file (definitely ok to thought). The generated html files have been updated to refer to the correct resources.

This is OK from `pub.dev`'s PoV. We may need to update a test that checks the list of output files, but that is not unusual.

jonasfj:
The size increase from 14kb to 444kb makes me a bit sad. I guess it's not a huge concern.

It might make loading the page slower especially on low powered devices. It might be worth checking the lighthouse score before/after..

isoos:
> The size increase from 14kb to 444kb makes me a bit sad. I guess it's not a huge concern.

It is also worth to take a look what makes the size to blow up. We've removed `package:intl` and replaced the single method we've used with a custom implementation, and gained almost 100 kb in total size (https://github.com/dart-lang/pub-dev/pull/5526)

devoncarew:
The implementation for the web frontend itself is pretty simple: https://github.com/dart-lang/dartdoc/tree/master/web (but still worth looking to see if there are areas contributing to compiled code size).

parlough:
Compiling with most optimizations enabled is pretty low hanging fruit: https://github.com/dart-lang/dartdoc/pull/3028

PR #3030, opened 2022-05-02
Fixed: YouTube directive respects width/height argument

goderbauer:
The `{@youtube}` directive now respects the provided width/height arguments. They are used to configure the `max-height` / `max-width` of the video to ensure that the video can be smaller than the provided measurements if there isn't enough screen space (e.g. on mobile). The video will never be larger than the provided measurements.

Without the fix, the videos looked awkward (e.g. on wide screens) where they would take over the full width (ignoring the configured width) and push other important content below the fold (see screenshots below).

Fixes https://github.com/dart-lang/dartdoc/issues/2797.

## Before:
Screen Shot 2022-05-02 at 10 52 04 AM

## After:
Screen Shot 2022-05-02 at 11 51 49 AM

## Mobile (with this change, same as before):
Screen Shot 2022-05-02 at 11 52 15 AM


srawlins:
Looks like this fixes https://github.com/dart-lang/dartdoc/issues/2797 ?

srawlins (approved):

goderbauer:
> Looks like this fixes https://github.com/dart-lang/dartdoc/issues/2797 ?

It does. Added to PR description.

PR #3032, opened 2022-05-03
Create scorecards-analysis.yml

sealesj:
Enables scorecards for dartdoc repo as part security scanning effort on multiple dart-lang repos.

scorecards-analysis.yaml generated through integration with OSSF Scorecards

### DIFF
--- a/.github/workflows/scorecards-analysis.yml
+++ b/.github/workflows/scorecards-analysis.yml
@@ -1,0 +1,55 @@
+name: Scorecards supply-chain security
+on:
+  # Only the default branch is supported.
+  branch_protection_rule:
+  schedule:
+    - cron: '30 19 * * 2'
+  push:
+    branches: [ master ]
+
+# Declare default permissions as read only.
+permissions: read-all
+
+jobs:
+  analysis:
+    name: Scorecards analysis
+    runs-on: ubuntu-latest
+    permissions:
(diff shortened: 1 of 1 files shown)


devoncarew:
Can you provide more context for this PR? Is this related to https://github.com/dart-lang/.github/issues/16?

sealesj:
@devoncarew Yes, thank you for linking the issue. This is a part of the same issue.

devoncarew:
Can you add context - here or in the other issue - about what this will do? How will it be maintained? Any other useful details? We have ~100 repos in dart-lang/; having 50 lines of config file in each one - maintained separately - will not be cheap.

sealesj:
@devoncarew Enabling the scorecard analysis will allow for vulnerability alerting after scans triggered via a push to master. Maintenance for the versions of actions used in scorecards-analysis.yml itself should be taken care of by dependabot. Also not all repos in dart-lang are targeted to have this scorecard enabled, but rather just the top critical repos. cc @athomas Could you provide more context if necessary as well / if I missed anything?

athomas:
> No more questions (though I see you're sending PRs to pin versions in github actions files, so may want context for that later :) ). @srawlins is generally the reviewer here however; I'll add him.

See go/github-actions#actions for context on the pinning. It is required.

srawlins (approved):

PR #3034, opened 2022-05-12
fix exception for empty dartdoc_options.yaml file

timmaffett:
This PR fixes an exception that is thrown when dartdoc encounters an empty (or all comments) dartdoc_options.yaml file.

I discovered this bug when I commented all the lines of a dartdoc_options.yaml file and ran dartdoc.  An exception was thrown because the return value of loadYaml() on the 'empty' yaml file was null and this was passed directly to _YamlFileData() constructor which throws an exception because it doesn't accept null values.

When an empty file is found it is now just treated like it didn't exists in the first place.

### DIFF
--- a/lib/src/dartdoc_options.dart
+++ b/lib/src/dartdoc_options.dart
@@ -990,10 +990,14 @@ abstract class _DartdocFileOption<T> implements DartdocOption<T> {
             .pathContext
             .join(dir.path, 'dartdoc_options.yaml'));
         if (dartdocOptionsFile.exists) {
-          yamlData = _YamlFileData(
-              loadYaml(dartdocOptionsFile.readAsStringSync()),
-              resourceProvider.pathContext.canonicalize(dir.path));
-          break;
+          final yaml = loadYaml(dartdocOptionsFile.readAsStringSync());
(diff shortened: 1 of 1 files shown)


lib/src/dartdoc_options.dart
```diff
@@ -990,10 +990,14 @@ abstract class _DartdocFileOption<T> implements DartdocOption<T> {
...
-              loadYaml(dartdocOptionsFile.readAsStringSync()),
-              resourceProvider.pathContext.canonicalize(dir.path));
-          break;
+          final yaml = loadYaml(dartdocOptionsFile.readAsStringSync());
+          // [loadYaml()] will return null for empty (or all comment) yaml files,
+          // so we must check for that case.
+          if (yaml != null) {
+            yamlData = _YamlFileData(
+                yaml, resourceProvider.pathContext.canonicalize(dir.path));
+            break;
```
srawlins, line 999: I feel like we should still break, even if `yaml == null`, right? I think once we hit a dartdoc options file, even if its empty, we should stop walking ancestors.

timmaffett: Ok, I have changed that so we now break even when yaml==null.

srawlins (approved):
Thanks much!

PR #3042, opened 2022-05-18
Add the docs.dart.js.map file to published package

srawlins:
Fixes bug in dartdoc 5.1.1

CC @goderbauer

### DIFF
--- a/.pubignore
+++ b/.pubignore
@@ -20,4 +20,3 @@ pubspec.lock
 *.ipr
 *.iws
 *.js.deps
-*.js.map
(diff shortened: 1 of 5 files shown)


devoncarew (approved):
The map file might also have been listed in the generated resources? If so, we could also remove it from there.

srawlins:
> The map file might also have been listed in the generated resources? If so, we could also remove it from there.

What list is this?

devoncarew:
I was thinking of the files generated from the mustache templates, but I don't see it there. I do see a reference from `lib/src/generator/html_resources.g.dart` however.

srawlins:
Ah I see. Good point. It seems like it should be fine to publish this file though? For error reporting, debugging, developing, etc.

devoncarew:
Yeah, that sounds fine asl well.

PR #3045, opened 2022-05-21
inline the used package:charcode constants

devoncarew:
- Similar to https://github.com/dart-lang/markdown/pull/423, this PR in-lines the portions of package:charcode that dartdoc uses.
- This also revs the dep on package:markdown; that's not a critical aspect of this PR, but we will need to resolve against the latest `markdown` in order to not pull in `charcode` transitively.


srawlins (approved):
Thanks!!

PR #3082, opened 2022-07-13
Move FileWriter from being a parameter on 100 methods to a field on DartDocGeneratorBackend

srawlins:
I think how this got to its current state is that FileWriter needed the outputDir, so it had always been initialized well after the Dartdoc object, or the Generator frond-end object... so then it was a necessary parameter to every parameter on DartDocGeneratorBackend. Oy.

This was not _super_ straightforward to change it to be a field on DartDocGeneratorBackend, but well worth the re-working.


keertip (approved):
Awesome refactor! Totally makes sense to make FileWriter a field. Wonder how it got to being a parameter on sooo many methods :)

PR #3091, opened 2022-07-19
WIP - Only scope lookups

srawlins:
Just seeing what CI looks like on a wild change.


srawlins:
This is super stale and @kallentu has taken over this effort.

PR #3098, opened 2022-07-25
Make MarkdownDocument unnamed constructor to be private

srawlins:
Also move _makeLinkNode to be a static member


lib/src/markdown_processor.dart
```diff
@@ -15,6 +15,7 @@ import 'package:dartdoc/src/model/model.dart';
 import 'package:dartdoc/src/runtime_stats.dart';
 import 'package:dartdoc/src/warnings.dart';
 import 'package:markdown/markdown.dart' as md;
+import 'package:markdown/markdown.dart';
```
keertip, line 18: Duplicate import?

keertip (approved):

lib/src/markdown_processor.dart
srawlins: lol thanks! Done

PR #3099, opened 2022-07-25
Bump analyzer to 4.3.1

srawlins:


Levi-Lesches:
Interesting that [test (ubuntu-latest, dev, sdk-analyzer)](https://github.com/dart-lang/dartdoc/runs/7506716126?check_suite_focus=true#logs) failed yet the GitHub UI shows it as passed.

scheglov (approved):

lib/src/model/package_builder.dart
```diff
@@ -192,7 +192,9 @@ class PubPackageBuilder implements PackageBuilder {
         for (var export in element.libraryExports) {
           _addKnownFiles(export.exportedLibrary);
         }
-        for (var part in element.parts2) {
+        for (var part in element.parts2
+            .map((e) => e.uri)
+            .whereType<DirectiveUriWithUnit>()) {
           _knownFiles.add(part.source.fullName);
```
scheglov, line 198: No, here `part.source` would be the source where the `part` (type `PartElement`) directive is located, i.e. the library source, not the source of the included unit.

You need `element.parts2.map((e) => e.uri).whereType<DirectiveUriWithUnit>().map((uri) => uri.unit)` to get included `CompilationUnitElement`s.

srawlins: Thanks!

Issue #3047, opened 2022-05-23
Adopt badge design for all features


guidezpl:
All features could adopt the same styling as the Null safety feature badge

Currently | Proposal
--- | ---
<img width="222" alt="Screen Shot 2022-05-23 at 11 27 07" src="https://user-images.githubusercontent.com/6655696/169788711-aa7b5b8e-bf2a-4ff8-b8db-bc36e2b93d44.png"> | <img width="188" alt="Screen Shot 2022-05-23 at 11 24 14" src="https://user-images.githubusercontent.com/6655696/169788428-bdac5c2f-f165-4d06-aa35-5eaacff25a87.png">

Related https://github.com/dart-lang/dartdoc/issues/1286

PR #3101, opened 2022-07-28
Adopt badge design for all features #3047

klr981:
## Description:
- style.css : the "features" styling now contains the same style as the "feature" styling (The styling for the badge for Null Safety).  
- model_element_renderer.dart : The line changed now iterates each feature and makes it a HTML element type span containing the name of the feature for the element class and the text inside of the element.
- _features.html : To avoid the div in the template to adopt the styling of features the name was changed from "features" to "features-div".

Fixes [#3047](https://github.com/dart-lang/dartdoc/issues/3047)


parlough:
This standardization will be nice! Could you add a before and after screenshot to the PR as a reference?

Also, you can add something "Closes https://github.com/dart-lang/dartdoc/issues/3047" or "Fixes https://github.com/dart-lang/dartdoc/issues/3047" to the PR description so it automatically links the PR and issue and will close it on merge.

klr981:
### Screenshots for reference:

- Before the changes
Before

- After the changes
After

Also, I added the fix issue tag to the PR description, thank you for the tip!


> This standardization will be nice! Could you add a before and after screenshot to the PR as a reference?
> 
> Also, as a small note, you can add something "Closes #3047" or "Fixes #3047" to the PR description so it automatically [links the PR and issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) and then will close the issue automatically when merged.

devoncarew:
It looks like the text for annotations has changed? Here's the before and after for the annotations on main(), above:

Screen Shot 2022-07-29 at 10 57 11 AM

Screen Shot 2022-07-29 at 10 56 58 AM

And from those screenshots, it's not clear that we do want to use badge decorations for user annotations / all annotations? You may want to include some other screenshots (of a deprecated annotation, or other typical annotations) to better see what they would look like.

keertip:
We want to adopt the badge for features and not all user annotations. @klr981

lib/resources/styles.css
```diff
@@ -462,6 +462,7 @@ h1 .category {
   vertical-align: middle;
 }
 
+/* The badge under a declaration for things like "const", "read-only", etc. and for the badges inline like Null safety*/
```
srawlins, line 465: I think it would be beneficial to link to the items in the backend code, I think in feature.dart?

srawlins (changes requested):

klr981:
Screenshot 2022-08-02 1 34 45 PM

New changes now only changes the feature styling and not annotations.

lib/resources/styles.css
srawlins: Looks good.

lib/src/model/feature.dart
```diff
@@ -60,4 +60,6 @@ class Feature implements Privacy {
   static const overrideFeature = Feature('override', 3);
   static const overrideGetter = Feature('override-getter', 3);
   static const overrideSetter = Feature('override-setter', 3);
+
+  String  cssClassName() => 'feature';
```
srawlins, line 64: Looks like this file is not formatted?

srawlins, line 64: This should be a getter like `String get cssClassName`

srawlins (changes requested):
I suspect you will need to rebase from master and re-run `dart run build_runner build`.

lib/src/model/annotation.dart
```diff
@@ -74,4 +74,7 @@ class Annotation extends Feature with ModelBuilder {
 
   @override
   int get hashCode => annotation.hashCode;
+
+  @override
+  String get cssClassName => 'annotation';
```
srawlins, line 79: If we have no CSS rules associated with this class name, it would be better to just leave it blank.

srawlins (approved):

srawlins (approved):
Looks great!

PR #3105, opened 2022-08-02
Fork built-in features into _BuiltInFeature

srawlins:
* Make `Feature` abstract; built-in features are instantiated as a subclass. This formalizes the two types of Features: annotations, and built-in "features".
* `ElementFeatureNotFoundError.message` is a non-nullable string


keertip (approved):
+1

PR #3106, opened 2022-08-03
Move htmlBase from individual TemplateData classes to shared mixins

srawlins:
This formalization of the two htmlBase cases will help with the Single Page Application implementation.


keertip (approved):
Nice refactor!

PR #3107, opened 2022-08-03
Dartdoc Dark Mode

klr981:
### Description:
- theme.dart: has an event handler for the toggle. If the toggle is activated it removes first the current value and changes it to the new one. For example, if value="light-theme" and toggle is activated it changes the value to dark-theme. 
- style.css: it has two classes one called light and the other one dark. These two classes have variables that are named the same but hold different values depending on the theme. It also has the CSS for the theme toggle that is located in the top right corner. 
- _head.html: it contains the HTML code for the sun and moon icon representing the themes and the toggle button as well.
- docs.dart: theme.dart was added to it.

#### Light Theme
Light Theme2
Light Theme1

#### Dark Theme
Dark Theme2
Dark Theme1

Closes https://github.com/dart-lang/dartdoc/issues/1885


lib/src/generator/templates.aot_renderers_for_html.dart
```diff
@@ -2969,7 +2969,7 @@ String _deduplicated_lib_templates_html__head_html(
   buffer.write('''<!DOCTYPE html>
 <html lang="en">
 <head>
-  <meta charset="utf-8">
+  <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.7.0/css/font-awesome.min.css">  <meta charset="utf-8">
```
parlough, line 2972: It seems like including all of Font Awesome for two icons might be a lot, it's nice to keep the site download size down.

Since we include it already, could we use Material icons instead? They might match better as well. We might use them in a few places already, at least the [Source button](https://github.com/dart-lang/dartdoc/blob/master/lib/templates/html/_source_link.html#L2).

web/theme.dart
```diff
@@ -0,0 +1,63 @@
+import 'dart:html';
+
+void init() {
+  var theme = document.getElementById('theme') as InputElement;
+
+  var elementB = document.body;
+
+  theme.addEventListener('change', (event) {
```
srawlins, line 8: To keep things a little simpler, we can say something here like, `if (elementB == null) return;`

web/theme.dart
```diff
@@ -0,0 +1,63 @@
+import 'dart:html';
+
+void init() {
+  var theme = document.getElementById('theme') as InputElement;
+
+  var elementB = document.body;
```
srawlins, line 6: We can just call this variable something like `body` or `bodyElement`. We try not to abbreviate too much.

web/theme.dart
```diff
@@ -0,0 +1,63 @@
...
+
+  theme.addEventListener('change', (event) {
+    if (theme.checked == true) {
+      if (theme.getAttribute('value') == 'light-theme') {
+        elementB?.classes.remove('light-theme');
+        elementB?.classes.toggle('dark-theme');
+        theme.setAttribute('value', 'dark-theme');
+      } else {
+        elementB?.classes.remove('light-theme');
```
srawlins, line 15: Something is strange here. Lines 15, 16, 17 are the same as 11, 12, 13?

web/theme.dart
```diff
@@ -0,0 +1,63 @@
...
+        elementB?.classes.toggle('dark-theme');
+        theme.setAttribute('value', 'dark-theme');
+      }
+      window.localStorage['checked'] = 'true';
```
srawlins, line 19: Should the name of this variable refer to 'theme' in some way? It seems weird to just have a variable in the `localStorage` map called "checked."

web/theme.dart
```diff
@@ -0,0 +1,63 @@
...
+      elementB?.classes.toggle('light-theme');
+      theme.setAttribute('value', 'light-theme');
+      window.localStorage['checked'] = 'false';
+      theme.checked = false;
+    }
+  }
+
+  void setChecked() {
+    if (window.localStorage['checked'] != null) {
+      if (window.localStorage['checked'] == 'true') {
```
srawlins, line 52: It looks like all of this can be simplified:

* `toggleTheme()` does not need to be in each then/else body.
* Instead of an if/else, the variable could be assigned directly according to `window.localStorage['checked']` with `theme.checked = window.localStorage['checked'] == 'true'`.

srawlins (changes requested):
Very cool!

lib/src/generator/templates.aot_renderers_for_html.dart
klr981: Thank you for the feedback! I went ahead and searched for Material icons and found that it does not have a moon icon atleast what I saw in the [Flutter API](https://api.flutter.dev/flutter/material/Icons-class.html), just the sun icons and both icons would be needed.

parlough: They are a bit tricky to find, but there are a few options called "Dark Mode" (which corresponds to "Light Mode") and another option called "Nightlight".

https://fonts.google.com has a better search which supports synonyms like "Moon": https://fonts.google.com/icons?icon.set=Material+Icons

klr981: Thank you! working on changing it currently

klr981: Changed the icons to the ones I found in fonts.google, thank you for the feedback!

darkmode-new-icons

lightmode-new-icons

web/theme.dart
```diff
@@ -0,0 +1,49 @@
...
+    }
+    else {
+      bodyElement.setAttribute('class','light-theme');
+      theme.setAttribute('value', 'light-theme');
+      window.localStorage['colorTheme'] = 'false';
+      theme.checked = false;
+    }
+  }
+
+  void setChecked() {
```
srawlins, line 41: I don't think this needs to be a local function if it is just called once, just below.

web/theme.dart
```diff
@@ -0,0 +1,49 @@
...
+
+  var bodyElement = document.body;
+
+  if( bodyElement == null ){
+    return;
+  }
+
+  var theme = document.getElementById('theme') as InputElement;
+
+  theme.addEventListener('change', (event) {
```
srawlins, line 13: What does `theme.addEventListener('change'` do? This callback is called when?

srawlins (changes requested):

web/theme.dart
klr981: It's called every time the user changes the theme using the toggle input.

srawlins: Haha brain fart. I have not written web in a _long_ time. Thanks for bearing with me!

srawlins (changes requested):

parlough:
Superseded by https://github.com/dart-lang/dartdoc/pull/3113

Issue #3037, opened 2022-05-16
Improve search experience


keertip:
- Based on a static JSON doc. Just autocomplete in a drop-down. Limited results. Limited items shown

Suggestions
- Copy Rust (static file based)
See https://docs.rs/wasmer/2.0.0/wasmer/
Also static JSON-file based
Results are shown full screen with a lot more context
- Elasticsearch (or similar)?
Wire up a text indexer to provide deep results across file contents
CONS: wire up a service, pay licensing, will cost (a lot?) more
- [Morsels](https://github.com/ang-zeyu/morsels)
Looks similar to what docs.rs uses. Generate server side. Run client side (with WASM).

- https://typesense.org/ - free simple version + hosted option
- [Algolia](https://www.algolia.com/)
Is it tacky to use their free version? Hrm…

- [Firebase Flutter docs](https://firebase.flutter.dev/docs/) – great experience!

keertip:
other search related feedback:

- [prioritize dart:core](https://github.com/dart-lang/dartdoc/issues/954)
- [indicate more results availabe](https://github.com/dart-lang/dartdoc/issues/955),
- [allow full text search](https://github.com/dart-lang/dartdoc/issues/1703)

PR #3111, opened 2022-08-09
Dartdoc Search

klr981:
### Description
- style.css: changes include a new look for the search box, how the search results are seen in the dropdown list
- generator_utils.dart: includes a one line description that is used in the new search results
- search_results_page.html and search_results_page.md: new templates for the search page results where the results are displayed dynamically.
- templates.dart, validator.dart, html_generator.dart and markdown_generator.dart: changes done are to include the new search page when building docs
- search.dart: 
&nbsp; - categories: This function takes the search suggestion in HTML form that is going to be displayed and the library it belongs to. It takes the two inputs and inserts it in a map type Map<String, Element>, if the Map does not contain the key it adds simply adds the library as key and the value of the first suggestion.
&nbsp; - createCategory: This function takes input, in this case the library name and makes the necessary calls to create a HTML element.
&nbsp; - iterateCategoriesMap: This function iterates through the categoriesMap to added it to given HTML element. It adds the value in the map to the given html element. 
&nbsp; - searchPageResults: This function is used when the press key enter is pressed and is directed to the search_results_page. It adds some html and uses the iterateCategoriesMap to display all the research found. 
&nbsp; -key enter code: Now if the key enter is pressed and there is no element selected it navigates to the search results page. Since not all the docs pages have base href it looks for a specific value in the HTML to navigate to the right path.


kevmoo:
Consider adding an entry to `CHANGELOG.md`, too!

lib/resources/styles.css
```diff
@@ -124,13 +124,13 @@ nav.navbar {
   border: 0;
 }
 
-@media (max-width: 768px) {
```
srawlins, line 127: Can you comment on this change (in the PR, I mean; not in the code)?

lib/resources/styles.css
```diff
@@ -975,9 +1008,86 @@ button {
...
+  .temp-box{
+    width: 0px;
+  }
+
+  #sidenav-left-toggle{
+   margin-right:10px;
+   margin-left:20px;
+  }
+
+  .self-name{
```
srawlins, line 1074: nit: space before `{`

lib/resources/styles.css
```diff
@@ -975,9 +1008,86 @@ button {
...
+    width: 100px;
+  }
+}
+
+@media screen and (max-width: 320px) {
+  .temp-box{
+    width: 0px;
+  }
+
+  #sidenav-left-toggle{
```
srawlins, line 1069: nit: space before `{`

lib/resources/styles.css
```diff
@@ -975,9 +1008,86 @@ button {
   margin: 0;
 }
 
+.tt-category-title {
+  font-size: 14px;
+  margin-right:10px;
```
srawlins, line 1013: nit: space after `:`

lib/resources/styles.css
```diff
@@ -975,9 +1008,86 @@ button {
...
+
+.tt-suggestion-title {
+  font-size: 16px;
+  margin-right:10px;
+  margin-left: 10px;
+  margin-bottom: 0px;
+  margin-top:4px;
+}
+
+.temp-box{
```
srawlins, line 1027: Nit: space before `{` here and in a few spaces below.

lib/src/generator/generator_utils.dart
```diff
@@ -34,6 +34,16 @@ String generateCategoryJson(Iterable<Categorization> categories, bool pretty) {
   return encoder.convert(indexItems.toList());
 }
 
+String regEx(String? input){
+  if(input==null){
```
srawlins, line 38: It looks like this file is not formatted. Make sure to format with something like `dart format lib/`

lib/src/generator/generator_utils.dart
```diff
@@ -34,6 +34,16 @@ String generateCategoryJson(Iterable<Categorization> categories, bool pretty) {
   return encoder.convert(indexItems.toList());
 }
 
+String regEx(String? input){
```
srawlins, line 37: Hmm I'm not sure about the names of different things in here. Can you look into naming the function differently, and the variables inside, and maybe add a doc comment above?

https://dart.dev/guides/language/effective-dart/documentation#do-use--doc-comments-to-document-members-and-types

lib/src/generator/html_generator.dart
```diff
@@ -40,6 +40,8 @@ class HtmlGeneratorBackend extends GeneratorBackendBase {
     var data = PackageTemplateData(options, graph, package);
     var content = templates.renderError(data);
     write(writer, '__404error.html', data, content);
+    var searchContent = templates.renderSearchPage(data);
+    write(writer, 'search_results_page.html', data, searchContent);
```
srawlins, line 44: The word 'page' is rather unnecessary here. Let's name this file something shorter like `search_results` or even just `search`? Maybe we want to reserve `search.html` for something else but I'm inclined to be as concise as possible, and if we add some big fancy search page later, then that and this search results page can both just be `search.html`.

web/search.dart
```diff
@@ -115,6 +118,9 @@ List<IndexItem> findMatches(List<IndexItem> index, String query) {
     } else if (query.length > 1) {
       if (name.startsWith(query) || qualifiedName.startsWith(query)) {
         score(750);
+      // } else if (oneLineDescription.runtimeType!=Null && oneLineDescription!='' && oneLineDescription?.contains(query)==true){
```
srawlins, line 121: Delete?

web/search.dart
```diff
@@ -203,24 +219,55 @@ void initializeSearch(
         query, "<strong class='tt-highlight'>$sanitizedText</strong>");
   }
 
+  Map<String,Element> categoriesMap = new LinkedHashMap();
```
srawlins, line 222: We prefer to use collection literals. Are you picking LinkedHashMap for a specific reason?

https://dart.dev/guides/language/effective-dart/usage#do-use-collection-literals-when-possible

srawlins (changes requested):
Leaving preliminary comments; very cool, and really excited for this!

lib/src/generator/generator_utils.dart
klr981: Addressed in new commit

web/search.dart
klr981: Addressed in new commit

lib/resources/styles.css
```diff
@@ -9,7 +9,7 @@
 .secondary-text-color  { color: #727272; }
 .divider-color         { border-color: #B6B6B6; }
 
-.light-theme{
```
kevmoo, line 12: Points for eliminating whitespace-only changes. Makes it hard to review!

lib/src/generator/generator_utils.dart
```diff
@@ -34,6 +34,16 @@ String generateCategoryJson(Iterable<Categorization> categories, bool pretty) {
   return encoder.convert(indexItems.toList());
 }
 
+String removeHTMLTags(String? input) {
```
srawlins, line 37: Nit: spell HTML like "Html" in a variable name like this. https://dart.dev/guides/language/effective-dart/style#do-capitalize-acronyms-and-abbreviations-longer-than-two-letters-like-words

lib/src/generator/generator_utils.dart
```diff
@@ -34,6 +34,16 @@ String generateCategoryJson(Iterable<Categorization> categories, bool pretty) {
...
+  if (input == null) {
+    return '';
+  }
+  RegExp exp = RegExp(r"<[^>]*>", multiLine: true, caseSensitive: true);
```
srawlins, line 41: We can give this regular expression a name that denotes what it matches, something like "tag" or "htmlTag".

lib/src/generator/generator_utils.dart
```diff
@@ -34,6 +34,16 @@ String generateCategoryJson(Iterable<Categorization> categories, bool pretty) {
...
+  String beforeParsing = input.toString();
+  String parsedString = beforeParsing.replaceAll(exp, '');
```
srawlins, line 43: I don't think `input.toString()` is doing anything necessary, is it? We can just return `input.replaceAll(exp, '');`.

lib/src/validator.dart
```diff
@@ -124,6 +124,9 @@ class Validator {
           if (const {'__404error.html', 'categories.json'}.contains(fullPath)) {
             _warn(PackageWarning.orphanedFile, fullPath, _origin);
           }
+          if (const {'search.html', 'categories.json'}.contains(fullPath)) {
```
srawlins, line 127: We can just fold `'search.html'` into the const set on line 124, since this is the same check.

web/search.dart
```diff
@@ -96,6 +96,7 @@ List<IndexItem> findMatches(List<IndexItem> index, String query) {
   for (var element in index) {
     void score(int value) {
       value -= (element.overriddenDepth ?? 0) * 10;
+      print(element.type);
```
srawlins, line 99: Remove?

web/search.dart
```diff
@@ -139,16 +140,19 @@ List<IndexItem> findMatches(List<IndexItem> index, String query) {
 }
 
 const minLength = 1;
-const suggestionLimit = 10;
+var suggestionLimit = 10;
+var allResults = 0;
```
srawlins, line 144: The name here makes it sound like this is a list or something. I think we prefer a noun like "length" or "count" in the name for an int like this.

web/search.dart
```diff
@@ -189,10 +193,16 @@ void initializeSearch(
   listBox.style.display = 'none';
   listBox.classes.add('tt-menu');
 
-  var presentation = document.createElement('div');
-  presentation.classes.add('tt-elements');
+  // Element use in listbox to inform the functionality of hitting enter in search box
+  var moreResults = document.createElement('div');
+  moreResults.classes.add('enter-search-message');
+  //moreResults.innerHtml = 'Press "Enter" key to see more results if available';
```
srawlins, line 199: Remove?

web/search.dart
```diff
@@ -189,10 +193,16 @@ void initializeSearch(
   listBox.style.display = 'none';
   listBox.classes.add('tt-menu');
 
-  var presentation = document.createElement('div');
-  presentation.classes.add('tt-elements');
+  // Element use in listbox to inform the functionality of hitting enter in search box
```
srawlins, line 196: Nit: make sure to end comments with a period.

https://dart.dev/guides/language/effective-dart/documentation#do-format-comments-like-sentences

web/search.dart
```diff
@@ -203,24 +213,54 @@ void initializeSearch(
         query, "<strong class='tt-highlight'>$sanitizedText</strong>");
   }
 
+  var categoriesMap = <String, Element>{};
+
+  // Function that takes the name of the library/class the search suggestion belongs to and puts in the map
+  void categories(Element suggestionLibrary, Element suggestion) {
```
srawlins, line 219: I think this is kind of a strange function name. Do you mean `categorize`?

web/search.dart
```diff
@@ -203,24 +213,54 @@ void initializeSearch(
...
+    var input = suggestionLibrary.innerHtml;
```
srawlins, line 220: Since we assert that `input` is non-null in both branches of the if-statement below (`input!`), would it make more sense to null-assert here? We can also do something like `if (input == null) return;`.

web/search.dart
```diff
@@ -298,7 +393,9 @@ void initializeSearch(
     }
 
     var suggestions = findMatches(index, newValue);
+    allResults = suggestions.length;
     if (suggestions.length > suggestionLimit) {
+      // moreResults.innerHtml = 'PressK "Enter" key to see all ${suggestions.length} results';
```
srawlins, line 398: Remove?

srawlins (changes requested):
Looking good!

keertip:
For #3037

lib/src/generator/generator_utils.dart
klr981: Addressed in new commit

lib/src/generator/generator_utils.dart
klr981: Addressed in new commit

lib/src/validator.dart
klr981: Addressed in new commit

web/search.dart
```diff
@@ -251,21 +298,70 @@ void initializeSearch(
...
-    if (presentation.hasChildNodes()) {
+    if (searchResults.hasChildNodes()) {
       listBox.style.display = 'block';
       listBox.setAttribute('aria-expanded', 'true');
     }
   }
 
+  // Function that iterates through the map and appends it to the given HTML element.
+  void iterateCategoriesMap(Element e) {
+    for (var k in categoriesMap.keys) {
```
srawlins, line 309: I think this might be easier with Map.forEach:

```dart
categoriesMap.forEach((k, values) {
...
});

File: web/search.dart [hunk 24]
```diff
@@ -251,21 +298,70 @@ void initializeSearch(
...
+        var values = categoriesMap[k];
+        e.append(values!);
+      }
+    }
+  }
+
+  // Function that creates the content displayed in the main-content element.
+  void searchResultPage(String input) {
+    var mainContent = document.getElementById('dartdoc-main-content');
+    mainContent?.text = '';
```
srawlins, line 320: Should we just `return;` if mainContent is null? Save all the work below and simplify the null checks.

web/search.dart
```diff
@@ -327,23 +425,45 @@ void initializeSearch(
...
-        window.location.assign('$htmlBase$href');
+      if (selectedElement != null) {
+        var selectingElement = selectedElement ?? 0;
+        var href = suggestionElements[selectingElement].dataset['href'];
+        if (href != null) {
+          window.location.assign('$htmlBase$href');
+        }
+        return;
+      }
+      // If there no search suggestion selected then change the window location to the search.html.
```
srawlins, line 443: Grammar nit: "If there are no search"

web/search.dart
```diff
@@ -327,23 +425,45 @@ void initializeSearch(
...
+      if (selectedElement == null ||
+          listBox.getAttribute('aria-expanded') == 'true' ||
+          suggestionElements.isEmpty) {
+        // Saves the input in the search to be used for creating the query parameter.
+        var input = htmlEscape.convert(actualValue);
+        var relativePath = '';
+        if (document
+                    .querySelector('body')
```
srawlins, line 451: Let's save `document.querySelector('body')` as a local variable. This should make the code below shorter, and avoid multiple null-checks and null-asserts.

srawlins (changes requested):

web/search.dart
```diff
@@ -251,21 +298,70 @@ void initializeSearch(
...
 
   void showSuggestions() {
-    if (presentation.hasChildNodes()) {
+    if (searchResults.hasChildNodes()) {
       listBox.style.display = 'block';
       listBox.setAttribute('aria-expanded', 'true');
     }
   }
 
+  // Function that iterates through the map and appends it to the given HTML element.
```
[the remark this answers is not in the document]
klr981: I made a new commit, let me know if that is better.

web/search.dart
```diff
@@ -203,24 +211,59 @@ void initializeSearch(
         query, "<strong class='tt-highlight'>$sanitizedText</strong>");
   }
 
+  var categoriesMap = <String, Element>{};
+
+  // Function that takes the name of the library/class the search suggestion belongs to and puts in the map.
```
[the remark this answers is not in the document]
klr981: Made a new commit that addresses this, let me know if that is better.

web/search.dart
```diff
@@ -203,24 +211,56 @@ void initializeSearch(
...
+    var input = suggestionLibrary.innerHtml;
+
+    if (input == null) {
+      return;
+    }
+
+    var element = categoriesMap[input];
+    if (element != null) {
+      element.append(suggestion);
+      categoriesMap.update(input, (value) => element);
```
srawlins, line 227: Since you're not using `value` here, I think it's easiest to just write `categoriesMap[input] = element;` like you do on line 230.

web/search.dart
```diff
@@ -160,6 +162,7 @@ void initializeSearch(
     }
   });
 
+  var body = document.querySelector('body');
```
srawlins, line 165: Let's declare this variable as close as possible to where it is used, around line 450.

We can also null-assert (`!`) when we declare it since it is null-asserted on line 457.

web/search.dart
```diff
@@ -203,24 +211,56 @@ void initializeSearch(
...
-      fromLib.classes.add('search-from-lib');
-      fromLib.innerHtml = 'from ${highlight(match.enclosedBy!.name, query)}';
-
-      suggestion.append(fromLib);
+    // The new one line description to use in the search suggestions.
+    if (match.desc != '') {
+      var inputDesc = document.createElement('div');
+      inputDesc.classes.add('one-line-description');
+      var innerHtml = '';
+      if (match.desc != null) {
```
srawlins, line 259: Some of this is a little weird in the latest commit. We check on line 255 that `match.desc != ''`, then check here that it's not `null`. It seems like it should just be one check?

web/search.dart
```diff
@@ -233,7 +273,12 @@ void initializeSearch(
         event.preventDefault();
       }
     });
-
+    var suggestionEnclose = match.enclosedBy;
+    if (suggestionEnclose != null) {
+      var category = '${suggestionEnclose.name} ${suggestionEnclose.type}';
+      var toMap = createCategory(category, suggestionEnclose.href);
```
srawlins, line 279: Hmm, I don't understand this variable name. But to simplify things, you could just inline the `createCategory` call, haha:

```dart
categorize(createCategory(...`
```

web/search.dart
```diff
@@ -251,21 +296,75 @@ void initializeSearch(
...
       listBox.style.display = 'block';
       listBox.setAttribute('aria-expanded', 'true');
     }
   }
 
+  // Function that iterates through the categoriesMap and appends it to the given HTML element
+  // The received Element is the one used for search results page or dropdown list in the search bar
+  void iterateCategoriesMap(Element e) {
+    categoriesMap.forEach((keys, values) {
+      if (categoriesMap[keys] != null) {
```
srawlins, line 309: You don't have to do this check since you are holding onto the `values` value from line 308. I think just `e.append(values);`.

web/search.dart
```diff
@@ -251,21 +296,75 @@ void initializeSearch(
...
+
+    var section = document.createElement('section');
+    section.classes.add('search-summary');
+    mainContent.append(section);
+
+    var title = document.createElement('h2');
+    title.innerHtml = 'Search Results';
+    mainContent.append(title);
+
+    var summary = document.createElement('div');
```
srawlins, line 333: Optional, since we don't have a lot of code like this, but internally it is very idiomatic to use [cascades](https://dart.dev/guides/language/language-tour#cascade-notation) in these HTML nodes:

```dart
var summary = document.createElement('div')
    ..classes.add('search-summary')
    ..innerHtml = '$suggestionLength results for "$input"';
```

Might shorten the code or make it look more readable.

web/search.dart
```diff
@@ -251,21 +296,75 @@ void initializeSearch(
...
+    }
+  }
+
   void hideSuggestions() {
     listBox.style.display = 'none';
     listBox.setAttribute('aria-expanded', 'false');
   }
 
+  void enterMessage() {
+    if (suggestionLength > 10) {
```
srawlins, line 355: We can save some lines here and simplify the logic with a conditional expression:

```dart
moreResults.innerHtml = suggestionLength > 10
  ? 'Press "Enter" key to see all $suggestionLength results'
  : '';
```

web/search.dart
```diff
@@ -327,23 +428,40 @@ void initializeSearch(
...
-        window.location.assign('$htmlBase$href');
+      if (selectedElement != null) {
+        var selectingElement = selectedElement ?? 0;
+        var href = suggestionElements[selectingElement].dataset['href'];
+        if (href != null) {
+          window.location.assign('$htmlBase$href');
+        }
+        return;
+      }
+      // If there are no search suggestion selected then change the window location to the search.html.
```
srawlins, line 446: Grammar nit: "no search suggestions selected" I think.

web/search.dart
```diff
@@ -327,23 +428,40 @@ void initializeSearch(
...
+        if (body?.getAttribute('data-using-base-href') == 'true' &&
+            body?.getAttribute('data-base-href') == '') {
+          relativePath = document.querySelector('base')!.getAttribute('href')!;
+        } else {
+          relativePath = body!.getAttribute('data-base-href')!;
+        }
+        var href = Uri.parse(window.location.href);
+        var base = href.resolve(relativePath);
+        var search = Uri.parse('${base}search.html');
+        search = search.replace(queryParameters: {'query': input});
```
srawlins, line 462: This can be done inline above: (`Uri.parse(...).replace(...);`

web/search.dart
```diff
@@ -427,6 +544,17 @@ void initializeSearch(
 
     event.preventDefault();
   });
+
+  // Verifying the href to check if the search html was called to generate the main content elements that are going to be displayed.
+  if (window.location.href.contains('search.html')) {
+    var input = uri.queryParameters['query'];
+    input = htmlEscape.convert(input!);
```
srawlins, line 551: If the user just typed this page into the browser, then I think `input!` here will crash in JavaScript, crashing the whole web app. Better to just return early if `input` is null.

srawlins (changes requested):
Few more changes. Looking great!

web/search.dart
klr981: You are right, it should just be checked once. New commit addresses this.

web/search.dart
klr981: I liked the format and went ahead and changed it to your suggestion. Thank you

web/search.dart
```diff
@@ -203,24 +210,54 @@ void initializeSearch(
...
+    var input = suggestionLibrary.innerHtml;
+
+    if (input == null) {
+      return;
+    }
+
+    var element = categoriesMap[input];
+    if (element != null) {
+      element.append(suggestion);
+      categoriesMap[input] = element;
```
srawlins, line 226: This line should not be needed. When you extract `categoriesMap[input]` to a local variable, `element`, it does not make a copy of the object. So `element.append` above will apply to the object in the Map.

web/search.dart
```diff
@@ -251,21 +292,71 @@ void initializeSearch(
...
+      mainContent.append(noResults);
+    }
+  }
+
   void hideSuggestions() {
     listBox.style.display = 'none';
     listBox.setAttribute('aria-expanded', 'false');
   }
 
+  void enterMessage() {
```
srawlins, line 349: Since this changes the page, I think an action-y verb in the name would be good, like "showEnterMessage" or something similar.

web/search.dart
```diff
@@ -251,21 +292,71 @@ void initializeSearch(
...
+    moreResults.innerHtml = suggestionLength > 10
```
srawlins, line 350: Since there is no HTML here, I think it would be safer and maybe simpler to instead set `text =`. I can't remember if that's the right call.

web/search.dart
```diff
@@ -327,23 +420,41 @@ void initializeSearch(
...
-      return;
-    }
-
     if (event.type != 'keydown') {
       return;
     }
 
     event = event as KeyboardEvent;
 
+    var body = document.querySelector('body');
```
srawlins, line 429: Putting a null-assert at the end here (`document.querySelector('body')!;`) will simplify the null-checking code below.

web/search.dart
```diff
@@ -251,21 +292,71 @@ void initializeSearch(
...
+  // The received Element is the one used for search results page or dropdown list in the search bar
+  void iterateCategoriesMap(Element e) {
+    categoriesMap.forEach((keys, values) {
+      var value = categoriesMap[keys];
+      e.append(value!);
+    });
+  }
+
+  // Function that creates the content displayed in the main-content element.
+  void searchResultPage(String input) {
```
srawlins, line 311: I think this function name would read better with a verb in front, like 'showSearchResultPage'. Although I can see the appeal of a more declarative API, with functions named as the nouns they produce... I just don't think its a convention we have.

srawlins (changes requested):
Maybe last round!

web/search.dart
klr981: That is a good suggestion, thank you

web/search.dart
```diff
@@ -203,24 +210,53 @@ void initializeSearch(
...
+      element.append(suggestion);
+    } else {
+      suggestionLibrary.append(suggestion);
+      categoriesMap[input] = suggestionLibrary;
+    }
+  }
 
-      suggestion.append(fromLib);
+  // Using the name of the library/class creates the Element for it.
+  Element createCategory(String lib, String href) {
```
srawlins, line 233: Instead of "Category", which is a different Dartdoc concept, could we use a similar word, like Group or Grouping or something similar.

srawlins, line 233: Since its not always a library, could we use a different word from 'lib'. 'encloser' is an idea, but what a terrible word... Not too important.

lib/src/generator/html_generator.dart
```diff
@@ -40,6 +40,8 @@ class HtmlGeneratorBackend extends GeneratorBackendBase {
     var data = PackageTemplateData(options, graph, package);
```
srawlins, line 40: If we make `navLinks` a mutable field on PackageTemplateData, then we can edit it here to be `[graph.defaultPackage]` instead of being empty.''

If there are static analysis problems with turning that into a field, you could (😞 ) subclass PackageTemplateData with something like `PackageTemplateDataForSearch extends PackageTemplateData {}` in template_data.dart, and implement `get navLinks`.

srawlins (changes requested):
Could we add some screenshots (I wish we had a web testing framework for dartdoc... 😢 ):

* search_page.html
* search_page.html?query=foo (with results)
* search_page.html?query=foo with "sparse" results, like 1-2 results per "grouping" (maybe a common field/method on many classes)
* existing search autocomplete in the top right
* shrunken down page (for mobile, < 800 px wide) of search_page.html?query=foo and a normal page using the search in the mobile sidebar. Chrome DevTools has a cool mobile button thing in the top left.

klr981:
@srawlins the screenshots for search:

### search.html
search html

### Search list with more than 10 results for "string"
Search list with more than 10 results

### Search page results for "string"
Search page results for string

### Search list with less than 10 results for library
Search list with less than 10 results

### Search results for groupings for "has"
Search results for with groupings

### Search page results with groupings for "has"
Search page results with groupings

### No results page
No results page

### Search with less than 800 screen size
Search page with less than 800 screen size

### Search list with less than 800 screen size
Search results page with less than 800 screen size

### Search results page with less than 800 screen size
Search results page with less than 800 screen size

srawlins:
🎉 Thank you for thinking of good screenshots that I totally missed (like "no results")!

srawlins (approved):
I think this looks great; @keertip anything else?

keertip:
No changes! Screenshots look great! LGTM!

keertip (approved):

PR #3112, opened 2022-08-09
Move to analyzer 4.4.0

srawlins:
I tried to make this a no-op. It's as close as it can get 😛

### DIFF
--- a/lib/src/element_type.dart
+++ b/lib/src/element_type.dart
@@ -12,6 +12,7 @@ import 'package:dartdoc/src/model/comment_referable.dart';
 import 'package:dartdoc/src/model/model.dart';
 import 'package:dartdoc/src/model/model_object_builder.dart';
 import 'package:dartdoc/src/render/element_type_renderer.dart';
+import 'package:dartdoc/src/type_utils.dart';
 import 'package:meta/meta.dart';
 
 mixin ElementTypeBuilderImpl implements ElementTypeBuilder {
@@ -36,7 +37,7 @@ abstract class ElementType extends Privacy
 
   factory ElementType._from(
       DartType f, Library library, PackageGraph packageGraph) {
-    var fElement = f.element;
+    var fElement = DartTypeExtension(f).element;
     if (fElement == null ||
         fElement.kind == ElementKind.DYNAMIC ||
         fElement.kind == ElementKind.NEVER) {
@@ -84,6 +85,9 @@ abstract class ElementType extends Privacy
         f as ParameterizedType, library, packageGraph, element);
   }
 
+  /// The element of [type].
+  TypeDefiningElement? get typeElement => DartTypeExtension(type).element;
+
   bool get canHaveParameters => false;
 
   bool get isTypedef => false;
@@ -129,9 +133,9 @@ class UndefinedElementType extends ElementType {
   @override
   String get name {
     if (type.isVoid) return 'void';
-    assert((const {'Never', 'void', 'dynamic'}).contains(type.element!.name),
+    assert((const {'Never', 'void', 'dynamic'}).contains(typeElement!.name),
         'Unrecognized type for UndefinedElementType: ${type.toString()}');
-    return type.element!.name!;
+    return typeElement!.name!;
   }
 
   @override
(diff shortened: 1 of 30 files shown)


lib/src/element_type.dart
```diff
@@ -296,13 +300,12 @@ abstract class DefinedElementType extends ElementType {
...
-        !(_bound as InterfaceType)
-            .typeArguments
-            .every((t) => t is InterfaceType)) {
+    final bound = _bound;
+    if (bound is InterfaceType &&
+        !bound.typeArguments.every((t) => t is InterfaceType)) {
       var typeSystem = library.element.typeSystem;
       return typeSystem.instantiateToBounds2(
-          classElement: _bound.element as ClassElement,
+          classElement: bound.element2 as ClassElement,
```
scheglov, line 308: Not every `InterfaceType.element2` is `ClassElement`.
Ah, but `instantiateToBounds2` requires `ClassElement`.
I will need to change it to `InterfaceElement`.

lib/src/generator/template_data.dart
```diff
@@ -384,7 +385,8 @@ class MethodTemplateData extends TemplateData<Method>
 
   MethodTemplateData(super.htmlOptions, super.packageGraph, this.library,
       this.container, this.method, this._sidebarForContainer)
-      : _containerDescription = container.isClass ? 'class' : 'extension';
+      : _containerDescription =
+            container is ClassElement ? 'class' : 'extension';
```
scheglov, line 389: What if we get `MixinElement` here?
Should we check for `InterfaceElement` instead?

lib/src/type_utils.dart
```diff
@@ -0,0 +1,27 @@
...
+extension DartTypeExtension on DartType {
+  /// Returns the 'element' for this type, which is sometimes `null`.
+  TypeDefiningElement? get element {
+    final self = this;
+    if (self is InterfaceType) {
+      return self.element2;
+    } else if (self is DynamicType) {
+      // This may be made available in analyzer.
+      // ignore: deprecated_member_use
+      return self.element as TypeDefiningElement;
```
scheglov, line 17: You cannot do this.
I will remove `DartType.element`.
Deprecation is to let you know that it will be gone, not to be ignored.

tool/mustachio/codegen_aot_compiler.dart
```diff
@@ -207,7 +208,8 @@ Future<Method> _redirectingMethod(
 
   var parameters = <Parameter>[];
   for (var context in compiler._usedContextStack) {
-    var contextElement = context.type.element;
+    var contextElement =
+        DartTypeExtension(context.type).element as ClassElement;
```
scheglov, line 212: Any cast is a potential land mine.
Are you sure that it is always `ClassElement`, and not `EnumElement` or `MixinElement`?

lib/src/element_type.dart
srawlins: Sounds good!

lib/src/generator/template_data.dart
srawlins: lol technical debt. Fixed to InterfaceElement and added a TODO.

lib/src/type_utils.dart
srawlins: Should I return `(self as DynamicTypeImpl).element`?

tool/mustachio/codegen_aot_compiler.dart
srawlins: Fixed

lib/src/type_utils.dart
scheglov: This will still fail at runtime.
Do you need this element?

srawlins: Maybe not... it was maybe only needed in one spot, and I refactored. Uploaded a new change.

scheglov (approved):

srawlins:
CI bots are now failing with a single issue:

```none
stdout: analyze: Analyzing ....
analyze:
analyze:    info - lib/src/element_type.dart:308:25 - 'instantiateToBounds2' is deprecated and shouldn't be used. Use instantiateInterface/TypeAliasToBounds() instead. Try replacing the use of the deprecated member with the replacement. - deprecated_member_use
analyze:
analyze: 1 issue found.
```

which I think is from the just-released analyzer 4.5.0. So I will land this with red bots, then move on to analyzer 4.5.0.

PR #3113, opened 2022-08-09
Dark Mode: Landed changes in this PR

klr981:
Opened new PR due to Git issues on previous PR.

[See Previous PR](https://github.com/dart-lang/dartdoc/pull/3107)

Closes https://github.com/dart-lang/dartdoc/issues/1885


web/theme.dart
```diff
@@ -0,0 +1,40 @@
...
+      theme.setAttribute('value', 'dark-theme');
+      window.localStorage['colorTheme'] = 'true';
+    } else {
+      bodyElement.setAttribute('class', 'light-theme');
+      theme.setAttribute('value', 'light-theme');
+      window.localStorage['colorTheme'] = 'false';
+    }
+  });
+
+  void setChecked() {
```
srawlins, line 24: We should just inline this function body since it is only called once, immediately after it is declared.

srawlins (changes requested):

srawlins (approved):

web/theme.dart
klr981: Addressed in the most recent commit

keertip (approved):

keertip:
@klr981 , could you take a look at the test failures?

keertip:
> @klr981 , could you take a look at the test failures?

It turns out the failures are due to other changes and not this PR, so ignore the comment.

LGTM!

srawlins:
The only static failures are related to analyzer 4.5.0, and a lint rule that is not useful for us (`eol_at_end_of_file`).

I'll merge and work on analyzer 4.5.0 tomorrow.

PR #3115, opened 2022-08-10
Move to analyzer 4.5.0

srawlins:


lib/src/element_type.dart
```diff
@@ -305,8 +305,9 @@ abstract class DefinedElementType extends ElementType {
     if (bound is InterfaceType &&
         !bound.typeArguments.every((t) => t is InterfaceType)) {
       var typeSystem = library.element.typeSystem;
-      return typeSystem.instantiateToBounds2(
-          classElement: bound.element2 as ClassElement,
+      return typeSystem.instantiateInterfaceToBounds(
+          // TODO(srawlins): It sure seems like this will crash on MixinElement.
```
scheglov, line 309: Well, then why do you cast it?
`instantiateInterfaceToBounds()` needs `InterfaceElement`, not `ClassElement`.

srawlins: Done!

scheglov (approved):

PR #3119, opened 2022-08-12
Bump to analyzer 4.6.0

srawlins:


lib/src/type_utils.dart
```diff
@@ -12,9 +12,9 @@ extension DartTypeExtension on DartType {
     if (self is InterfaceType) {
       return self.element2;
     } else if (self is NeverType) {
-      return self.element as TypeDefiningElement;
+      return self.element2 as TypeDefiningElement;
     } else if (self is TypeParameterType) {
-      return self.element;
+      return self.element2;
     } else {
       // Remaining cases like `DynamicType`, `FunctionType`, and `VoidType`.
```
scheglov, line 19: BTW, we had an issue with `dynamic`, with this kind of rollback, you still can use `DynamicType.element2`.

scheglov (approved):

lib/src/type_utils.dart
srawlins: Ah ok, it looks like it is not needed. But I'll keep it in mind.

PR #3128, opened 2022-08-24
Show if the method is static in method page

### LINKED ISSUE #3108: Method documentation does not show whether the method is static or not
Consider https://api.dart.dev/stable/2.17.6/dart-core/Object/hashAll.html

There is no mentions in the header ("hashAll method") or in the type signature that the method is static.

I need to look at the implementation, or see if the method name is in "STATIC METHODS" menu on the left.

It would be easier if the header said "hashAll static method".

osa1:
Fixes #3108

### DIFF
--- a/lib/src/model/method.dart
+++ b/lib/src/model/method.dart
@@ -62,7 +62,9 @@ class Method extends ModelElement
       '${enclosingElement.library.dirName}/${enclosingElement.name}/$fileName';
 
   String get fullkind {
+    // A method cannot be abstract and static at the same time.
     if (element.isAbstract) return 'abstract $kind';
+    if (element.isStatic) return 'static $kind';
     return kind;
   }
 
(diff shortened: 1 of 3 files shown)


lib/src/model/method.dart
```diff
@@ -62,7 +62,9 @@ class Method extends ModelElement
       '${enclosingElement.library.dirName}/${enclosingElement.name}/$fileName';
 
   String get fullkind {
+    // A method cannot be abstract and static at the same time
```
srawlins, line 65: Nit: add a period at the end of a comment. https://dart.dev/guides/language/effective-dart/documentation#do-format-comments-like-sentences

srawlins (changes requested):
Thanks!

srawlins (approved):
Big win, thanks!

PR #3129, opened 2022-08-24
Script added for full text search alternative solution

klr981:
This change is to include an alternative solution for full text search in api.flutter.dev and api.dart.dev. 

### Message to indicate that user input can be found using the alternative search
Screenshot 2022-08-24 10 04 52 AM

### Search page results using the script
Screenshot 2022-08-24 10 05 02 AM

### Current issue: Dark mode is not available since the script overrides the style.css changes to change the colors.
Screenshot 2022-08-24 10 07 07 AM


keertip:
@klr981 , instead of running the script, we could show a link to the main page itself when there are no search results. Then if the user chooses to click on the link it will not show up as a search panel within the api docs, thus eliminating this issue with dark mode.

It would be great to get the search script fixed, but since we are short of time lets try this approach.

klr981:
That sounds like a good idea @keertip, addressed in the new commit. 


[new approach.webm](https://user-images.githubusercontent.com/106621169/186528992-c0e47afc-853f-4043-adb6-f46a6c55e636.webm)

keertip (approved):

web/search.dart
```diff
@@ -334,7 +334,16 @@ void initializeSearch(
...
-            'There was not a match for "$input". Please try another search.';
+            'There was not a match for "$input". Want to try searching from additional Dart-related sites? ';
+
+      var buildLink = Uri.parse(
+              'https://dart.dev/search?cx=011220921317074318178%3A_yy-tmb5t_i&ie=UTF-8&hl=en&q=')
+          .replace(queryParameters: {'q': input});
+      var link = document.createElement('a')
+        ..setAttribute('href', buildLink.toString())
+        ..classes.add('seach-options')
+        ..innerHtml = 'click here to search in dart.dev.';
```
parlough, line 345: Maybe something more concise and focused on the action (we try [avoid "click here"](https://developers.google.com/style/link-text?hl=en#write-link-text)).:

```suggestion
        ..innerHtml = 'Search on dart.dev.';
```

parlough: Also, I'm not super familiar, but could we use [`text`](https://api.dart.dev/stable/dart-html/Node/text.html) here to avoid the browser unnecessarily parsing the specified string?

```suggestion
        ..text = 'click here to search in dart.dev.';
```

klr981: Addressed in new commit

web/search.dart
```diff
@@ -334,7 +334,16 @@ void initializeSearch(
       var noResults = document.createElement('div')
         ..classes.add('search-summary')
         ..innerHtml =
-            'There was not a match for "$input". Please try another search.';
+            'There was not a match for "$input". Want to try searching from additional Dart-related sites? ';
+
+      var buildLink = Uri.parse(
```
srawlins, line 339: Ooo very fancy

PR #3131, opened 2022-08-25
Refactor Documentation.asHtml and asOneLiner

srawlins:
This code was written in a way that sort of collided with null safety, or perhaps with lateness.

Multiple late fields can be initialized by one of two entrypoints, and one of them conditionally. A bit rough. But by setting some simple (if crude) private bool fields, we can make more explicit the coupling between the two fields.

This should be a no-op, except that many fields get to be non-nullable (🎉 ).

CC @Levi-Lesches


parlough:
I was also a little nervous to refactor this one, but this change looks great 🥳

keertip (approved):

PR #3135, opened 2022-08-30
Remove Material Icons and replace usages of it with Material Symbols

parlough:
As the dark theme implementation added Material Symbols which is an expanded/updated version of Material Icons, it likely doesn't make sense to keep Material Icons around anymore. This PR removes the old font and migrates dartdoc's only usage to use Material Symbols instead.

This could be **potentially breaking** for downstream users if they use Material icons elsewhere. They can re-add the font or migrate their usage to Material Symbols.

Migrated source code symbol (compare with [live dart:async docs](https://api.dart.dev/stable/2.18.0/dart-async/dart-async-library.html)):
dart:async page with new source code symbol

\cc @klr981

### DIFF
--- a/lib/resources/styles.css
+++ b/lib/resources/styles.css
@@ -81,7 +81,7 @@
   display: none;
 }
 
-.material-symbols-outlined {
+#theme-button .material-symbols-outlined {
   padding-top: 6px;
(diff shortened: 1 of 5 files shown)


srawlins (approved):
Thanks!

PR #3136, opened 2022-08-30
Make theme toggle not text selectable and have pointer on hover

parlough:
This prevents the theme toggle from being text selectable(which causes text selection to show up after clicking) as well as making the pointer cursor show up when hovering.

\cc @klr981


srawlins (approved):

srawlins:
Thanks!

PR #3139, opened 2022-08-31
Enable and fix unnecessary_parenthesis lint

kevmoo:
Also removed explicit lints that are already defined in pkg:lints


srawlins (approved):
🎉

PR #3151, opened 2022-09-10
Use `toList(growable: false)` more

srawlins:
I looked through every `toList()` call, for the ones which could be `growable: false`, and looked at Observatory's allocation profile, before and after, when documenting package:collection. Here are the result of 3 runs, excluding a 4th run outlier:

* Used MB: 444 (before) &xrarr; 411 (after) -- 7% reduction
* Allocated MB: 464 (before) &xrarr; 450 (after) -- 3% reduction
* _List instance MBs: 149 (before) &xrarr; 129 (after) -- 13% reduction
* _List instances: 868510 (before) &xrarr; 870317 (after) -- 0% reduction
* _GrowableList instance MBs: 15.1 (before) &xrarr; 13.3 (after) -- 12% reduction
* _GrowableList instances: 495092 (before) &xrarr; 435337 (after) -- 12% reduction

I can't super explain these. The reduction in _GrowableLists is expected. But why did the MBs of _List also shrink dramatically? 🤷

### DIFF
--- a/lib/src/model/package_graph.dart
+++ b/lib/src/model/package_graph.dart
@@ -486,7 +486,7 @@ class PackageGraph with CommentReferable, Nameable, ModelBuilder {
     return locatable.fullyQualifiedName.replaceFirst(':', '-');
   }
 
-  List<Package> get packages => packageMap.values.toList();
+  List<Package> get packages => packageMap.values.toList(growable: false);
 
   late final List<Package> publicPackages = () {
     assert(allLibrariesAdded);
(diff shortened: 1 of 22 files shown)


lib/src/model/package_graph.dart
```diff
@@ -827,11 +830,12 @@ class PackageGraph with CommentReferable, Nameable, ModelBuilder {
       // for an inherited element whose defining Class is not canonical.
       if (matches.length > 1 && preferredClass != null) {
         // Search for matches inside our superchain.
-        var superChain = preferredClass.superChain
-            .map((et) => et.modelElement)
-            .cast<InheritingContainer>()
-            .toList();
-        superChain.add(preferredClass);
+        var superChain = [
```
keertip, line 833: 👍

keertip (approved):
👍

mkustermann:
> I can't super explain these. The reduction in _GrowableLists is expected. But why did the MBs of _List also shrink dramatically? 🤷

Growable list is a wrapper object and a backing store. The backing store grows by power-of-2.  By switching to non-growable list, you avoid the wrapper object and shrink the backing store to the precise length you need it (modulo some VM-specifics, we round up to even number of slots).

srawlins:
Ah perfect, thanks so much for the explanation!

PR #3153, opened 2022-09-11
Adjust menu button color according to theme

### LINKED ISSUE #3144: Menu icon is wrong color in dark mode for narrow views
When using dark mode, and the viewport is narrow, we show a black menu icon. I think it should be white.
Screen Shot 2022-09-08 at 13 11 16

parlough:
Fixes #3144

light theme menu button
dark theme menu button


srawlins (approved):
Love it

PR #3154, opened 2022-09-11
Use const where possible on un-altered empty lists

srawlins:
I looked through every `[]` and `{}`, for the ones which could be `const`, and looked at Observatory's allocation profile, before and after, when documenting package:collection. Here are the result of 3 runs:

Used MB: 416 (before) ⟶ 395 (after) -- 5% reduction
Allocated MB: 441 (before) ⟶ 430 (after) -- 2% reduction
_List instance MBs: 123 (before) ⟶ 117 (after) -- 5% reduction
_List instances: 858428 (before) ⟶ 820412 (after) -- 4% reduction
_GrowableList instance MBs: 15.2 (before) ⟶ 14.6 (after) -- 4% reduction
_GrowableList instances: 497769 (before) ⟶ 479713 (after) -- 4% reduction
_InternalLinkedHashMap instance MBs: 6.7 (before) ⟶ 6.2 (after) -- 8% reduction
_InternalLinkedHashMap instances: 109505 (before) ⟶ 101055 (after) -- 8% reduction


mnordine:
What about when compiled to javascript?

srawlins:
You hopefully get an error that you cannot compile dartdoc to JavaScript, because it uses `dart:io`

mnordine:
Sorry, I meant what impact, if any, does using const lists have when compiled to js?

srawlins:
A valid question! I'm not sure though. Presumably they save space.

keertip (approved):

PR #3163, opened 2022-09-13
Refactor renderers

srawlins:
This is a big ball of... related... work. The two big pieces:

* Introduced two abstract classes, `ElementTypeRendererHtml` and `ElementTypeRendererMd`, which include a shared `_renderLinkedName` and `_renderNameWithGenerics` implementation. There was a lot of piecemeal HTML-string building which seemed fragile.
* Introduce unit tests for how various types get printed. This is needed in order to support record types. (I removed some duplicate test work from model_test.dart, which is an end-to-end test.)

Smaller work:

* `ModelElement.allParameters` - not even sure what this is for; it recursively walks parameter lists for function-typed parameters? Maybe it is just for referenceChildren, which is legit. Anyways, tidied.
* `ElementTypeRenderer.wrapNullability` - deprecated; more efficient to add to StringBuilder.
* Standardize on `StringBuffer` variables being named `buffer`, rather than `buf` or `output`. It was confusing to see different variable names.
* Standardize on variables being named `parameter` rather than sometimes `param`.
* Use cascades where possible.

### DIFF
--- a/lib/src/model/model_element.dart
+++ b/lib/src/model/model_element.dart
@@ -832,9 +832,9 @@ abstract class ModelElement extends Canonicalization
   late final List<Parameter> allParameters = () {
     var recursedParameters = <Parameter>{};
     var newParameters = <Parameter>{};
-    if (this is GetterSetterCombo &&
-        (this as GetterSetterCombo).setter != null) {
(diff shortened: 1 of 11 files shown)


keertip (approved):
Nice! 👍

Issue #3173, opened 2022-09-19
Rendering errors near search field


Hixie:
There are a few stray pixels rendering under the search bar.

<img width="152" alt="" src="https://user-images.githubusercontent.com/551196/191117145-684dd153-d523-41a2-8557-d0ebdaa535b0.png">

It appears they are from a magnifying glass of a text field rendered under the search text field.

Sometimes when you type text in the text field, stray pixels of text (maybe autocomplete text?) appear between the text field and the dropdown box.

<img width="513" alt="" src="https://user-images.githubusercontent.com/551196/191117384-1c464800-8a1c-43ea-a600-f2594a57a3f3.png">

PR #3179, opened 2022-09-20
Remove search hint feature

srawlins:
The Search Hint feature has not been rendering for some time... Not sure for how long. But if you type, at <https://api.flutter.dev>, something like "ele" and then hit Tab, it will autocomplete for you. But there is no typeahead hint showing the feature even exists.

If we want it back, reverting this and fixing the rendering issue might be straightforward.

Fixes https://github.com/dart-lang/dartdoc/issues/3173


devoncarew (approved):

srawlins:
CI failure seems transient. I'll land.

Issue #3175, opened 2022-09-19
Not clear which library search results are from


Hixie:
Searching for [State] returns results that appear to all be from the "widgets library": it says "widgets library", then "State class", then "State class", then "State constructor", then "BackgroundFetchClickEvent". The latter is actually from dart:html and has nothing to do with the others, but there is no clear separation between them to indicate that they are unrelated. The next result is back to being from the Flutter package which further confused the issue.

<img width="501" alt="" src="https://user-images.githubusercontent.com/551196/191123410-7daaaf6c-e09b-43e3-8c02-fe8cfba494df.png">

One could imagine fixing this in conjunction with https://github.com/dart-lang/dartdoc/issues/3171 by having every result be of the form "Class.member (library) description..." or some such, as in:

* `State` class (`widgets`) The logic and internal state for a Statef...
* `State` constructor (`widgets`)
* `BackgroundFetchClickEvent.state` (`dart:html`) 
* `PrimaryPointerGestureRecognizer.state` (`gestures`) The cur...

srawlins:
Fixed with your suggestions here: https://github.com/dart-lang/dartdoc/pull/3181.

I'll close this, but after another round of dartdoc bump and push to master-api, of course please re-open or open new bugs with the new UX.

PR #3181, opened 2022-09-21
Overhaul search results style

### LINKED ISSUE #3171: Search results dropdown should only use one line per result
I've been using the new Search experience in dartdoc for a few days now and I think we can improve it further with some small changes. Because each entry in the drop-down takes four lines and three font styles plus a big box, it's relatively hard to scan the results and there are very few results on the screen at a time.
 

I think this could be improved by using only one line per result with fewer words overall, more like a typical autocomplete UI, for example more like what GitHub does when you type `@` (but with the class and member names, and maybe library names, rather than people's identities, of course!).

srawlins:
This should fix #3171 and #3175. It's a big PR as I refactored a lot of code so that I could understand it better.

**Mini results**

Screen Shot 2022-09-20 at 8 44 05 PM

**Full page**

Screen Shot 2022-09-20 at 8 40 26 PM

* Separate the style of results shown in the "mini results" below the search bar, and those on the full search page.
  * Mini search results below search bar:
    * Use @Hixie's intuition that each result should be a single line of text
    * Add 60 pixels to the width, to accommodate for the single line results
    * Hide text overflow, but each result's full "one line doc" is available as title text (hover)
  * Full search bar: more or less unchanged
* Search code
  * `initializeSearch` was ~450 lines, so I split out the local functions where I was editing code.
  * Change anything named "category" or "group" to be "container" which more closely aligns with other dartdoc terminology for "a thing that can be an enclosing element".
  * Delete `iterateCategoriesMap` and just inline its code.
  * Rename `handle` to `handleSearch`.
  * Change many comments to doc comments.
  * Wrap many comments.
  * Change `handle`'s `bool` optional parameter to be named, to fit with Effective Dart.

There is a fair amount that could still be improved here (see other bugs), but (a) we don't have _any_ automated testing for the generated JS app 😦 and (b) I should stop here and land something.


srawlins:
CC @keertip @kevmoo

keertip (approved):
Difficult to review due to the refactor, the diff does not really show only code that has changed. Went by your comments as to the changes.

srawlins:
I know, I apologize ☹️ . I tried to stop refactoring as early as I could. And I didn't want 2 or 4 refactor PRs to precede this one as I hope to get this fixes out quickly. My bad for rushing it maybe. Follow up PRs for search should be smaller.

Issue #3172, opened 2022-09-19
"Enter" to see full search results doesn't work


Hixie:
If I press enter after typing in a search term with many results, I get a UI that says to press Enter to get all the results.

<img width="518" alt="" src="https://user-images.githubusercontent.com/551196/191116527-c4c400d4-a5f7-4db6-86f7-1a1273fe28d9.png">

However, nothing happens when I press enter.

kevmoo:
Reproduces on https://api.dart.dev/be/187340/index.html?

kevmoo:
Sadly, this seems unfixed. For reasons I do not understand!

Hixie:
FWIW, this doesn't work on https://master-api.flutter.dev/ as far as I can tell. I don't know what version of dartdoc that's on though, maybe we just haven't rolled far enough yet?

kevmoo:
It's on latest – 6.1.2 – it IS working on https://api.dart.dev/be/

Wondering if the JS is copied or modified with Flutter...

devoncarew:
We have a regular (semi-automated) process to roll the latest dartdoc into the dart sdk. My most recent knowledge here (which may be out of date) is that it's a manual process to roll the latest dartdoc into flutter.

devoncarew:
Ah; last updated (by @kevmoo) ~15 days ago: https://github.com/flutter/flutter/blob/master/dev/bots/docs.sh.

kevmoo:
...but there was ALSO some concern that the way flutter rolls dartdoc is non-standard. maybe the JS files are different than what's in the package, etc.

srawlins:
It's also weird that https://master-api.flutter.dev/search.html does not load. That could be related, as that is what pressing Enter will load.

Hixie:
This is not working on the Flutter API docs. Any chance we could reopen the issue and make it a high priority? We're about to have a significant branch cut.

cc @itsjustkevin @gspencergoog

kevmoo:
Yeah, we need this fixed. @keertip @srawlins @jacob314

itsjustkevin:
FYI @atsansone @MaryaBelanger

Issue #3027, opened 2022-04-19
Design and implement testing strategy for web app components


srawlins:
Likely using webdriver and pageloader.

CC @parlough @devoncarew

srawlins:
@yjbanov pointed me to https://github.com/flutter/engine/blob/main/lib/web_ui/dev/webdriver_browser.dart for the infra for Flutter Web's webdriver testing.

srawlins:
Then again, maybe an out-of-the-box Hybrid test from the test package will work. https://pub.dev/packages/test#browservm-hybrid-tests. I'm going to look into this first.

srawlins:
After implementing this hybrid style test, I think it is _not_ what we are looking for, as we need to send keys into the text field. I think we want a straight up black-box WebDriver test.

srawlins:
I'm doing a browser tab cleanup and just need to leave a note that I was looking into DDS's smoke test w.r.t. this idea. I cannot remember any details, so just posting a link: https://github.com/dart-lang/sdk/blob/main/pkg/dds/test/sse_smoke_test.dart

PR #3185, opened 2022-09-26
client search code: stop event propagation when navigating

kevmoo:
Fixes https://github.com/dart-lang/dartdoc/issues/3172


kevmoo:
I verified this locally. It's a "big lift", but we really should have interactive tests w/ a browser for some of these things. So much behavior that's completely interactive. We don't want to find breaks when we deploy to dart.dev or flutter.dev!

srawlins (approved):
🎉  Thanks!

devoncarew:
> It's a "big lift", but we really should have interactive tests w/ a browser for some of these things. So much behavior that's completely interactive. We don't want to find breaks when we deploy to dart.dev or flutter.dev!

This is probably worth tracking as a github issue? This may only be a few days for someone reasonably knowledgable in webdriver testing / whatever the current best practice is. As Kevin says, the alternative is finding breakages once the next version hits our big web properties.

srawlins:
@devoncarew got you covered with https://github.com/dart-lang/dartdoc/issues/3027

PR #3187, opened 2022-09-27
tool/grind.dart: cleanup directory signature calculation

kevmoo:
Use better conventions for stream processing
Drop now unneeded AccumulatorSink


srawlins (approved):
❤️ thanks!

PR #3191, opened 2022-09-27
Make InheritingContainer._inheritedElements late final

srawlins:
* Make it a late final initializer
* Extract a local function which was defined within a for loop, into an extension method
* Improve variable names
* Use a little UI-as-code


srawlins:
I'm pretty sure that failure is a transient flake...

srawlins:
Skipping past continuing CI failure

PR #3195, opened 2022-09-28
Prevent default when hitting enter

srawlins:
Fixes https://github.com/dart-lang/dartdoc/issues/3172

Use `event.preventDefault` instead of `event.stopPropagation`.

CC @kevmoo I don't know what I'm doing here. I tested this a lot and it seems to work OK.


kevmoo (approved):
Double-checked locally, too. Yup. User error. Well done! thanks!

web/search.dart
```diff
@@ -146,6 +146,10 @@ const _htmlEscape = HtmlEscape();
 
 final _containerMap = <String, Element>{};
 
+// TODO(srawlins): Break up this huge function into smaller parts. One big trick
+// here is maintaining how `selectedElement` is used. I suspect a class with
+// fields maintaining state would be a good solution. Another big trick is
```
kevmoo, line 151: +1 to putting all of the state in a private class!

Issue #3196, opened 2022-09-28
dartdoc styling seems off on bleeding edge and dev


mraleph:
Compare 

dev

![Screen Shot 2022-09-28 at 13 03 29](https://user-images.githubusercontent.com/131846/192763371-59c96d4e-7fed-4fa4-8174-a206f3898424.png)

beta

![Screen Shot 2022-09-28 at 13 04 05](https://user-images.githubusercontent.com/131846/192763482-4105c45c-6597-4f1a-a02b-aa2b5c4171f7.png)

stable

![Screen Shot 2022-09-28 at 13 04 38](https://user-images.githubusercontent.com/131846/192763595-9e07dbdc-6a94-48b7-abec-f67d737eda79.png)

Stable looks fine visually. Good padding between dropdown and contents, easy to read (method name first, classname under it). Then progressively UI deteriorates. Beta introduces weird boxes around method names, class name is too prominent, padding from top and left is inconsistent. Dev gets even worse. No padding on the left, too much padding on top, class name is written two times (on top and on the right). 

I suggest to go back to nice and compact output from the stable unless there is some strong reason why we don't want that. Otherwise maybe we need some CSS tweaks to make things more consistent. 

/cc @mit-mit

Hixie:
The screenshot for dev has two lines per result still, is that intentional? I agree that the stable screenshot is cleaner, but I think it'd be even cleaner if there was literally just one line per result (and no headers in the results) and also if the library was explicitly mentioned in the line (so you can distinguish dart:html and package:flutter results, for example).

srawlins:
> The screenshot for dev has two lines per result still, is that intentional?

No, accidental.

mraleph:
> I agree that the stable screenshot is cleaner, but I think it'd be even cleaner if there was literally just one line per result (and no headers in the results)

Yeah. Rust does this (though their design does not fit for us - because our search box is located in the corner) and theirs in the center.

![Screen Shot 2022-09-29 at 09 42 14](https://user-images.githubusercontent.com/131846/192970851-bf241d64-7730-457b-bf80-bd290f29c460.png)

Probably we could still render things as `dart:html > Event > stopPropagation` following the same format as breadcrumbs at the top of the page use. Though this will require extending the width of the dropdown.

PR #3197, opened 2022-09-28
Fix styles in inline search results

srawlins:
Fix each of the accidental issues noticed in #3196

* zero padding on the left of each result
* too much padding on the top
* class name is written twice

also:
* make the container italic, to visually separate it from the result name
* also the boldness of the suggestion prevented it from being highlighted with... boldness
* also if the search query did not match the _case_ of the result, the inline boldness would not appear

Screen Shot 2022-09-28 at 9 30 58 AM

Screen Shot 2022-09-28 at 9 31 45 AM


srawlins:
The CI failure is transient again.

kevmoo (approved):

srawlins:
I had neglected to upload screenshots. I've now attached them to the summary.

PR #3199, opened 2022-09-29
Save the nullability suffix of a type as a field

srawlins:
I think this String is computed many times, and it may be non-trivial. Better to make it a final field.


keertip (approved):
👍

PR #3200, opened 2022-09-29
Use more final; readable variable names

srawlins:
I'm walking through all of this code and modeling the record-rendering code after it. This just replaces `var` with `final` here and there, and makes some variable names shorter.


keertip (approved):

srawlins:
Pushing past the transient error.

PR #3217, opened 2022-10-11
Refactor ModelElement.element to be a getter; subclasses provide the field

srawlins:
I think this is the more correct layout for ModelElement and its subclasses. Code in ModelElement (and outside its subclasses) just needs to know about the Element. Subclasses define the backing field with Element subtypes. This does not seem to have grand performance characteristics, but should result in simpler code, fewer casts, and fewer parameters in super calls.


keertip (approved):
Totally makes sense! Nice!

PR #3231, opened 2022-10-26
Command-line arguments for limit max file count or total size.

isoos:


isoos:
/cc @jonasfj @srawlins

I couldn't replicate the `ResourceProvider` approach that `pub-dev` is using: https://github.com/dart-lang/pub-dev/blob/master/pkg/pub_dartdoc/lib/src/pub_hooks.dart#L35 as there was seems to be no natural way to hook it in based on the command-line arguments. However, the writer seems to be just as good place for it.

CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 ## 6.1.3-dev
 
 * Bump analyzer to 5.2.0.
-
+* `--max-file-count` and `max-total-size` command-line arguments to limit file system resource use.
```
jonasfj, line 4:
```suggestion
* `--max-file-count` and `--max-total-size` command-line arguments to limit file system resource use.
```

lib/src/dartdoc_options.dart
```diff
@@ -1592,6 +1597,14 @@ List<DartdocOption> createDartdocOptions(
...
             '`html` for html.',
         hide: false),
+    DartdocOptionArgOnly<String>('maxFileCount', '0', resourceProvider,
+        help:
+            'The maximum number of files dartdoc is allowed to create (0 for no limit).',
+        hide: false),
+    DartdocOptionArgOnly<String>('maxTotalSize', '0', resourceProvider,
+        help:
+            'The maximum total size dartdoc is allowed to create (0 for no limit).',
+        hide: false),
```
jonasfj, line 1607: I don't think we need these to show up in `--help`, I also don't mind if they do -- I think that's sort of a dartdoc product question @srawlins 
```suggestion
    DartdocOptionArgOnly<String>('maxFileCount', '0', resourceProvider,
        help:
            'The maximum number of files dartdoc is allowed to create (0 for no limit).',
        hide: true),
    DartdocOptionArgOnly<String>('maxTotalSize', '0', resourceProvider,
        help:
            'The maximum total size dartdoc is allowed to create (0 for no limit).',
        hide: true),
```

lib/src/dartdoc.dart
```diff
@@ -36,14 +37,45 @@ class DartdocFileWriter implements FileWriter {
   @override
   final Set<String> writtenFiles = {};
 
-  DartdocFileWriter(this._outputDir, this.resourceProvider);
+  final int? _maxFileCount;
+  final int? _maxTotalSize;
```
jonasfj, line 41: If `0` means "no limit", then we don't need nullability. No reason to have two values that both mean "no limit".
```suggestion
  final int _maxFileCount;
  final int _maxTotalSize;
```

lib/src/dartdoc.dart
```diff
@@ -36,14 +37,45 @@ class DartdocFileWriter implements FileWriter {
...
+
+  int _fileCount = 0;
+  int _totalSize = 0;
+
+  DartdocFileWriter(
+    this._outputDir,
+    this.resourceProvider, {
+    int? maxFileCount,
+    int? maxTotalSize,
```
jonasfj, line 50:
```suggestion
    int maxFileCount = 0,
    int maxTotalSize = 0,
```

lib/src/dartdoc.dart
```diff
@@ -36,14 +37,45 @@ class DartdocFileWriter implements FileWriter {
...
+        _maxFileCount! < _fileCount) {
+      throw DartdocFailure(
+          'Maximum file count reached: $_maxFileCount ($filePath)');
+    }
+    if (_maxTotalSize != null &&
+        _maxTotalSize! > 0 &&
+        _maxTotalSize! < _totalSize) {
+      throw DartdocFailure(
+          'Maximum total size reached: $_maxTotalSize ($filePath)');
+    }
```
jonasfj, line 69:
```suggestion
    if (_maxFileCount! > 0 &&
        _maxFileCount! < _fileCount) {
      throw DartdocFailure(
          'Maximum file count reached: $_maxFileCount ($filePath)');
    }
    if (_maxTotalSize! > 0 &&
        _maxTotalSize! < _totalSize) {
      throw DartdocFailure(
          'Maximum total size reached: $_maxTotalSize ($filePath)');
    }
```

jonasfj (approved):
LGTM, maybe @srawlins has concerns.

lib/src/dartdoc.dart
```diff
@@ -36,14 +37,41 @@ class DartdocFileWriter implements FileWriter {
...
+    this.resourceProvider, {
+    int? maxFileCount,
+    int? maxTotalSize,
+  })  : _maxFileCount = maxFileCount ?? 0,
+        _maxTotalSize = maxTotalSize ?? 0;
+
+  void _aboutToWrite(String filePath, int length) {
+    _fileCount++;
+    _totalSize += length;
+    print([filePath, _fileCount, _totalSize, _maxFileCount, _maxTotalSize]);
```
srawlins, line 57: Should this be removed?

lib/src/dartdoc.dart
```diff
@@ -36,14 +37,41 @@ class DartdocFileWriter implements FileWriter {
...
+    if (_maxFileCount > 0 && _maxFileCount < _fileCount) {
+      throw DartdocFailure(
+          'Maximum file count reached: $_maxFileCount ($filePath)');
+    }
+    if (_maxTotalSize > 0 && _maxTotalSize < _totalSize) {
+      throw DartdocFailure(
+          'Maximum total size reached: $_maxTotalSize ($filePath)');
```
srawlins, line 64: Should this have a unit here, `bytes` or something?

lib/src/dartdoc.dart
```diff
@@ -36,14 +37,41 @@ class DartdocFileWriter implements FileWriter {
...
+
+  DartdocFileWriter(
+    this._outputDir,
+    this.resourceProvider, {
+    int? maxFileCount,
+    int? maxTotalSize,
+  })  : _maxFileCount = maxFileCount ?? 0,
+        _maxTotalSize = maxTotalSize ?? 0;
+
+  void _aboutToWrite(String filePath, int length) {
```
srawlins, line 54: I think we can give this a name indicating it is checking maximums, like `_validateMaxWriteStats` or something...

lib/src/dartdoc.dart
```diff
@@ -60,14 +88,17 @@ class DartdocFileWriter implements FileWriter {
 
   @override
   void write(String filePath, String content, {Warnable? element}) {
+    final bytes = utf8.encode(content);
```
srawlins, line 91: I imagine this might increase memory pressure, even though it's local... we can verify whether it is or not with DevTools.

lib/src/dartdoc_options.dart
srawlins: I think they should not appear in `--help`

test/options_test.dart
```diff
@@ -584,4 +584,50 @@ class Foo {}
         throwsA(const TypeMatcher<DartdocFailure>().having((f) => f.message,
             'message', startsWith('Missing required template file'))));
   });
+
+  group('limit files created', () {
+    test('maxFileCount', () async {
```
srawlins, line 589: Let's add a test for each case where a Failure is not thrown, one for maxFileCount and one for maxTotalSize

lib/src/dartdoc_options.dart
```diff
@@ -1592,6 +1597,14 @@ List<DartdocOption> createDartdocOptions(
         help: 'The format of documentation to generate: `md` for markdown, '
             '`html` for html.',
         hide: false),
+    DartdocOptionArgOnly<String>('maxFileCount', '0', resourceProvider,
+        help:
+            'The maximum number of files dartdoc is allowed to create (0 for no limit).',
+        hide: false),
+    DartdocOptionArgOnly<String>('maxTotalSize', '0', resourceProvider,
```
srawlins, line 1604: This needs to reference some unit. Bytes?

lib/src/dartdoc_options.dart
```diff
@@ -1592,6 +1597,14 @@ List<DartdocOption> createDartdocOptions(
...
+        help:
+            'The maximum total size dartdoc is allowed to create (0 for no limit).',
```
srawlins, line 1606: I think we can use the term, 'write' here.

srawlins (changes requested):
Thanks!!

lib/src/dartdoc.dart
isoos: `File.writeAsStringSync` will also call `writeAsBytesSync(encoding.encode(contents), ...` internally, so it wouldn't really change much, the encoding happens a few instructions earlier, and maybe the GC of this object may trigger slightly later, but otherwise I don't think it will affect much.

isoos:
I've updated the PR and addressed all (most?) of the comments. PTAL!

lib/src/dartdoc.dart
```diff
@@ -48,24 +48,19 @@ class DartdocFileWriter implements FileWriter {
     this.resourceProvider, {
     int? maxFileCount,
     int? maxTotalSize,
-  })  : _maxFileCount = maxFileCount,
-        _maxTotalSize = maxTotalSize;
+  })  : _maxFileCount = maxFileCount ?? 0,
+        _maxTotalSize = maxTotalSize ?? 0;
```
jonasfj, line 53: nit: it's not nice to let callers pass `null`.

lib/src/dartdoc.dart
jonasfj: It would be interesting to see if writing `utf8` as a stream is faster, but that would probably require a lot of changes to the code -- if we don't write files concurrently it might not matter.

jonasfj (approved):

lib/src/dartdoc.dart
srawlins: Ah I see. This is good then.

> It would be interesting to see if writing `utf8` as a stream is faster

yeah I think for large packages we do spend a lot of time just pouring bytes onto disk, and I'd be game for any way to speed that up :D

isoos: IIRC I had experimented with tuning the writes at one point, making them happening async and also caching them and do writes after all the work was done in dartdoc. Neither proved significant change (but it revealed a costly operation in the fs abstraction which got fixed :). Unless we have a strong suspect where we have a bottleneck here, I wouldn't spend much more time on it.

srawlins (approved):
Sorry, lost track of this; looks great!

PR #3248, opened 2022-11-10
Remove unnecessary override

srawlins:


scheglov (approved):

srawlins:
Oof, mustachio does not like this. Reverting.

PR #3256, opened 2022-12-04
Fix error when use head base-href

### LINKED ISSUE #3172: "Enter" to see full search results doesn't work

parlough:
Fixes #3172


srawlins:
This base href stuff freaks me out. It affects all links, so requires a lot of test coverage, which we don't have, combinatorically speaking. :/ Also we don't have web tests. https://github.com/dart-lang/dartdoc/issues/3027

srawlins (approved):
🎉  Thank you!

srawlins:
Approving past the existing linter breakage.

PR #3259, opened 2022-12-05
Simplify search.html base-href handling

parlough:
After taking a closer look, it doesn't seem like `_relativePath` is necessary, as the `_htmlBase` logic seems to be sufficient unless I'm missing something. The `search.html` file should be in the same base location as the `index.json` which already used `_htmlBase`.

This PR standardizes on just using `_htmlBase`. If we generate the `search.html` URL, we are navigating to another page anyway, so I also don't bother storing it in a lazily initialized variable.

Tested manually with Dart SDK (using body base href) and Flutter (using head base href) docs.


srawlins:
Thanks @parlough , for what it's worth, I think the other test path for this base-href stuff is the markdown generation, which is a flat directory structure, I think. But we don't have search in markdown generation, so there is nothing to test for search-with-markdown.

srawlins:
I've been out, but I'd like to test this myself manually, and I'll post what I tested, by the end of the week.

srawlins:
OK I've reviewed manually, with docs that use `base href` and docs that don't. I've documented this manual testing strategy, and I'll be mailing a PR.

srawlins (approved):

srawlins:
Thanks for the simplification!

PR #3269, opened 2022-12-12
Tidy up grind.dart

srawlins:
Mostly style things:

* Make classes and top-level getters and variables private, so un-use can be determined.
* Wrap strings to 80 characters
* Short-circuit return in `checkFormat`
* Remove `smokeTest` and `longTest` tasks; all tests should be done, the slow ones have significantly sped up in the last few years.
* Do not start constants with `k`, like `kPubCachePathReplacement`.
* Remove unused parameters from `_buildTestPackageDocs`
* Make `label` a required parameter in `_FlutterRepo.fromPath`, `copyFromExistingFlutterRepo`, and `buildFlutterDocs`

I can break this up into 2 or 3 PRs if it would be significantly easier to review.

### DIFF
--- a/tool/grind.dart
+++ b/tool/grind.dart
@@ -12,6 +12,7 @@ import 'package:crypto/crypto.dart' as crypto;
 import 'package:dartdoc/src/io_utils.dart';
 import 'package:dartdoc/src/package_meta.dart';
 import 'package:grinder/grinder.dart';
+import 'package:meta/meta.dart';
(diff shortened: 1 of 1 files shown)


parlough:
This can be closed with https://github.com/dart-lang/dartdoc/commit/e8d00c079d83e9ee465cfeb9702eb2da765070d9 :) 

Great job with that huge migration.

srawlins:
Oh yes, huzzah!

PR #3271, opened 2022-12-12
Fix arrow key functions in search by using a flat list of suggestions

### LINKED ISSUE #3204: Arrow keys do not iterate through search suggestions in order
In this video I am hitting the down arrow key every time to move the highlight. The highlight does not move in the order I would have expected:

https://user-images.githubusercontent.com/1227763/193352809-c153e39d-596f-4d65-8360-2233b179da87.mov

srawlins:
Fixes #3204 

The problem is that the suggestions were grouped by their container, which is useful for the main body of search.html, but not for the search box.

This code is very hard for me to understand. https://github.com/dart-lang/dartdoc/issues/3272 would improve that.


pq (approved):

srawlins:
If Flutter is keen on having this fix, it is found in dartdoc 6.1.5, cc @goderbauer @gspencergoog

parlough:
@srawlins Consider cherry-picking 6.1.5 to beta if it's still possible. This also affects the SDK docs (and users who generate with `dart doc`) and it's a little awkward.

srawlins:
@parlough is api.dartdoc.dev really generated strictly from the `dartdoc` that is in the SDK? I hoped it was completely detached, like pub.dev is.

parlough:
@srawlins I believe it is generated as part of the release infrastructure with the dartdoc in third_party/pkg: https://dart.googlesource.com/recipes/+/refs/heads/main/recipes/release/sdk.py#155

I'm not super familiar with the SDK setup or those recipes, but I guess that should be the same one as in DEPS?

kevmoo:
@srawlins – I do believe it's based on the SDK dep. Went through this churn last time we rolled dartdoc

kevmoo:
CC @devoncarew RE rolling DEPS

sortie:
Yeah, the api.dart.dev documentation is built by the sdk recipe. It uses the Dart SDK that was just freshly build invoked on `third_party/pkg/dartdoc/bin/dartdoc.dart` to generate the dartdocs, which are then uploaded to cloud storage, which is then automatically served by api.dart.dev. Notably the documentation for older releases is not regenerated and remain as it was permanently. I hope that answers your questions :)

srawlins:
Thanks! Maybe I can cherry-pick one line in a DEPS roll CL

PR #3273, opened 2022-12-12
Don't show null in search results when no description

parlough:
This was showing up when a topic had a `null` description.

This also sorts topics the same as libraries.

Packages like [`package:go_router`](https://pub.dev/documentation/go_router/latest/) are starting to use topics as a more comprehensive destination for documentation, so it makes sense to me to priortize them as they are broad categories like libraries.

**Before:**
Before picture with null description

**After:** 
After picture with no null description


srawlins:
Thank you!!

kevmoo:
Approved by you, @srawlins ?

srawlins (approved):

Issue #3272, opened 2022-12-12
Refactor search into class with fields maintaining state.


srawlins:
Right now search is implemented in a fairly classical JavaScript way, with variables and functions all just sort of declared in a main method. Re-factoring this code into a class with fields can help improve readability,  understandability, and testability.

PR #3285, opened 2022-12-28
Refactor search into a stateful class

srawlins:
Fixes https://github.com/dart-lang/dartdoc/issues/3272 by changing `initialize` into a stateful class. Really helps with readability, promotion, etc.

This PR has a few little changes as well:

* rename 3 distinct variables named `input` which each represent something totally different
* add support for pressing 'Escape' in the search box
* fix result collection on search.html page
* `selectedElement` was a nullable int, with `null` meaning "nothing is selected." I found a nullable hint very awkward to work with, seeing lots of expressions like `selectedElement! - 1`. I was able to make this a non-nullable `int`, with `-1` meaning "nothing is selected," and some helper functions like an extension on `int` called `isBlurred`.


keertip (approved):
👍

PR #3297, opened 2023-01-14
Fix build by removing test opting out of null safety

parlough:
This was breaking the tests on the `dev` branch due to the new `illegal_language_version_override` diagnostic. We could possibly ignore it for now, but it seems any next release would be for Dart 3 alpha anyway.


srawlins (approved):
❤️  thank you!

PR #3302, opened 2023-01-14
Fix, cleanup, and simplify CSS

parlough:
- Use same search icon on both top nav and sidenav search bars (rather than pulling one from a website)
- Use a local SVG (from Material Symbols) for search icon
- Remove now unused search PNG
- Remove unnecessary padding causing horizontal scrollbar on offcanvas sidenav menu
- Remove various unused and debugging styles
- Adjust some formatting for easier reading of theme variables
- Remove unnecessary pixel specifiers
- Use shorthand margin and padding syntax where possible


srawlins (approved):
Thanks for running through all of these!

PR #3319, opened 2023-02-01
removed the validate-links command line flag and related functionality

devoncarew:
- removed the `validate-links` command line flag and related functionality

This feature was enabled by default when running dartdoc via `bin/dartdoc.dart`. From testing, when generating docs for package:win32, 90% of the generation time was spent parsing and validating html which we ourselves had just generated. I don't think this feature is worth the cost.

`dart doc` does expose the `--validate-links` flag (validation is off by default). When dartdoc next rolls into the sdk we'll need to update the `dart` tools, changelog, etc.


devoncarew:
It looks like the flutter bot is failing due to the fact that we no longer support a `no-validate-links` cli flag. I think we should just remove the use of the flag on the next roll. (For the `dart doc` command, we'd likely want to keep the flag, turn it into a no-op, and warn people when it is used)

srawlins:
I think we need flutter buy-in that they are ok with removing this functionality.

I also think the functionality is useful when making changes to dartdoc. So it is good to use in our tests and development workflows, but I don't think it is very useful for end users. The only cost is time, so maybe flutter and pub think the verification is worth it. 🤷

srawlins:
Oh, lol, I didn't read that closely enough. That's a big win then!

Issue #3320, opened 2023-02-01
integrate with markdown 7.0.0


srawlins:
markdown has many changes and breaking changes leading up to a 7.0.0 release. We need to bump our dependency on markdown and verify integration. These are the steps we need to make:

1. [ ] land a change here which bumps to a specific markdown 7.0.0-ish commit.
2. [ ] simultaneously bump markdown and dartdoc in Google internal.
3. [ ] simultaneously bump markdown and dartdoc in dart-lang/sdk DEPS.
4. [ ] release markdown 7.0.0.
5. [ ] land a change here which bumps to markdown ^7.0.0.

I see one issue already. One unit test is hanging. To repro, run `dart test test/end2end/dartdoc_integration_test.dart -p vm --plain-name 'invoking dartdoc on a basic package with --no-generate-docs is quiet and does not generate docs'`. I'll investigate.

srawlins:
CC @devoncarew @kevmoo

srawlins:
I haven't nailed it down but https://github.com/dart-lang/markdown/commit/ccd3c8b04c31b599cbbe007be0aac09d97642385 doesn't jive with dartdoc. dartdoc uses a custom BlockParser with a `parseLinesGenerator` that is hanging. Switching it to use `parseLines` (conceivably less performant) prevents the hanging.

srawlins:
OK I figured it out. The internals of BlockParser have changed substantially, and we can no longer implement our own little `parseLinesGenerator`. I am fairly certain though that it was unnecessary. It was used to avoid parsing multiple blocks in markdown documents when they weren't necessary but (a) most doc comments are 1 or 2 paragraphs, so no loss, and (b) even if we do parse all of the block elements, we still delay parsing the inline text, so there should be hardly any performance change.

_Also_, I think this would fix a bug. I imagine today we don't resolve links which refer to a link reference (`[text]: url` as its own block) in 1-paragraph doc summaries. Using `parseLines` instead of `parseLinesGenerator` should solve this... theoretical... bug.

srawlins:
I'm wrapping up a PR, will mail tonight.

srawlins:
I guess this is done. 🤷

PR #3321, opened 2023-02-02
Bump markdown dep to a pre-release of 7.0.0

srawlins:
* There are new newlines in table output
* Markdown 7.0.0 is incompatible with the custom `parseLinesGenerator` method. I remove it, and just use `parseLines`.
  * `parseLinesGenerator` was used to avoid parsing multiple blocks in markdown documents when they weren't necessary but I think we will suffer none-to-minimal performance loss:
      1. most doc comments are 1 or 2 paragraphs,
      2. even if we do parse all of the block elements, we still delay parsing the inline text,
      3. we are forced to use `sync*` in the generator, which is known to be slow
  * Also, I think this fixes an unreported bug. I imagine today we don't resolve links which refer to a link reference ([text]: url as its own block) in 1-paragraph doc summaries. Using parseLines instead of parseLinesGenerator should solve this... theoretical... bug.

Work towards #3320

### DIFF
--- a/lib/src/markdown_processor.dart
+++ b/lib/src/markdown_processor.dart
@@ -141,22 +141,6 @@ final List<md.BlockSyntax> _markdownBlockSyntaxes = [
 // Remove these schemas from the display text for hyperlinks.
 final RegExp _hideSchemas = RegExp('^(http|https)://');
 
-class _IterableBlockParser extends md.BlockParser {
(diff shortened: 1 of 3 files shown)


kevmoo (approved):

srawlins:
This is a bummer but I just have to merge even though it introduces a `pub publish` warning. 🤷

PR #3324, opened 2023-02-02
introduce an environmentProvider abstraction

devoncarew:
- introduce an `environmentProvider` abstraction into the `PackageMetaProvider` class
- update the two normal PackageMetaProvider instances (the one used by normal dartdoc, and the one used by tests) to pass in values for `environmentProvider`

This addresses an issue I had been seeing locally. For quite a while now I have not been able to run the dartdoc tests locally - I always saw several hundred test errors, and didn't know why I saw that locally vs. what I saw on the bots. I tracked this down to:
- the first `dart` version on my command line path is from an install of the flutter sdk
- that dart, when run, automatically has a value populated for the `FLUTTER_ROOT` env var
- having that var set changes what the `flutterRoot` dartdoc option will return
- tests use a mock sdk with a memory file system
- there was some validation that the dir pointed to by flutterRoot exists if set; this would throw an exception in these circumstances as `flutterRoot` was pointing to an (existing) path on disk, but is was validating against the memory file system

This PR allows tests to both provide an instance of an SDK as well as provide something you can get env. vars from; this ensures things are in sync (using the real file system and `Platform.environment`, or using the test file system and a mock env. provider).

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -1,5 +1,3 @@
-# Change analysis_options.yaml and analysis_options_presubmit.yaml
-# together.
 include: package:lints/recommended.yaml
 
 analyzer:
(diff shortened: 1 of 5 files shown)


srawlins (approved):
Clever! LGTM

srawlins:
Rebasing should clear the errors now.

Issue #3333, opened 2023-02-12
Running the command `dart doc --dry-run` creates an empty `UNUSED` folder


chykon:
Running the command `dart doc --dry-run` creates an empty `UNUSED` folder.

Running on Dart version 2.19.2:

```bash
$ dart --version
Dart SDK version: 2.19.2 (stable) (Unknown timestamp) on "linux_x64"
```

srawlins:
Oof that is unintentional. Good catch.

PR #3334, opened 2023-02-13
fix an issue where we created a temporary directory when performing a dry run

devoncarew:
- fix an issue where we created a temporary directory when performing a dry run
- fix https://github.com/dart-lang/dartdoc/issues/3333

### DIFF
--- a/lib/src/dartdoc.dart
+++ b/lib/src/dartdoc.dart
@@ -204,9 +204,6 @@ class Dartdoc {
     var libs = packageGraph.libraryCount;
     logInfo("Initialized dartdoc with $libs librar${libs == 1 ? 'y' : 'ies'}");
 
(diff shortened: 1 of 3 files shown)


lib/src/dartdoc.dart
```diff
@@ -204,9 +204,6 @@ class Dartdoc {
     var libs = packageGraph.libraryCount;
     logInfo("Initialized dartdoc with $libs librar${libs == 1 ? 'y' : 'ies'}");
 
-    // Create the out directory.
-    if (!_outputDir.exists) _outputDir.create();
```
devoncarew, line 208: This is created later on when actually generating docs (or at least, not creating it doesn't seem to prevent doc generation).

srawlins (approved):
Thanks!!

PR #3336, opened 2023-02-13
enable additional lints

devoncarew:
- enable additional lints (mostly found via trying out `package:dart_flutter_team_lints` here, and temporarily turning on `strict-casts` here)
- use a temp dir to avoid a test creating an `UNUSED` dir

Most of the rest of the strict cast violations look related to cli args parsing.

### DIFF
--- a/lib/src/failure.dart
+++ b/lib/src/failure.dart
@@ -6,7 +6,7 @@
 ///
 /// An instance of this is returned if dartdoc fails in an expected way (for
 /// instance, if there is an analysis error in the library).
-class DartdocFailure {
+class DartdocFailure implements Exception {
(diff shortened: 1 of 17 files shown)


lib/src/failure.dart
```diff
@@ -6,7 +6,7 @@
 ///
 /// An instance of this is returned if dartdoc fails in an expected way (for
 /// instance, if there is an analysis error in the library).
-class DartdocFailure {
+class DartdocFailure implements Exception {
```
srawlins, line 9: 👍 thanks!

srawlins (approved):
These are all so great, thanks!

PR #3404, opened 2023-05-02
Update validate-sdk-docs task to allow 20 libs

srawlins:


jcollins-g (approved):

jcollins-g (commented):
I don't think we support an SDK that has 18 libraries anymore, so that can be dropped, optionally.

srawlins:
I thought so too; I thought "well, it's probably 18 on old SDKs and 19 on newish" but I didn't want to investigate 🤣

Issue #3176, opened 2022-09-19
Search results rankings are not always what I'd expect


Hixie:
I'm not sure what to recommend regarding the sort order of members in the search results but here are some cases where I did not get the order I expected intuitively:

* [min]: the top-level member in `dart:math` seems more important than the properties on classes in `dart:html`, but the initial list only includes the latter.
* [max]: weirdly, [max] gives a very different set of results than [min], though it also doesn't include `dart:math`'s `max` which would be my top choice. I suspect most of the instances of `max` have a corresponding `min` and I would expect the results to reflect the same order for both.
* [Element]: This may seem backwards given what I just said about `dart:math` but I think the `widgets` library should take priority over `dart:html` in the Flutter docs.

srawlins:
I think we currently use _no_ heuristics regarding search order. So we're pretty free to implement anything that would improve anything 🤣 . There might be some logic about preferring top-level elements over instance methods, etc. But it's very light.

I think there is a pretty simple set of rules which would help these examples:

When documenting package X, and including the docs for other packages, Y, Z, and `dart:` packages:
* prefer results from package X
* then prefer results from the "core" SDK libraries: dart:async, dart:collection, dart:convert, dart:core, dart:developer, dart:math, and dart:typed_data
* then prefer the other `dart:` libraries
* then prefer the other packages, Y, Z.

Maybe weights need to be used so that a top-level element (maybe json`) from an SDK library is still preferred over an instance method on a class in package X.

Hixie:
We could try to be clever and add in some logic where members that are referenced more get higher rank, too. That would leverage the many cross-references in the Flutter docs. cc @gspencergoog

Levi-Lesches:
On top of the above heuristics, I'd suggest, for a search query `q`: 

1. A class that matches the whole of `q`
2. A top-level function/variable that matches the whole of `q`
3. A member of a class that matches the whole of `q`
4. A class that _contains_ `q`
5. A top-level function/variable that contains `q`
6. A member of a class that contains `q`.

That should fix the issues with `min` and `max`

InMatrix:
What @Levi-Lesches suggested sounds like an improvement. The current ranking method (or lack thereof) often leads to a quite broken experience. Here's what I got from searching "button":

<img width="580" alt="image" src="https://user-images.githubusercontent.com/348942/219752574-7c451893-9669-45c8-8f1e-b6799dc07803.png">

[...]

PR #3424, opened 2023-05-30
Refactor search ranking

srawlins:
A few items of note:

* Move search filtering and ranking to `lib/` and add rudimentary tests. (work towards https://github.com/dart-lang/dartdoc/issues/3419)
* Switch from a system of "scores" where each result gets an independent score to a system of ranks where results are definitely ordered amongst each other. Under the scores system, it was hard to know in what order results would ultimately be shown.
* Prefer same-named elements in higher-ranked packages to lower-ranked packages, as per the `--package-order` flag. (work towards https://github.com/dart-lang/dartdoc/issues/3176)
* Remove sort ranking differentiating by case (preferring case-sensitive matches to case-insensitive matches).


lib/src/search.dart
```diff
@@ -0,0 +1,203 @@
...
+
+    return allMatches.map((match) => match.item).toList();
+  }
+}
+
+class IndexItem {
+  final String name;
+  final String qualifiedName;
+
+  // TODO(srawlins): Store the integer of in the package order instead of this
```
jcollins-g, line 116: nit: `the index of the package in package order` or similar

lib/src/search.dart
```diff
@@ -0,0 +1,203 @@
...
+  });
+
+  // Example Map structure:
+  //
+  // ```dart
+  // {
+  //   "name":"dartdoc",
+  //   "qualifiedName":"dartdoc",
+  //   "href":"dartdoc/dartdoc-library.html",
+  //   "type":"library",
```
jcollins-g, line 144: this might be a field that could be crushed into an integer too, if needed.

lib/src/search.dart
```diff
@@ -0,0 +1,203 @@
...
+      packageName: data['packageName'],
+      href: data['href'],
+      type: data['type'],
+      overriddenDepth: (data['overriddenDepth'] as int?) ?? 0,
+      desc: data['desc'],
+      enclosedBy: enclosedBy,
+    );
+  }
+
+  int get _scope =>
```
jcollins-g, line 175: A comment explaining this would be nice.

jcollins-g (approved):
nits only.  Nice!

lib/src/search.dart
srawlins: Good call, done! And added a TODO, as its not well-tested.

lib/src/search.dart
```diff
@@ -0,0 +1,203 @@
...
+    required this.overriddenDepth,
+    required this.enclosedBy,
+  });
+
+  // Example Map structure:
+  //
+  // ```dart
+  // {
+  //   "name":"dartdoc",
+  //   "qualifiedName":"dartdoc",
```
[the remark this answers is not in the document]
srawlins: Done

lib/src/search.dart
srawlins: Yeah, it would be nice. We actually display the "type" in the search result so we would have to store the "type" as an int (rather convert to a scope), but yes still savings.

lib/src/search.dart
srawlins: Thanks! Done

test/search_test.dart
```diff
@@ -0,0 +1,147 @@
...
+import 'package:test_reflective_loader/test_reflective_loader.dart';
+
+void main() {
+  defineReflectiveSuite(() {
+    defineReflectiveTests(SearchTest);
+  });
+}
+
+@reflectiveTest
+class SearchTest {
```
devoncarew, line 16: 👍

devoncarew (commented):
Very cool!

PR #3426, opened 2023-05-30
Convert kind to an enum

srawlins:
The goal here is to benefit from some structure around "kind." For example, in the search functionality, it seems that updates may have been made to the models that were not reflected in the search ranking, such that top-level constants or properties were not ranked with other library members.

Using an enum here means we benefit from exhaustiveness checks, and we have the collection of kinds in one location.

I also:

* update the `index.json` field for kinds to be 'kind' (from 'type').
* store the kind enum index in `index.json` which is then used to get back the kind name, for display in search results.


jcollins-g (approved):
LGTM!

Issue #3423, opened 2023-05-30
Flutter search index is 27MB on disk


srawlins:
I just generated it, and it's 27MB. It is probably highly compressible, so maybe isn't a huge transfer cost. There are easy wins here:

* Remove overrides which do not have explicit docs.
* Create indexes. For example, an index of packageNames, and then each search item does not need to spell out its packageName. Additionally, each search item can be given an ID, and then the "enclosedBy" references can point to an ID.

PR #3427, opened 2023-06-01
Search: specify package rank in generation; distinguish non-core Dart libs

srawlins:
More work towards https://github.com/dart-lang/dartdoc/issues/3176.

Also helps with https://github.com/dart-lang/dartdoc/issues/3423, bringing the size of Flutter's search index down to 25MB.

* Instead of writing the String names of each element's package name, compute the package rank which is derived from the package name and the package order.
* Prefer "core" Dart libraries like dart:core over non-"core" libraries like dart:html or dart:io. This is the main request in https://github.com/dart-lang/dartdoc/issues/3176.
* Introduce the first tests for `index.json`.


jcollins-g (approved):
LGTM, but it looks like a bug in DartdocOptions might be causing a case mismatch on Windows?

PR #3446, opened 2023-06-16
Rank libraries lower than classes and class members

### LINKED ISSUE #3436: Libraries ranked higher than classes in search results
Some users found it surprising that a library is ranked higher than classes of the same name. Here's an example:

image

It makes sense that developers are more likely to be looking up classes and methods than libraries. @srawlins pointed out it's done at https://github.com/dart-lang/dartdoc/blob/main/lib/src/search.dart#L176. Maybe we can revisit this heuristic.

srawlins:
Fixes #3436


keertip (approved):
👍

srawlins:
The one CI bot is broken because of https://github.com/dart-lang/sdk/issues/52742.

PR #3457, opened 2023-06-29
Remove a non-useful type check on generic type variable

natebosch:
This type test was introduced during the null safety migration and I do
not see discussion about the intention.
In practice this is `assert(true)` so it is safe to remove.

I think the intention was to express that options which support multiple
values _must_ split commas, and now that this class is only used without
splitting commas the intention is that the options must be scalar
values.
It is not easy to perform this type of test on a type variable, so remove it.


natebosch:
This wouldn't catch cases where the type variable is bound to `Set`... so maybe it's better to just remove it after all?

cc @lrhn WDYT?

lib/src/dartdoc_options.dart
```diff
@@ -597,7 +597,8 @@ class DartdocOptionArgSynth<T> extends DartdocOption<T>
       String help = '',
       OptionKind optionIs = OptionKind.other,
       this.negatable = false})
-      : assert(T is! Iterable),
+      : assert(const <Never>[] is! T,
+            '$T should not be bound to a type which allows iterables'),
```
lrhn, line 601: This would allow `T` to be `Set`, which is iterable.

To check whether `T` is/is not a subtype of `Iterable`, you need to put `T` on the LHS of the `is` check, e.g., `<T>[] is! List<Iterable>`.

lrhn: But... the comment says to not allow any type which allows an iterable. That suggests not allowing`Object` either.
In practice any non-final type may allow an iterable, so I wouldn't try to do anything for that.

srawlins (approved):
Sorry for the delay, thanks!

PR #3474, opened 2023-08-06
Bump to analyzer 6.1.0

srawlins:


lib/src/generator/templates.runtime_renderers.dart
```diff
@@ -16380,6 +16380,8 @@ const _invisibleGetters = {
     'runtimeType'
   },
   'ClassElement': {
+    'augmentation',
```
scheglov, line 16383: How is it used? These methods are marked `@experimental`, meaning that we would not like clients to use them yet.

srawlins: See the [README for mustachio](https://github.com/dart-lang/dartdoc/tree/main/tool/mustachio#map-of-properties). Every property of important classes is enumerated here; it is not hand-picked properties, but every property.

We can update the generator to exclude `@experimental` properties.

scheglov: OK, now I understand that these are generated.
I think we should not using experimental properties outside of the analyzer.

srawlins:
They're not used. They're just made available. I can change the generator in a follow-up PR to not include `@experimental` properties.

lib/src/generator/templates.runtime_renderers.dart
scheglov: What I worry is that if we change these getters in the analyzer, this will break `dartdoc`.
If this is not an issue, we can land this of course.

scheglov (approved):

srawlins:
It's not a fun API. Everytime a getter is added or removed (even if it is never used), the generated code must be re-generated.

lib/src/generator/templates.runtime_renderers.dart
srawlins: Oh sorry, I missed where this is used. This is in the list of "invisible getters", so they cannot be used in the runtime renderers. An error is reported if an attempt is made to access this field.

PR #3476, opened 2023-08-11
Adjust links to dart.dev class modifier documentation

parlough:
The feature specification docs were moved, leading to a redirecting link, and there is now relevant documentation on dart.dev. 

\cc @MaryaBelanger Just so you're aware these docs are linked to from `dart doc` generated documentation. You can see an example of what they linked to before with the chips on the top of https://api.dart.dev/stable/3.0.7/dart-async/StreamConsumer-class.html.


srawlins (approved):
Thanks!

PR #3516, opened 2023-10-11
Make Library.inheritanceManager static

srawlins:
Better to not store a ton of junk in many InheritanceManagers. Just need one.

---

<details>

</details>


scheglov:
Why do you get many?
Theoretically there is just one per session.

srawlins:
> Why do you get many?

It's not a singleton class, as far as I can see; in the code before this change, each instance of Library (the Dartdoc class, not the Analyzer element class) has an instance variable, with a new InheritanceManager3. It should be better to have one instance (in the worst case its a no-op, but conceptually better to be static).

scheglov (approved):

PR #3519, opened 2023-10-11
Convert complex late final fields to getters: InheritingContainer, Library, and Package

srawlins:
Testing showed this change does not slow down documenting a package (collection 1.18.0), nor does it increase the peak RSS for the dartdoc process (documenting collection 1.18.0).

These are not the only late final fields in these classes; these are the ones defined by an immediately executed closure, which can be hard to read, reason about, and are generally uncommon and surprising.

---

<details>

</details>

### DIFF
--- a/lib/src/model/inheriting_container.dart
+++ b/lib/src/model/inheriting_container.dart
@@ -79,14 +79,14 @@ mixin Constructable on InheritingContainer {
 abstract class InheritingContainer extends Container
     with ExtensionTarget
     implements EnclosedElement {
-  late final DefinedElementType? supertype = () {
+  DefinedElementType? get supertype {
     final elementSupertype = element.supertype;
     return elementSupertype == null ||
             elementSupertype.element.supertype == null
         ? null
         : modelBuilder.typeFrom(elementSupertype, library)
             as DefinedElementType;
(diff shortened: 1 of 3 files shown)


lib/src/model/inheriting_container.dart
```diff
@@ -79,14 +79,14 @@ mixin Constructable on InheritingContainer {
 abstract class InheritingContainer extends Container
     with ExtensionTarget
     implements EnclosedElement {
-  late final DefinedElementType? supertype = () {
+  DefinedElementType? get supertype {
```
scheglov, line 82: We don't need caching these values?

scheglov (approved):
If you are OK with re-computing these, LGTM.

lib/src/model/inheriting_container.dart
srawlins: I ran some benchmarks that showed no increase in time-to-document, nor an increase in RSS (or at least, if there is, it's noise).

But if you're asking about returning a new instance at each call, dartdoc's "ModelElement"s are all cached in the package graph, so `modelBuilder.typeFrom` should return the same instance every time, when called with the same type and library, like on line 87.

PR #3521, opened 2023-10-11
Move doc directive documentation to source code

srawlins:
This documentation currently lives in [the wiki](https://github.com/dart-lang/dartdoc/wiki/Doc-comment-directives). However, we are planning to move the dartdoc code to the Dart SDK repo, and this wiki will be going away.

The text in this PR is more-or-less a dump of the wiki contents, and I tweaked the wording and typos a bit.

A full review of the text is fine, but I will likely respond for requests to document "more" or document "better" by writing in TODOs. I still encourage a full read of the text.

---

<details>

</details>

### DIFF
--- a/doc/directives.md
+++ b/doc/directives.md
@@ -14,6 +14,9 @@ The supported directives are listed below:
 An element whose doc comment should not appear in the generated documenation can
 include the `@nodoc` directive.
 
+Note that the `@nodoc` directive does not have curly braces, like most of the
+other directives.
+
 ## `{@category}` and `{@subCategory}` - Categories
 
 Elements such as libraries and classes can be grouped into categories and
@@ -23,7 +26,7 @@ own documentation page, listing all of the categorized elements.
 
 ## `{@template}` and `{@macro}` - Templates and macros
 
-TODO(srawlins): Document
+TODO(srawlins): Document.
 
 ## `{@example}` - Examples (deprecated)
 
@@ -32,7 +35,7 @@ directive. The file path, the region, and the example language can all be
 specified with the following syntax:
 
 ```none
-{@example PATH [region=NAME] [lang=NAME]}
+/// {@example PATH [region=NAME] [lang=NAME]}
 ```
 
 All example file names must have the extension, `.md`, and this extension must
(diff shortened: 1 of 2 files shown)


parlough:
For non-deprecated directives that are well supported, especially the ones now supported by the analyzer, it would be nice for their docs to live on dart.dev (https://github.com/dart-lang/site-www/issues/4347).

Perhaps a proper review of the text can wait until that is worked on?

srawlins:
Thanks for the link @parlough !! I like this direction. I think we could maybe document each directive one-by-one, where we can focus on getting the text and semantics and support correct.

parlough:
That sounds great to me.

I've mostly delayed that work since I wasn't sure if the directives would stick around and in what way. Same for the comment references with the import work.

doc/directives.md
```diff
@@ -0,0 +1,117 @@
...
+place of the directive. Not all directives are supported for package
+documentation at https://pub.dev/.
+
+A doc comment directive consists of either one directive tag, which looks
+something like `{@DIRECTIVE ARG=VALUE ...}`, or two such tags (an opening and a
+closing tag), with content in between.
+
+The supported directives are listed below:
+
+## `@nodoc` - Do not include documentation
```
bwilkerson, line 12: Should this have `{` and `}` for consistency with what's below, or is this a special case (in which case we should call that out explicitly)?

doc/directives.md
```diff
@@ -0,0 +1,117 @@
...
+
+An element whose doc comment should not appear in the generated documenation can
+include the `@nodoc` directive.
+
+## `{@category}` and `{@subCategory}` - Categories
+
+Elements such as libraries and classes can be grouped into categories and
+sub-categories by adding `{@category CATEGORY NAME}` and
+`{@subCategory SUB-CATEGORY NAME}` in doc comments. Each category then gets its
+own documentation page, listing all of the categorized elements.
```
bwilkerson, line 22: We should consider coming back at some point and removing / generalizing descriptions of the output so that this doesn't need to change every time we change the render engine (especially if we separate the render engine from this package).

doc/directives.md
```diff
@@ -0,0 +1,117 @@
...
+
+## `{@inject-html}` - Injected HTML
+
+HTML can be rendered unmodified by including it between `{@inject-html}` and
+`{@end-inject-html}` directive tags. The tags take no arguments:
+
+```none
+{@inject-html}
+INJECTED HTML
+{@end-inject-html}
```
bwilkerson, line 57: Do we want to make the examples valid code by inserting `\\\` before each line? It's not necessary, but it might make it easier to copy/paste.

bwilkerson (approved):
Nice.

doc/directives.md
srawlins: Ah that's a good point. I think it is important to tell the user what the end result of creating and using categories is. But I think you're right: the text here should be a little broad, and a separate doc should flesh out the concept of categories and how they are displayed and sorted, etc.

PR #3525, opened 2023-10-13
Fix missing search in left/mobile sidenav

parlough:
This can be seen on any class after https://github.com/dart-lang/dartdoc/commit/c147aa19a2a655a2cf9af7a8e6a0f104e6116137. For example on the `main` SDK docs: https://api.dart.dev/be/44f99e1fb2ff7f849cca18d2203fd7cd52c73d2b/dart-core/DateTime-class.html

https://github.com/dart-lang/dartdoc/commit/c147aa19a2a655a2cf9af7a8e6a0f104e6116137 fixed the sidebar showing up for the non-class templates, but in doing so, overrode the rest of the left sidebar, including the search. The fix is to add back the removed `-content` child and add it to the other templates as well.


srawlins (approved):
Thanks, @parlough ! I will do a local check of this as well tomorrow.

parlough:
As always, thank _you!_

srawlins:
Sorry I took a while to review. But I just played with this locally and see the bug and the fix. Thank you! I really want to dive back into some front-end tests. I might have a lead there.

parlough:
Thanks for the review. As for the testing, I'm always happy to follow your lead (and learn from it)!

parlough:
@srawlins I saw you're working on a lot of cool changes! Do you think you would be able to publish 7.0.1 before those land though?

I'm away for a few days, but it would be appreciated :)

srawlins:
Yes, I rolled this change into the Dart SDK today, so that will roll into google internal tomorrow, and then I'll push a 7.0.1.

parlough:
Thanks!

PR #3538, opened 2023-10-19
Remove use of NodeLocator2

srawlins:
When we first resolve a library, we have full access to both the LibraryElement and the CompilationUnit (AST nodes) objects. We should use those pronto, rather than painstakingly recalculate them with NodeLocator2.

This is a pretty big ball of wax. But this is one step in untangling it. Almost everything we need to know about source elements in Dartdoc is on the element model, with two exceptions:

* Element `documentationComment` fields only contain the documentation comment text; it's a String. To get knowledge about comment references, we have to go back to the syntax tree.
* In order to extract the literal source code, to write in the "implementation" section of docs, we must use offsets only found in the syntax tree.

To hang onto some of this syntax tree data, without hanging onto any AST nodes, we do some pre-processing into ModelNode objects. This pre-processing is led by Library.fromLibraryResult. It used to use an ElementVisitor named `_HashableChildLibraryElementVisitor`. This is simpler now, and called `_ModelNodeGatherer`.

---

<details>

</details>

### DIFF
--- a/lib/src/model/library.dart
+++ b/lib/src/model/library.dart
@@ -4,10 +4,11 @@
 
 import 'dart:collection';
 
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:analyzer/dart/element/element.dart';
 import 'package:analyzer/dart/element/scope.dart';
 import 'package:analyzer/dart/element/type_system.dart';
-import 'package:analyzer/dart/element/visitor.dart';
 import 'package:analyzer/source/line_info.dart';
 // ignore: implementation_imports
 import 'package:analyzer/src/generated/sdk.dart' show SdkLibrary;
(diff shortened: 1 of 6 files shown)


lib/src/model/library.dart
```diff
@@ -16,113 +17,55 @@ import 'package:dartdoc/src/model/model.dart';
...
-    packageGraph.populateModelNodeFor(element, resolvedLibrary);
-  }
-
-  @override
-  void visitTypeParameterElement(TypeParameterElement element) {
-    packageGraph.populateModelNodeFor(element, resolvedLibrary);
+  _ModelNodeGatherer(this.resolvedLibrary, this.packageGraph);
+
+  @override
+  void visitCompilationUnit(CompilationUnit node) {
```
scheglov, line 37: Why do we use visitor, if we just iterate manually?

srawlins: Well said! OK I've dropped the visitor, and made this a method on PackageGraph. Much simpler.

scheglov (approved):

PR #3552, opened 2023-10-25
Improve display text for libraries

srawlins:
Fixes https://github.com/dart-lang/dartdoc/issues/1658 and fixes https://github.com/dart-lang/dartdoc/issues/1000.

The changes in this PR do not ever print the full import URI (like `package:googleapis/admob/v1.dart`) or even the import path with the `.dart` extension (like `admob/v1.dart`), except in the breadcrumb. Mainly this PR uses the basename of the library with its directory path _after_ `lib/` (like `admob/v1`). Here are some screenshots:

Screenshot 2023-10-25 at 11 17 44 AM

---

Screenshot 2023-11-01 at 10 20 47 PM

---

Screenshot 2023-11-01 at 10 21 04 PM

---

Screenshot 2023-10-25 at 11 54 47 AM

---

Screenshot 2023-11-02 at 1 13 59 PM

---

Screenshot 2023-11-02 at 1 14 08 PM

---

Additionally:

* Make `Library.prefixToLibrary` private.
* Refactor `Library.dirName` to just inline all of its helper methods. I think it is much more readable this way.
* Remove `Library.nameFromPath` (as part of the above).

---

<details>

</details>

### DIFF
--- a/lib/src/generator/template_data.dart
+++ b/lib/src/generator/template_data.dart
@@ -215,8 +215,11 @@ class LibraryTemplateData extends TemplateData<Library>
   List<Documentable> get navLinks => [_packageGraph.defaultPackage];
 
   @override
-  String get layoutTitle => _layoutTitle(library.name, Kind.library.toString(),
-      isDeprecated: library.isDeprecated);
+  String get layoutTitle => _layoutTitle(
+        library.breadcrumbName,
+        Kind.library.toString(),
+        isDeprecated: library.isDeprecated,
+      );
 
   @override
   Library get self => library;
--- a/lib/src/generator/template_data.dart
+++ b/lib/src/generator/template_data.dart
@@ -104,9 +104,9 @@ abstract class TemplateData<T extends Documentable> extends TemplateDataBase {
   String? get belowSidebarPath => self.belowSidebarPath;
 
   String _layoutTitle(
-    String name,
-    String? kind, {
-    required bool isDeprecated,
+    String name, {
+    String? kind,
+    bool isDeprecated = false,
   }) =>
       _packageGraph.rendererFactory.templateRenderer
           .composeLayoutTitle(name, kind, isDeprecated);
@@ -140,8 +140,11 @@ class PackageTemplateData extends TemplateData<Package> {
   Package get self => package;
 
   @override
-  String get layoutTitle =>
-      _layoutTitle(package.name, package.kind.toString(), isDeprecated: false);
+  String get layoutTitle => _layoutTitle(
+        package.name,
+        kind: package.kind.toString(),
+      );
+
   @override
   String get metaDescription =>
       '${package.name} API docs, for the Dart programming language.';
@@ -187,9 +190,10 @@ class CategoryTemplateData extends TemplateData<Category>
   String get title => '${category.name} ${category.kind} - Dart API';
 
   @override
-  String get layoutTitle =>
-      _layoutTitle(category.name, category.kind.toString(),
-          isDeprecated: false);
+  String get layoutTitle => _layoutTitle(
+        category.name,
+        kind: category.kind.toString(),
+      );
 
   @override
   String get metaDescription =>
@@ -221,7 +225,6 @@ class LibraryTemplateData extends TemplateData<Library>
   @override
   String get layoutTitle => _layoutTitle(
         library.breadcrumbName,
-        null, // Do not display the word, 'library'
         isDeprecated: library.isDeprecated,
       );
 
@@ -291,9 +294,12 @@ abstract class InheritingContainerTemplateData<T extends InheritingContainer>
       '${library.name} library, for the Dart programming language.';
 
   @override
-  String get layoutTitle =>
-      _layoutTitle(clazz.nameWithLinkedGenerics, clazz.fullkind,
-          isDeprecated: clazz.isDeprecated);
+  String get layoutTitle => _layoutTitle(
+        clazz.nameWithLinkedGenerics,
+        kind: clazz.fullkind,
+        isDeprecated: clazz.isDeprecated,
+      );
+
   @override
   List<Documentable> get navLinks => [_packageGraph.defaultPackage, library];
 }
@@ -328,9 +334,12 @@ class ExtensionTemplateData<T extends Extension> extends TemplateData<T>
       '${library.name} library, for the Dart programming language.';
 
   @override
-  String get layoutTitle =>
-      _layoutTitle(extension.name, extension.kind.toString(),
-          isDeprecated: false);
+  String get layoutTitle => _layoutTitle(
+        extension.name, kind: extension.kind.toString(),
+        // TODO(srawlins): Why can't an extension be deprecated?
+        isDeprecated: false,
+      );
+
   @override
   List<Documentable> get navLinks => [_packageGraph.defaultPackage, library];
 }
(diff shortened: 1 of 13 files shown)


devoncarew:
For my 0.02, I do think it's useful to show the file extension (`admob/v1.dart` instead of `admob/v1`), even at the expense of some brevity. It helps the user know what they're looking at w/o having to think about it, and will be familiar with what they'd import or what they'd see in their IDE.

srawlins:
CC @kevmoo @goderbauer

kevmoo:
> For my 0.02, I do think it's useful to show the file extension (`admob/v1.dart` instead of `admob/v1`), even at the expense of some brevity. It helps the user know what they're looking at w/o having to think about it, and will be familiar with what they'd import or what they'd see in their IDE.

I'd actually argue LEAVE the `.dart` out and ship that way (maybe have it in a tooltip) and see if there are complains!

srawlins:
@goderbauer I found a few differences in how this shows api.flutter.dev; almost zero libraries are located deeper than `lib/` so this PR changes very little for Flutter. But the `material_color_utilities` package has some:

### BEFORE

Screenshot 2023-10-25 at 1 05 25 PM

---

### AFTER

Screenshot 2023-10-25 at 1 04 45 PM


There may be some misconceptions about `src/` directories in that package's layout.

srawlins:
For now, there were more recommendations to leave the `.dart` _off_ of library names, and I agree it kind of noises up the left sidebar for api.flutter.dev.

And yes, I really like the tooltip idea. I'd like to land that as well, separately.

goderbauer:
> There may be some misconceptions about src/ directories in that package's layout.

I would agree with that. Given the current layout, dartdoc's output seems reasonable, though. I'll see if we can fix up the layout, but that shouldn't block this work here.

srawlins:
Hi @kallentu , as part of learning the code here, you could participate in code review for this PR. You can read over the two issues that are fixed by this PR and ask me any questions about the code.

lib/src/model/library.dart
```diff
@@ -197,9 +197,40 @@ class Library extends ModelElement
...
+  /// An identifier for this library based on its location.
+  ///
+  /// This provides filename collision-proofing for anonymous libraries by
+  /// incorporating more from the location of the anonymous library into the
+  /// name calculation. Simple cases (such as an anonymous library in 'lib/')
+  /// are the same, but this will include slashes and possibly colons
+  /// for anonymous libraries in subdirectories or other packages.
+  late final String dirName = () {
+    String nameFromPath;
+    if (isAnonymous) {
```
kallentu, line 209: Quick question: What are anonymous libraries? When do these occur?

lib/src/model/library.dart
```diff
@@ -281,15 +312,36 @@ class Library extends ModelElement
...
+    var fullName = breadcrumbName;
+    if (fullName.endsWith('.dart')) {
+      const dartExtensionLength = '.dart'.length;
+      return fullName.substring(0, fullName.length - dartExtensionLength);
+    }
+    return fullName;
+  }
+
+  @override
+  String get breadcrumbName {
```
kallentu, line 326: Just a thought -- Is it weird that the breadcrumb changes from 
`googleapis > admob.v1 library` to 
`googleapis > admob.v1.dart > ...`? And similar with other parts like `googleapis package` to `googleapis`

We already show that it's `admob.v1 library` in the main title on the page. I just find the change in breadcrumbs a little bit odd?

lib/src/model/library.dart
srawlins: An anonymous library is either a library without a `library` directive, or one where the `library` directive has no name (`library;`).

lib/src/model/library.dart
srawlins: I presented this view to @goderbauer and he liked this functionality. I don't feel strongly either way. Most people seemed very against putting ".dart" everywhere but I like it so it's pretty subjective.

I agree there is something to be said for consistency. But Mike really didn't want the `<h1>` to be "animation.dart" or "animation.dart library"; too clinical maybe; not friendly. But in the breadcrumb that name is very meaningful for answering "where do I find this widget / class / function?"

kallentu: Oh yeah, I meant instead of changing the header to be "animation.dart", I was wondering if we could keep the breadcrumbs consistent as we open it and still keep the `.dart` part.

So when you're on the `admob.v1` page, you'd see `googleapis > admob.v1.dart` instead of `googleapis > admob.v1 library` in the breadcrumb, but the `<h1>` can stay the way it is right now. And then the next breadcrumb for App would line up well being `googleapis > admob.v1.dart > App`.

_shrug_ What we have is already fine, but I'm not fond of the breadcrumbs changing as we click into different parts of the library.

srawlins: > So when you're on the admob.v1 page, you'd see googleapis > admob.v1.dart instead of googleapis > admob.v1 library in the breadcrumb, but the <h1> can stay the way it is right now.

Ah I see. *shrug* yeah Mike commented that he liked this way; I don't want to do a whole UX study on it so I'm happy with this line-up. But 100% with the consideration that we can change these things; I don't think it would be jarring to make a few changes over a few months to the sidebar and breadcrumb text, if it doesn't work well for some packages or api.dart.dev or whatever.

Mike, to double check, WDYT about the breadcrumb text changing, when you move from a library page to an element inside that library?

goderbauer: For the h1 heading I very much would like to keep it as `animation library` instead of `animation.dart library`. Animation library is what we call this when we talk amongst ourselves or with customers. Nobody says "oh, you can find this in the animation _dot dart_ library". The docs should reflect this real-world use, IMO.

For the breadcrumps I feel less strongly. They feel more technical to me. If we want to change it there to always show the `.dart` for consistency I am not opposed to that. Now that I am thinking about this some more, I would actually slightly lean towards consistency here.

srawlins: SG! I'll make them consistently 'animation.dart' and 'admob/v1.dart' for example and post new screenshots.

srawlins: Screenshots are updated; this is ready for review.

goderbauer: Is the screenshot about the animation library still outdated - or is there a bug? I would have expected to see `animation.dart` in the breadcrumb now.

srawlins: Sorry I hadn't updated the Flutter screenshots (takes so long to generate!); updated now.

lib/src/generator/template_data.dart
```diff
@@ -103,7 +103,11 @@ abstract class TemplateData<T extends Documentable> extends TemplateDataBase {
 
   String? get belowSidebarPath => self.belowSidebarPath;
 
-  String _layoutTitle(String name, String kind, {required bool isDeprecated}) =>
+  String _layoutTitle(
+    String name,
+    String? kind, {
```
kallentu, line 108: Could make this a named parameter, then you don't need to send `null` over in the getter.

kallentu (approved):

lib/src/model/library.dart
kallentu: They look good, thanks for updating them Sam!

goderbauer: Agree. Screenshots look great!

lib/src/generator/template_data.dart
srawlins: Good call. Done.

PR #3571, opened 2023-11-04
Update RegExps in documentation_comment.dart

lrhn:
Some pure clean-up and simplification.
Some "linear backtracking"-protection.
One real change, from `[^ ]+` to `\S+`, to not allow tabs or newlines in an unquoted argument.

### DIFF
--- a/lib/src/model/documentation_comment.dart
+++ b/lib/src/model/documentation_comment.dart
@@ -14,22 +14,19 @@ import 'package:meta/meta.dart';
 import 'package:path/path.dart' as p show Context;
 
 final _templatePattern = RegExp(
-    r'[ ]*{@template\s+(.+?)}([\s\S]+?){@endtemplate}[ ]*(\n?)',
-    multiLine: true);
+    r'[ ]*\{@template\s+(\S.*?)\}([^]+?)\{@endtemplate\}[ ]*(\n?)');
 final _htmlPattern = RegExp(
-    r'[ ]*{@inject-html\s*}([\s\S]+?){@end-inject-html}[ ]*\n?',
-    multiLine: true);
+    r'[ ]*\{@inject-html\s*\}([^]+?)\{@end-inject-html\}[ ]*\n?');
 
 /// Matches all tool directives (even some invalid ones). This is so
 /// we can give good error messages if the directive is malformed, instead of
 /// just silently emitting it as-is.
 final _basicToolPattern = RegExp(
-    r'[ ]*{@tool\s+([^}]+)}\n?([\s\S]+?)\n?{@end-tool}[ ]*\n?',
-    multiLine: true);
+    r'[ ]*{@tool\s+([^\s}][^}]*)}\n?([^]+?)\n?{@end-tool}[ ]*\n?');
 
-final _examplePattern = RegExp(r'{@example\s+([^}]+)}');
+final _examplePattern = RegExp(r'{@example\s+([^\s}][^}]*)}');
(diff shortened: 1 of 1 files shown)


lib/src/model/documentation_comment.dart
```diff
@@ -14,22 +14,19 @@ import 'package:meta/meta.dart';
 import 'package:path/path.dart' as p show Context;
 
 final _templatePattern = RegExp(
-    r'[ ]*{@template\s+(.+?)}([\s\S]+?){@endtemplate}[ ]*(\n?)',
-    multiLine: true);
+    r'[ ]*\{@template\s+(\S.*?)\}([^]+?)\{@endtemplate\}[ ]*(\n?)');
```
lrhn: Added `\` before `{` and `}` just be cause it makes it clear they're not semantic.
The `[^]` is the same as `[\s\S]` (match *any* character).
The `multiLine: true` isn't used, there is no `^` or `$` in the regexp.

The `\s+(.+?)\}` pattern risks "linear backtracking" (not as bad as catastrophic backtracking, but still unnecessary).
An input of 
```

-- next line has many spaces -- 
{@template                                                                                      a
}
```  
will let all the space characters match `\s+`, then followed by an `a`, which must be matched by the `.`, and a newline which matches neither `.` nor `}`.
So it goes back and matches one less space in the `\s+`, lets `.` match that space instead, and continue to still not match. In the end, it tries all the combinations before giving up. 

Changing to `\s+([^\s}].*?)` makes the transition-point determinisitic, and also prevents matching `{@template }}` as having name `}`

lib/src/model/documentation_comment.dart
```diff
@@ -675,8 +672,8 @@ mixin DocumentationComment
   /// Match group 4 is the unquoted arg, if any.
   static final RegExp _argPattern = RegExp(r'([a-zA-Z\-_0-9]+=)?' // option name
       r'(?:' // Start a new non-capture group for the two possibilities.
-      r'''(["'])((?:\\{2})*|(?:.*?[^\\](?:\\{2})*))\2|''' // with quotes.
-      r'([^ ]+))'); // without quotes.
+      r'''(["'])((?:[^\\\r\n]|\\.)*?)\2|''' // with quotes.
```
lrhn, line 675: This should be equivalent, but simpler. The original RE stopped at the next quote not preceded by an odd number of backslashes.
The new one must match the character after a backslash as content.
Effect should be the same, just much easier to read.

lib/src/model/documentation_comment.dart
```diff
@@ -675,8 +672,8 @@ mixin DocumentationComment
...
+      r'(\S+))'); // without quotes.
```
lrhn, line 676: This is a change. It didn't seem reasonable to accept newlines or tabs as part of an unquoted argument. Matching non-whitespace seemed better than just non-space-char.

srawlins:
Thanks much @lrhn ! We're replacing these regular expressions with parsing in the analyzer. See [doc_comment_builder.dart](https://github.com/dart-lang/sdk/blob/main/pkg/analyzer/lib/src/fasta/doc_comment_builder.dart) and in particular, [`_parseDocDirectiveTag`](https://github.com/dart-lang/sdk/blob/6137e932417faa83e93719a7a7fb3bc4e2189ffa/pkg/analyzer/lib/src/fasta/doc_comment_builder.dart#L350).

srawlins (approved):

PR #3572, opened 2023-11-05
Simplify package_meta.dart

srawlins:
A few things:

* `PackageMetaProvider.getMessageForMissingPackageMeta` was a very round-about way to provide some text to an error-message. It turns out every single call site referenced the same message, so just tear down all of the redirection.
* Make `PackageMeta.pathContext` private.
* Simplify `PackageMetaProvider.fromDir` with `putIfAbsent`.

---

<details>

</details>

### DIFF
--- a/lib/src/package_meta.dart
+++ b/lib/src/package_meta.dart
@@ -11,7 +11,6 @@ import 'package:analyzer/file_system/file_system.dart';
 import 'package:analyzer/file_system/physical_file_system.dart';
 // ignore: implementation_imports
 import 'package:analyzer/src/generated/sdk.dart' show DartSdk;
-import 'package:dartdoc/src/dartdoc_options.dart';
 import 'package:dartdoc/src/failure.dart';
 import 'package:meta/meta.dart';
 import 'package:path/path.dart' as path;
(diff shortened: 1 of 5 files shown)


lib/src/package_meta.dart
```diff
@@ -23,9 +22,13 @@ class PackageMetaFailure extends DartdocFailure {
...
 
-/// For each list in this list, at least one of the given paths must exist
-/// for this to be detected as an SDK.  Update `_writeMockSdkBinFiles` in
-/// `test/src/utils.dart` if removing any entries here.
+/// Various relative paths that indicate that a root direoctory is an SDK (Dart
+/// or Flutter).
+///
+/// For a given directory to be detected as an SDK, at least one of the given
+/// paths must exist, for each list of paths.
+// Update `_writeMockSdkBinFiles` in `test/src/utils.dart` if removing any
```
kallentu, line 30: `///` for line 30 and 31?

lib/src/package_meta.dart
```diff
@@ -104,6 +91,8 @@ class PackageMetaProvider {
 ///
 /// Overriding this is typically done by overriding factories as rest of
 /// `dartdoc` creates this object by calling these static factories.
+// This class has a single direct subclass in this package, [PubPackageMeta],
```
kallentu, line 94: Same for here `///`?

lib/src/package_meta.dart
```diff
@@ -360,13 +329,10 @@ class _FilePackageMeta extends PubPackageMeta {
...
   @override
   List<String> getInvalidReasons() {
-    var reasons = <String>[];
-    if (_pubspec.isEmpty) {
-      reasons.add('no pubspec.yaml found');
-    } else if (!_pubspec.containsKey('name')) {
-      reasons.add('no name found in pubspec.yaml');
-    }
-    return reasons;
+    return [
```
kallentu, line 332: It feels a little weird how this List and its conditional values are regenerated every time we want to check `isValid`. If `_pubspec` is final, could we not have just saved this as a field or something?

kallentu (approved):

lib/src/package_meta.dart
srawlins: It should be easy to compute this value, even if multiple times (I think it's only a few times per package; not like a few times per field-being-documented or something.

We're following ["Avoid storing what you can calculate"](https://dart.dev/effective-dart/usage#avoid-storing-what-you-can-calculate) here. (Not that it would take much storage, one field per package as well.)

What I'd like to do is rename it and make it a getter (so it could be a field (without creating a private backing field)), but that is a breaking change with google3 so... too much work.

kallentu: Fair enough. Thanks for the explanation!

PR #3579, opened 2023-11-13
Add some progress bars for time-intensive tasks

srawlins:
It's a lot of code to get this to display nicely but here goes:

* Refactor logging.html into a class structure to store a few fields; the basic behavior is kept in-tact.
* Privatize `progressLevel`.
* Add top-level functions for setting up, completing, and making progress on a progress bar.
* Remove usage of `cli_util`, which provides it's own Logger class; we were only using it to detect ANSI support, and that can be done inline.
* Add a deterministic progress bar for when elements are being discovered and linked. Also for when docs are being precached. The precaching is necessary so that we can do global discovery of `{@template}`s. 😦 .
* Change `_allModelElements` from a late final field with a self-executing closure to a function. I realized this "field" was only being used once, and it can take minutes and minutes to initialize this "field" when documenting Flutter. 😛 . So... call it a function.
* The ProgressBar class comes from package:nnbd_migration.
* The task code is really awkwardly complex. But I think I added comments to help. I tried to simplify the commands to document a package, and then added a "To re-run this..." output line because all of the pre-work that the task does can take minutes. The awkwardness is mostly from the `--json` flag which is needed when doing things like comparing warnings, but cannot be used when you're trying to watch progress.

---

<details>

</details>

### DIFF
--- a/lib/src/progress_bar.dart
+++ b/lib/src/progress_bar.dart
@@ -3,6 +3,11 @@ import 'dart:math' as math;
 
 import 'package:logging/logging.dart';
 
+/// The logging Level of a progress bar update.
+///
+/// The number, 502, is not interesting; it is just unique, and one greater
+/// than logging.dart's `_progressLevel`. That library performs different
+/// actions when it sees logging messages with this Level.
 const Level progressBarUpdate = Level('PROGRESS_BAR', 502);
 
 /// A facility for drawing a progress bar in the terminal.
--- a/lib/src/progress_bar.dart
+++ b/lib/src/progress_bar.dart
@@ -1,0 +1,88 @@
+import 'dart:io' as io;
+import 'dart:math' as math;
+
+import 'package:logging/logging.dart';
+
+const Level progressBarUpdate = Level('PROGRESS_BAR', 502);
+
+/// A facility for drawing a progress bar in the terminal.
+///
+/// The bar is instantiated with the total number of "ticks" to be completed,
+/// and progress is made by calling [tick]. The bar is drawn across one entire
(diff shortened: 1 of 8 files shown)


tool/task.dart
```diff
@@ -391,42 +391,50 @@ Future<String> docPackage({
...
+    executable,
+    ['pub', 'get'],
+    environment: environment,
+    workingDirectory: pubPackageDir.absolute.path,
+  );
+  await launcher.runStreamed(
+    executable,
+    // Add the `--json` flag for the support in `runStreamed` to tease out
+    // data and messages.
+    [...arguments, '--json'],
```
kallentu, line 423: This is interesting, yeah. What happens when you don't include this flag? I'm having a hard time reasoning with why this is here.

lib/src/progress_bar.dart
```diff
@@ -0,0 +1,88 @@
+import 'dart:io' as io;
+import 'dart:math' as math;
+
+import 'package:logging/logging.dart';
+
+const Level progressBarUpdate = Level('PROGRESS_BAR', 502);
```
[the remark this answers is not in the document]
srawlins: Good catch. Added. The number, is not important; it's just a unique Level for the logging code.

tool/task.dart
srawlins: I added more to the comment. But briefly, we add `--json` for comparison commands, like comparing the dartdoc output (or warnings) for two different dartdoc versions. And those comparison commands require the stdout to be a structured format, for making the comparison.

kallentu (approved):

PR #3583, opened 2023-11-17
Bump version to 8.0.0

kallentu:
I tried my best with the CHANGELOG. Let me know if you'd want any changes!

- [X] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

<details>

</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,30 @@
+## 7.0.3
+
+* Apply @hideConstantImplementations by default. (#3576)
+* Remove unused parameter in zone thing (#3581)
+* Add some progress bars for time-intensive tasks (#3579)
+* Privatize almost all elements in package_builder. (#3580)
+* Privatize several members on PackageGraph (#3578)
+* Correct looping bug in library discovery (#3574)
+* Remove @image directive. (#3568)
+* Privatize `packageName`, `name`, `libraryName`, and
+  `specialFileUri` in `SpecialClass`. (#3569)
+* Privatize `PackageMeta.pathContext`. (#3572)
+* Remove @samples directive. (#3567)
+* Remove @api directive. (#3566)
+* Improve display text for libraries. (#3552)
+* Remove `ElementType.canHaveParameters`. (#3563)
+* Privatize `ElementType`` subclass constructors (#3560)
+* InheritingContainer: `hasPublicInheritedMethods`, `inheritedFields`,
+  `publicInheritedFields`, `publicInheritedMethods` are only visible
+  for testing. (#3559)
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,30 @@
+## 7.0.3
```
srawlins, line 1: Oh dear, I really should write CHANGELOG notes as I commit. Thank you for putting this all together!!

srawlins (approved):

parlough:
Looking at all the API breaking changes, shouldn't this be 8.0.0?

srawlins:
After 7.0.0 I have been treating dartdoc as if it is all internal API; nothing public (lots of "public" API was broken in 7.0.1 and 7.0.2). That's not strictly true as we export plenty from libraries in `lib/` outside `lib/src/`, so I guess step 1 would be to stop exporting from `lib/`. I'd be ok with either. We can call this 7.0.3 or 8.0.0.

parlough:
I don't know how many developers use `package:dartdoc` but I think if it's going to keep being published on pub, its version should match what has changed. It's only fair to any potential users or even explorers of the package. I think they'd much rather have a major version that increases every release than pub solves upgrading them to incompatible dependencies.

If it's planned to discontinue publishing, I recommend making this one final release as `8.0.0`. Then adding to the README details about the discontinuation, publishing 8.0.1, then [discontinuing it properly](https://dart.dev/tools/pub/publishing#discontinue) on pub.

If it's planned to just publish solely as a CLI tool, then make this one final `8.0.0` release and then update to `9.0.0` removing all exported APIs.

srawlins:
I publish for customers to use it via `dart run dartdoc`. Flutter uses this to generate flutter.api.dev, and other customers can use it for a tool that updates with a higher cadence than the Dart SDK stable releases.

parlough:
That makes sense and I know a few other cases of this use too. However that likely was not clear to many users, since it's not really stated anywhere. Also the changelog previously tracked breaking library changes. So how about this:

1. Release this version as `8.0.0`, with a note that the next release will not be intended for use a library.
2. Update the version to `9.0.0`.
3. Clarify in the README this package is intended for use as a CLI tool.
4. Stop exporting most public libraries.
5. Change libraries to our heart's content :)

srawlins:
I think that's a fine plan. We can `@Deprecate()` the exports in the `lib/dartdoc.dart` ASAP as well. (It could be in 8.0.0, but Kallen is trying to get a release out to Flutter, so not high priority.)

srawlins:
8.0.0 LGTM

PR #3605, opened 2023-12-13
Remove padding around inner frame

Hixie:
This leads to weird spacing at the bottom of the page, and generally the docs look better with this padding removed on all sides.


Hixie:
top before:


top after:


bottom before:


bottom after:

srawlins (approved):
❤️  it. Thanks!

Hixie:
(I don't have commit access so someone else will have to land this. :-) )

PR #3610, opened 2023-12-20
Migrate client code to package:web

parlough:


web/search.dart
```diff
@@ -90,56 +87,59 @@ class _Search {
...
+        }
+      }.toJS,
+    );
 
     // Prepare elements.
-    var wrapper = document.createElement('div')..classes.add('tt-wrapper');
+    var wrapper = document.createElement('div')..classList.add('tt-wrapper');
     inputElement
-      ..replaceWith(wrapper)
+      ..replaceWith(wrapper as JSAny) // TODO(parlough): How to without cast?
```
parlough, line 135: @srujzs Is there any better way to use [`replaceWith`](https://pub.dev/documentation/web/latest/web/ElementExtension/replaceWith.html) and its related methods that accept `JSAny nodes`? I'd expect to be able to just specify an `Element` or at least a array/list of `Element`.

parlough:
\cc @kevmoo

I'm not sure if I'm holding the APIs 100% right but most functionality seems to be working fine.

Interestingly, 213KB->205KB despite 1200 extra lines. The interceptor/leaf setup from the old APIs really contributed a lot more than I expected.

web/search.dart
srujzs: This is temporary. :) Since `@staticInterop` types (because they're classes) can't implement extension types, we had to remove the `implements` clauses. Once we release a version where all of these classes move to extension types themselves, the `implements` clauses will go back on and a cast will not be needed.

kevmoo:
@parlough – this is due to dropping a BUNCH of dart:html crazy

parlough:
I figured out why the size didn't drop as much as I expected. These three simple uses of `package:http` increase the final bundle size by ~70KB. I'll rewrite it to just use the web primitives since our uses are simple, but I plan to see if there's any way to reduce how much we bring in within `package:http`. Most of it comes from `package:http_parser` and its dependencies.

kevmoo:
you still cranking here @parlough ?

parlough:
Yeah I'm not sure what I was waiting for...but I'll get this updated tomorrow!

srawlins:
Is this ready @parlough ? Do you have an update on bundle size and use of package:http?

parlough:
Yep, this is ready for review! Sorry about the delay :) 

- I removed usage of `package:http` for now, since our few uses of `fetch` are simple.
- The largest change to pay attention to is changing the sidebar setup from using a `NodeTreeSanitizer` for adjusting the links to a more manual method implemented in `_updateLinks`.
- There's some misc cleanup, such as consolidating the sidebar-related code to one file and using pattern matching where it makes sense.

After these changes the bundle shrinks from 205KB -> 130KB.

srawlins (commented):
This looks great, @parlough ! The translation mostly looks mechanical, not too crazy.

Can you confirm that the following works in manual testing (we still don't have web tests; https://github.com/dart-lang/dartdoc/issues/3027):

1. search, typing, pressing enter to get to the search page, `/` to trigger search, arrow keys?
2. "using base href"; I think there are just two cases to test, using and not using. I think that serving the flutter docs is one state, and serving a package's docs use the other state. (I hate that we still have two modes here, but lots to do.)
3. mobile view; this is the side nav stuff; both the search in the mobile view, and the sidebar loading in the mobile view. I use Chrome's DevTools to emulate mobile.

Thanks again!

kevmoo:
@parlough ?

parlough:
Sorry, I've been traveling the past week :)

Yes, I've tested these three scenarios and a few others. They all worked well :)

kevmoo (approved):

PR #3613, opened 2023-12-29
Simplify web code http requests and use static interop

parlough:
Extracted from https://github.com/dart-lang/dartdoc/pull/3610 for easier review on the road towards modern static JS interop.

`package:http` already uses static interop (through `package:web`), so we shouldn't try to repeat the work from there.


parlough:
I'm confused how changing three calls to use a single method from a new package ends up with a ≈75KB bundle size increase!

I'm going to keep this in draft a bit to investigate. The simplification is worth it, but everyone using `package:http` on the web shouldn't have to pay this cost.

Issue #3562, opened 2023-10-30
issues rolling the latest dartdoc into google3


devoncarew:
We're seeing issues rolling the current dartdoc into google3:

https://fusion2.corp.google.com/invocations/a94742f7-0a21-4518-b126-ad4f8a028292/targets/%2F%2Fdart%2Fdartdoc:dartdoc_e2e_test_run/tests

//dart/dartdoc:dartdoc_e2e_test_run

It looks like some golden test files might need to be regenerated:

```
Expected: '<!DOCTYPE html>\n'...
Expected: '<!DOCTYPE html>\n'
            '<html lang="en">\n'
            '<head>\n'
            '  <meta charset="utf-8">\n'
            '  <meta http-equiv="X-UA-Compatible" content="IE=edge">\n'
            '  <meta name="viewport" content="width=device-width, height=device-height, initial-scale=1, user-
...
```

devoncarew:
Ah, on closer inspection:

```
   Which: is different.
          Expected: ... rtdoc 7.0.1">\n  <me ...
            Actual: ... rtdoc 7.0.2">\n  <me ...
                                  ^
 ```

We may want to make that test not depend on the specific version text.

srawlins:
> We may want to make that test not depend on the specific version text.

+1. I'd prefer to delete those golden tests in favor of unit tests.

devoncarew:
fixed in association with b/321827375

PR #3624, opened 2024-01-19
no longer write the dartdoc version into generated html files

devoncarew:
- no longer write the dartdoc version into generated html files
- fully or partially alleviate https://github.com/dart-lang/dartdoc/issues/3562

---

<details>

</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 8.0.4-wip
 
 * Remove explicit library names. (#3609)
+* No longer write the dartdoc version into generated html files.
 
 ## 8.0.3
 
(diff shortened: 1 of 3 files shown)


kevmoo:
If we're going to mess with this, would it be better to be a bit less cute and a bit more helpful

Either point to the pub.dev page or github repo?

kallentu (approved):
I'm good with these changes, but was there a reason that we were doing this in the first place?

devoncarew:
> If we're going to mess with this, would it be better to be a bit less cute and a bit more helpful Either point to the pub.dev page or github repo?

I think we could just remove this line entirely honestly:

```
<meta name="generator" content="made with love by dartdoc"
```

> I'm good with these changes, but was there a reason that we were doing this in the first place?

I think this is purely historical and not for any particular use case; @srawlins or @keertip may know more.

parlough:
I imagine the original idea was to quickly identify what version of dartdoc a site was generated with in case there are any issues to report or track down.

I think it's probably fine to drop the metadata though. I've never ended up using it :P

srawlins (approved):

devoncarew:
FYI, I updated this PR to remove the 'made with dartdoc' html comment entirely.

Issue #3618, opened 2024-01-12
Enable specifying a Markdown description of a category


parlough:
Then we can add descriptions on the API docs overview/index page: https://api.dart.dev/main

This would allow us to explain what the categories actually mean and why they exist. For example:

- That "Core" libraries should work across all platforms
- That "VM" libraries only work when ran or compiled for native platforms
- That usage of "WEB (LEGACY)" libraries should be migrated (and what to)
- Perhaps more :) 

While a normal string would be sufficient, supporting Markdown here would be great as it would enable easy links.

parlough:
\cc @devoncarew Since you're modifying the categories now, what do you think?

This could show up here on the index page:

<img width="585" alt="New legacy category" src="https://github.com/dart-lang/dartdoc/assets/18372958/484f6cad-4762-473a-a309-f707b448d59a">

devoncarew:
Yeah, I think this would be great - it would give more context to the categories.

We can manually describe things in the markdown above the generated list, but messaging things more than once can be useful.

srawlins:
I think I got mixed up, but this is still not implemented, right? I think @devoncarew landed something separate, which enabled `package:web` to be listed in the Web category of the Dart SDK, right?

PR #3632, opened 2024-01-22
Surface category one liner on index page

parlough:
Surface the first line of category documentation on the documentation index page.

Contributes to https://github.com/dart-lang/dartdoc/issues/3618, by enabling the first line a category/topic's documentation to show up on the index page. If a category does not have a documentation page (such as the Dart SDK's categories), this doesn't change anything.

Example from test_package index page:

Example of surfaced category

### DIFF
--- a/lib/src/generator/templates.aot_renderers_for_html.dart
+++ b/lib/src/generator/templates.aot_renderers_for_html.dart
@@ -1048,6 +1048,13 @@ String renderIndex(PackageTemplateData context0) {
           <h3>''');
       buffer.writeEscaped(context8.name);
       buffer.write('''</h3>''');
+      if (context8.isDocumented) {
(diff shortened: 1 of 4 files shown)


srawlins:
This is great! Can you paste what this does to the Dart SDK docs (which I think use categories for "web", "VM", etc.)?

parlough:
I've updated the PR description, but this is just a first step that could land on its own, as it does benefit existing topics.

As far as I understand, a category with documentation is treated like a topic, but I don't know if we want to do so for the SDK docs. So some mechanism needs to be introduced that prevents a category with documentation being treated that way. I haven't decided yet, so thought this would be a good initial change to land :)

srawlins:
OK, so this change would _not_ add any text to the api.dart.dev home page?

PR #3643, opened 2024-01-29
Add fallback text for sidebar failing to load

parlough:
Adds troubleshooting text/link to the sidebars when it the front-end codes fails to load them. Points to the `dart doc` page which I'll add troubleshooting information to soon.

Also migrates from `package:js` to `dart:js_interop` while standardizing the fetch code for the sidebars. This will make the follow-up transition to `package:web` easier.

Contributes to https://github.com/dart-lang/sdk/issues/54721 

I didn't spend much time making the fallback look nice. Happy to do so if you think it's necessary:

Example of failing to load sidenav


srawlins (approved):
Love it!

srawlins:
Conflicts. Otherwise good to go.

parlough:
Thanks for taking a look!

I'd like to land https://github.com/dart-lang/site-www/pull/5494 first so I can link to a more specific header. I'll update this PR to account for the conflicts and that change :)

srawlins (approved):
Thanks!

PR #3644, opened 2024-01-31
Roll analyzer and release 8.0.5

dcharkes:
Rolling analyzer as required per https://github.com/dart-lang/sdk/issues/54754 so that it can roll into Flutter.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 8.0.5
+
+* Require `analyzer: ^6.4.1`.
+
 ## 8.0.4
 
 * Remove explicit library names. (#3609)
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,5 +1,5 @@
 name: dartdoc
-version: 8.0.4
+version: 8.0.5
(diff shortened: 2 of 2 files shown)


kallentu (approved):

dcharkes:
@kallentu do we have an auto publisher for dartdoc?

kallentu:
> @kallentu do we have an auto publisher for dartdoc?

I don't think so? I typically create a tag manually with the changelog changes and then `dart pub publish` that tag.

dcharkes:
> > @kallentu do we have an auto publisher for dartdoc?
> 
> I don't think so? I typically create a tag manually with the changelog changes and then `dart pub publish` that tag.

👍  Should I go ahead and do that?

kallentu:
> > > @kallentu do we have an auto publisher for dartdoc?
> > 
> > 
> > I don't think so? I typically create a tag manually with the changelog changes and then `dart pub publish` that tag.
> 
> 👍 Should I go ahead and do that?

Yep! That'd be good.
If you need anything or want me to do it, I'm also happy to help.

PR #3649, opened 2024-02-06
Change all late final Iterable fields to Lists

srawlins:
I noticed a `late final Iterable<...>` field and thought to myself, "uh oh! That could be an expensive field!" The idea is that Iterables are lazy, so even if the field is "final", the Iterable calculations are re-executed on each access. So storing a lazily evaluated Iterable on a field seemed very scary 😱 . It may warrant a lint rule.

Luckily, as I walked through the cases I found with grep, most of the values at runtime are Lists, not Iterables. Shwew. Still, I ran through and changed all of the signatures for future-proofing.

Two fields that _were_ Iterables were converted to Lists. I did not benchmark.

---

<details>

</details>

### DIFF
--- a/lib/src/model/library.dart
+++ b/lib/src/model/library.dart
@@ -102,7 +102,7 @@ class Library extends ModelElement
   /// documented with this library, but these ModelElements and names correspond
   /// to the defining library where each originally came from with respect
   /// to inheritance and reexporting.  Most useful for error reporting.
-  late final Iterable<String> allOriginalModelElementNames =
+  late final List<String> allOriginalModelElementNames =
       allModelElements.map((e) {
     if (e is GetterSetterCombo) {
       Accessor? getter;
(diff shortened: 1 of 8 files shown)


lib/src/model/library.dart
```diff
@@ -331,16 +331,16 @@ class Library extends ModelElement
...
-         package.packagePath.contains('/google3/')) {
-       // In google3, `fullName` is specified as if the root of google3 was `/`.
-       // And `package.packagePath` contains the true google3 root.
-       var root = pathContext
-           .joinAll(pathContext.split(package.packagePath)..removeLast());
-       fullName = '$root$fullName';
-     }
-     var relativePath =
-         pathContext.relative(fullName, from: package.packagePath);
+    if (!pathContext.isWithin(fullName, package.packagePath) &&
```
srawlins, line 334: I confess I don't know how the formatter got into this state...

kallentu (approved):

PR #3659, opened 2024-02-16
Replace record keys in maps

jensjoha:
Something weird is happening - for dart JIT - when using records as keys in maps. See also https://github.com/dart-lang/sdk/issues/54928 where I reported it as a bug.

This PR uses 2 classes instead.
The result - for a very small example - is:

```
Difference at 95.0% confidence
        -6286.33 +/- 969.339
        -29.6791% +/- 4.57645%
        (Student's t, pooled s = 427.663)
```

(from ~21 seconds to ~15 seconds)

and trying on flutter the result is

```
Difference at 95.0% confidence
        -46593 +/- 2604.32
        -37.4728% +/- 2.09454%
        (Student's t, pooled s = 1149)
```

(from ~124 seconds to ~78 seconds)

There is ~no difference for an AOT compile.

### DIFF
--- a/lib/src/model/model_element.dart
+++ b/lib/src/model/model_element.dart
@@ -173,8 +173,8 @@ abstract class ModelElement extends Canonicalization
     }
 
     // Return the cached ModelElement if it exists.
-    var cachedModelElement = packageGraph
-        .allConstructedModelElements[(e, library, enclosingContainer)];
+    var cachedModelElement = packageGraph.allConstructedModelElements[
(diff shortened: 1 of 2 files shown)


srawlins:
Thanks for the great analysis, @jensjoha! The gardeners had noticed the `dart doc` test taking a long time as well, and filed https://github.com/dart-lang/sdk/issues/53278, which I have not looked at in a while. I'll land this change and watch that bot.

srawlins (approved):
Very straightforward, thanks for the fix!

PR #3660, opened 2024-02-16
Make a few optimizations

jensjoha:
This:

  * Make 'allFields' a late field instead
  * Cache _inheritedElements in allModelElements call
  * Make documentedWhere a late field

Measusred on top of (and compared to) https://github.com/dart-lang/dartdoc/pull/3659 (I couldn't figure out how to actually do a stacked thing on github):

Very small, JIT:
```
Difference at 95.0% confidence
        -2052.67 +/- 417.162
        -13.7812% +/- 2.80075%
        (Student's t, pooled s = 184.048)
```

Very small, AOT:
```
Difference at 95.0% confidence
        -1370.33 +/- 214.258
        -20.2832% +/- 3.17138%
        (Student's t, pooled s = 94.5287)
```

Flutter, JIT:
```
Difference at 95.0% confidence
        -5111.33 +/- 2168.41
        -6.57446% +/- 2.78912%
        (Student's t, pooled s = 956.682)
```

Flutter, AOT:
```
Difference at 95.0% confidence
        -3224.33 +/- 599.597
        -5.33441% +/- 0.991988%
        (Student's t, pooled s = 264.537)
```

So it seems better across the board, but comparably less so for a big application (flutter in this case) --- which makes sense because most of the time is spent in the "generator.generate" (name from the "Runtime performance" when running with -v) where these changes is optimizing the "initializePackageGraph" part...

### DIFF
--- a/lib/src/model/inheriting_container.dart
+++ b/lib/src/model/inheriting_container.dart
@@ -104,10 +104,15 @@ abstract class InheritingContainer extends Container
               packageGraph.rendererFactory.languageFeatureRenderer)
           .toList();
 
-  late final List<ModelElement> _allModelElements = [
-    ...super.allModelElements,
-    ...typeParameters,
-  ];
+  late final List<ModelElement> _allModelElements = () {
+    _inheritedElementsCache = _inheritedElements;
+    var result = [
+      ...super.allModelElements,
+      ...typeParameters,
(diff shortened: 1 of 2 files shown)


srawlins (approved):
Thanks these look fine!

PR #3664, opened 2024-02-20
Remove unnecessary Builder classes

srawlins:
Big ball of yarn that grew and grew.

The main idea is to remove a bunch of classes, whose necessity I never understood: `ElementTypeBuilder` (and `Impl`), `ModelElementBuilder` (and `Impl`), `ModelObjectBuilder`, `ModelBuilder`, and `ModelBuilderInterface`. These might have had more of a purpose when we could output to HTML or Markdown, but their complexity is now unwarranted.

The purpose they served is this: In order to fetch a ModelElement or ElementType from from the package graph, you always needed an instance of... the package graph. Putting a `*Builder*` in a class you're working in gave you a way to instantiate (or fetch from the package graph) a ModelElement or ElementType.

But, lo and behold, we can declare that a `Nameable` must have a `PackageGraph`, and then write little utility functions to fetch these things. Much simpler. Here are the other changes needed to facilitate this:

* I renamed the various factory constructors. I generally changed the word 'from' to 'for', because often the value that the factory constructor returns is a cached instance. It feels weird to me to say `ModelElement.from` some Element, when you know that you already have the instance, and you're just fetching it, using the Element as a key. In that case, I think 'for' is more appropriate.

  * `ModelElement._from` --> `ModelElement.for_`
  * `ModelElement._fromElement` --> `ModelElement.forElement`
  * `ModelElement._fromPropertyInducingElement` --> `ModelElement.forPropertyInducingElement`
  * `ElementType._from` --> `ElementType.for_`
  * These did have to be made public, which is unfortunate; I think there are some gymnastics we can do to re-privatize them, but it would more-or-less mean moving `Nameable`, or a `Nameable` supertype... I think not worth it.

* Make Nameable a proper mixin. 👍 
* Move `_resolveMultiplyInheritedElement` to be an extension method, to the bottom of the library.

I believe this may also open up nicer refactorings where some `late final` `ElementType` fields do not need to be late. It might not be a hole in one, but once we're only relying on `packageGraph`, we may be able to assign some more instance fields at construction-time.

---

<details>

</details>


lib/src/model/nameable.dart
```diff
@@ -29,8 +37,49 @@ abstract mixin class Nameable {
...
+  }) =>
+      ModelElement.for_(
+        element,
+        library,
+        packageGraph,
+        enclosingContainer: enclosingContainer,
+      );
+
+  /// Returns the [ModelElement] for [element], instantiating it if needed.
+  ModelElement getModelForElement(Element element) =>
```
kallentu, line 57: What's the difference between `getModelForElement` and `getModelFor`? 

Additionally, could we not bundle this into the same helper with an optional field for `library`? So there's less choice in which to call. They all seem very similar.

lib/src/model/model_element.dart
```diff
@@ -237,7 +189,7 @@ abstract class ModelElement extends Canonicalization
   // TODO(jcollins-g): Allow e to be null and drop extraneous null checks.
   // TODO(jcollins-g): Auto-vivify element's defining library for library
   // parameter when given a null.
-  factory ModelElement._from(
+  factory ModelElement.for_(
```
kallentu, line 192: Also, what is the ending `_` in the name for? Why not just `ModelElement.for`?

lib/src/model/constructor.dart
```diff
@@ -33,7 +33,7 @@ class Constructor extends ModelElement
 
   @override
   Container get enclosingElement =>
-      modelBuilder.from(element.enclosingElement, library) as Container;
+      packageGraph.getModelFor(element.enclosingElement, library) as Container;
```
kallentu, line 36: This and other files -- why do we sometimes call `getModelFor` or `getModelForPropertyInducingElement` with `packageGraph` and other times we don't? Can we just call the helper itself (`getModelFor(element.enclosingElement, library) as Container`)?

kallentu (commented):
Thank you for doing this. I saw the original ModelBuilder mixins and impls and I wanted to fix it up, so I'm glad you did.

lib/src/model/model_element.dart
srawlins: This is because `for` is a reserved keyword. 🤷 

I'm open to suggestions, but I would like to stay away from `from`, because of the reasons in the summary. `of` is not great either, I think.

lib/src/model/nameable.dart
srawlins: Asking the hard questions!

These three always trip me up, and I've just added a bit more documentation above `ModelElement.forElement`. I could add more documentation somewhere, but here's the story as I understand it:

* The most "naive" "generous" factory constructor is `forElement` (and this convenience method, `getModelForElement`. You can pass it an Element that is property-inducing, or not, and you don't even need to calculate the Library first.
* That factory constructor (1) determines the Library first, then (2) determines whether the element is property-inducing, and forks on that decision, calling either `for_` or `forPropertyInducingElement`.
* Now that we have two "delegatee" factory constructors, `for_` and `forPropertyInducingElement`, there are some callers that would prefer to call _them_, over `forElement`. Namely if you want to specify a specific enclosingContainer, or specific `getter`/`setter`.

Now, a lot of this is assumption: looking at the callers of `modelForPropertyInducingElement`, it looks like most callers calculate a `getter` and a `setter` in the same way that `forElement` does, and maybe that can all be collapsed. It would be great if I could eliminate all of the library-external calls to `modelForPropertyInducingElement`/`ModelElement.forPropertyInducingElement`.

It's also possible that we can strip the `library` parameter to these methods, if it is always the same object that would be determined by `ModelElement.forElement`. `ModelElement.forElement` uses a simple lookup. But maybe some callers need to construct a Library...

Also open to better names for these instead of relying wholly on doc comments to differentiate. ☹️

lib/src/model/constructor.dart
srawlins: Thanks for pointing this out!

This is a relic from halfway through this refactoring, before I moved the convenience methods to `Nameable`. I've just pushed a commit that simplifies a dozen call sites.

lib/src/model/model_element.dart
kallentu: Ah, I forgot about that. `for_` is fine to me.

lib/src/model/nameable.dart
kallentu: > It would be great if I could eliminate all of the library-external calls to modelForPropertyInducingElement/ModelElement.forPropertyInducingElement.

Agreed.

Thanks for the documentation! Honestly, I think it's fine for now. Naming is hard.

kallentu (approved):

Issue #3480, opened 2023-08-24
Remove support for custom HTML templates


srawlins:
We wish to remove support for custom HTML templates (`--templates-dir`) option. This option was primarily added for one customer, and they no longer need it.

First step is to deprecate it and see what comes out of the woodwork.

srawlins:
We've deprecated and removed this in a few pieces. I think the final nail in the coffin was https://github.com/dart-lang/dartdoc/commit/efff7c1f3e856b10fc2602aa4627b88f9167a955.

PR #3667, opened 2024-02-21
Deprecate `--templates-dir` option.

kallentu:
Related to https://github.com/dart-lang/dartdoc/issues/3480.

Deprecate the option. (Which I'm actually not sure what more there is to do besides writing it in the command help. Let me know.)
Added an entry to the changelog for this change.

---

- [ ] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

<details>

</details>


srawlins:
In `documentation_comment.dart`, you can see how we `warn(` if the `examplePathPrefix` value is non-null. The same can be done in `templates.dart`, when `templatesDir` is non-null.

kallentu:
> In `documentation_comment.dart`, you can see how we `warn(` if the `examplePathPrefix` value is non-null. The same can be done in `templates.dart`, when `templatesDir` is non-null.

Oh, happy to do that!
I looked around for a bit and couldn't find a `Warnable` that I could "warn" on. I think at that point in `templates.dart`, we haven't even built the package graph. Hmm, am I missing something?

srawlins:
Ah good catch. No I can't find anything easy to hang the warning on there. But the RuntimeTemplates there is used in a GeneratorBackend, so we could just delay the warning until it is used by the GeneratorFrontEnd (\*grumble* inconsistent camel casing 😠), and trigger it based on `is RuntimeTemplates`, which is a little hacky.

A good spot might be at the beginning of `GeneratorFrontEnd._generateDocs`, something like `if ((_generatorBackend as GeneratorBackendBase).templates is RuntimeTemplates) { packageGraph.defaultPackage.warn(...); }`.

I don't like that cast, so maybe `GeneratorFrontEnd._generatorBackend` can be made a `GeneratorBackendBase`, or maybe one of the types in the hierarchy can be removed. A spoonful of tech debt helps the simple feature deprecation go down. 😁

kallentu:
Added the warning for this, PTAL again :)

srawlins (approved):

PR #3669, opened 2024-02-22
Fix enum representationsto show correct (implicit) args, const

srawlins:
When adding support for enhanced enums, I think I overlooked some validation of "old scool" enums-with-implicit-constructors. I discovered this when trying to move all tests to use Dart Language version 3.2.0 (next PR).

Unfortunately, the analyzer's `constantInitializer` property for such enum values is not really meant for re-constructing what a constant's source was, and it returns code like `E()` for the constant initializer for `E.one` given `enum E { one, two; }`. (It should implicitly use an index of 0, and show `E(0)`; see the [de-sugaring examples in the spec](https://github.com/dart-lang/language/blob/main/accepted/2.17/enhanced-enums/feature-specification.md#examples).) I filed https://github.com/dart-lang/sdk/issues/54988.

In this code, we also add the word `const` to enhanced enum values (with an explicit constructor); previously, only old-school enums had the keyword `const`. So now that's unified.

---

<details>

</details>

### DIFF
--- a/lib/src/model/getter_setter_combo.dart
+++ b/lib/src/model/getter_setter_combo.dart
@@ -133,7 +133,8 @@ mixin GetterSetterCombo on ModelElement {
       // For an enum value with an implicit constructor call (like
       // `enum E { one, two; }`), `constantInitializer.toString()` does not
       // include the implicit enum index argument (it is something like
-      // `const E()`). We must manually include it.
+      // `const E()`). We must manually include it. See
+      // https://github.com/dart-lang/sdk/issues/54988.
       initializerString =
           initializerString.replaceFirst('()', '(${self.index})');
     }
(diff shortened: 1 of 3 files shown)


lib/src/model/getter_setter_combo.dart
```diff
@@ -110,8 +110,37 @@ mixin GetterSetterCombo on ModelElement {
...
+    initializerString = 'const $initializerString';
+
+    var isImplicitConstructorCall = constantInitializer
+            .constructorName.staticElement?.isDefaultConstructor ??
+        false;
+    if (isImplicitConstructorCall) {
+      // For an enum value with an implicit constructor call (like
+      // `enum E { one, two; }`), `constantInitializer.toString()` does not
+      // include the implicit enum index argument (it is something like
+      // `const E()`). We must manually include it.
```
kallentu, line 136: Perhaps link the issue here and revisit this if it's fixed in the future.

kallentu (approved):

lib/src/model/getter_setter_combo.dart
srawlins: Done.

Issue #3452, opened 2023-06-22
First class functions are missing generic type parameters in API


nshahan:
I noticed this when looking at the docs for package:dart_internal. For example look at [extractIterableTypeArgument](https://pub.dev/documentation/dart_internal/latest/extract_type_arguments/extractIterableTypeArgument.html)

Notice how in the signature at the top the function parameter is displayed in the old format and is missing its generic type parameter:
```dart
extractIterableTypeArgument(Iterable iterable, Object? extract()) → Object?
```

While in the implementation you can see the newer function type syntax and the type parameter:
```dart
Object? extractIterableTypeArgument(
        Iterable iterable, Object? Function<T>() extract) =>
    internal.extractTypeArguments<Iterable>(iterable, extract);
```

This issue appears on the main API page as well:
https://pub.dev/documentation/dart_internal/latest/extract_type_arguments/extract_type_arguments-library.html

srawlins:
Whoa yikes, great catch.

PR #3670, opened 2024-02-22
Render generic type parameters in functions.

kallentu:
Added generic type parameters to function parameters, fixing https://github.com/dart-lang/dartdoc/issues/3452.

For that issue's example, it'd show as the following now.
```
// Before: extractIterableTypeArgument(Iterable iterable, Object? extract()) → Object?
extractIterableTypeArgument(Iterable iterable, Object? extract<T>()) → Object?
``` 

@srawlins 
Wondering if we should revisit the function format and make it `Object? Function<T>() extract` instead of `Object? extract<T>()`? This CL isn't for that, but it wouldn't be too hard to make that change.

- Renamed some tests in `parameter_test.dart` because I was itching to do so.
- The two new tests added are `test_formalParameter_generic_method` and `test_formalParameter_generic_topLevelFunction`.

---

- [X] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

<details>

</details>

### DIFF
--- a/lib/src/render/parameter_renderer.dart
+++ b/lib/src/render/parameter_renderer.dart
@@ -164,6 +164,22 @@ abstract class ParameterRenderer {
           : modelType.returnType.linkedName;
       buffer.write(typeName(returnTypeName));
       buffer.write(' ${parameterName(param.name)}');
+
+      // Writes out the generic type parameters for a function type.
+      // TODO(kallentu): Pull this type parameter generation into a helper for
+      // other renderers that also do this same work.
+      if (modelType is FunctionTypeElementType) {
+        if (modelType.typeFormals.isNotEmpty) {
+          if (!modelType.typeFormals.every((t) => t.name == 'dynamic')) {
+            buffer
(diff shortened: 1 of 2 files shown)


srawlins (approved):
Very nice! Nice test refactoring.

srawlins:
> Wondering if we should revisit the function format and make it `Object? Function<T>()` extract instead of `Object? extract<T>()`?

I think I would be in favor of this. I can see supporting one of two designs:

* Always display the new format.
* Always display what is in the source code; if they wrote the old school format, support that. There are plenty of cases in the Dart SDK at least that use the old school format, like [Future.then](https://api.dart.dev/stable/3.3.0/dart-async/Future/then.html). I'm not sure if it's intentional, but if it is 🤷  maybe leave it that way in dartdoc too.

kallentu:
I'll look into it in my free-time, but if we could, I think maybe it'd be good to display what's in the source code.

I'll make an issue for this.

PR #3688, opened 2024-02-27
Do not require FLUTTER_ROOT any longer, even for flutter packages

srawlins:
I saw this TODO from @devoncarew, and tried it myself: I can run `dart doc`, with a `dart` script from the flutter/flutter repo, and no `FLUTTER_ROOT` env variable is necessary. Swoosh!

Also delete the test for this exception. And then delete the `testing/` package, which was only used by that last test.

---

<details>

</details>


kallentu (approved):
LGTM.

Could we remove the `flutterRoot` option? Is that still needed?

srawlins:
> Could we remove the `flutterRoot` option? Is that still needed?

I think the `flutterRoot` option (which is typically set via `FLUTTER_ROOT`, in `dartdoc_options.dart`, can still serve a purpose if you are manually overriding the default value. I don't know if there is a reason to do this. We could deprecate it (since `FLUTTER_ROOT` is no longer _required_), and see if it causes any problems.

PR #3689, opened 2024-02-28
Simplify InheritingContainer.allModelElements and _inheritedElements

srawlins:
I could _not_ wrap my head around the interactions between `InheritingContainer._allModelElements`, which seemed to fill a cache, then do work, then clear the cache, and `InheritingContainer._inheritedElements`. `_inheritedElements` and `_inheritedElementsCache` seemed like a clear instance of a `late final` field before null safety (before `late`), but then the cache invalidation in `_allModelElements` kind of nullifies that theory...

Anyways, it can all be simpler:

* `_allModelElements` can just be the list.
* `_inheritedElementsCache` is no more; `_inheritedElements` is now a late final.
* I cannot fathom why `inheritanceChainElements` was initialized in the for loop... so its now initialized outside the for loop. 🤷 
* Iterating over `someMap.keys` and immediately calling back into the map and null-asserting the result is too much work. We can instead iterate over the entries (hopefully performant...) and destructure the result.
* Add more comments for the next archaeologist.

---

<details>

</details>

### DIFF
--- a/lib/src/model/inheriting_container.dart
+++ b/lib/src/model/inheriting_container.dart
@@ -103,15 +103,10 @@ abstract class InheritingContainer extends Container
           .asLanguageFeatureSet(const LanguageFeatureRendererHtml())
           .toList();
 
-  late final List<ModelElement> _allModelElements = () {
-    _inheritedElementsCache = _inheritedElements;
-    var result = [
-      ...super.allModelElements,
-      ...typeParameters,
-    ];
-    _inheritedElementsCache = null;
-    return result;
-  }();
+  late final List<ModelElement> _allModelElements = [
+    ...super.allModelElements,
+    ...typeParameters,
+  ];
 
   Iterable<Method> get inheritedMethods {
     var methodNames = declaredMethods.map((m) => m.element.name).toSet();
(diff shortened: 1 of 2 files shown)


lib/src/model/inheriting_container.dart
```diff
@@ -148,55 +143,63 @@ abstract class InheritingContainer extends Container
...
-    if (_inheritedElementsCache != null) return _inheritedElementsCache!;
+  /// A list of the inherited executable elements, one element per inherited
+  /// `Name`.
+  ///
+  /// In this list, elements that are "closer" in the inheritance chain to
+  /// _this_ element are preferred over elements that are further away. In the
+  /// case of ties, concrete inherited elements are prefered to non-concrete
+  /// ones.
+  late final List<ExecutableElement> _inheritedElements = () {
     if (element is ClassElement && (element as ClassElement).isDartCoreObject) {
```
kallentu, line 154:
```
    if (element case ClassElement classElement
        when classElement.isDartCoreObject) {
      return const <ExecutableElement>[];
    }
```
Maybe? Instead of casting.

lib/src/model/inheriting_container.dart
```diff
@@ -148,55 +143,63 @@ abstract class InheritingContainer extends Container
...
+          inheritedElement.enclosingElement as InterfaceElement;
       assert(inheritanceChainElements.contains(enclosingElement) ||
           enclosingElement.isDartCoreObject);
 
-      // If the concrete object from
-      // [InheritanceManager3.getInheritedConcreteMap2] is farther from this
-      // class in the inheritance chain than the one provided by
+      // If the concrete object from `getInheritedConcreteMap2` is farther in
+      // the inheritance chain from this class than the one provided by
       // `inheritedMap2`, prefer `inheritedMap2`. This correctly accounts for
```
kallentu, line 189: Which is `inheritedMap2`? Is this comment outdated or am I missing something?
Or does this just mean `inheritanceMap` and "the one provided" == `inheritedElement`?
Maybe fix the comments to be a tad more clear.

lib/src/model/inheriting_container.dart
```diff
@@ -148,55 +143,63 @@ abstract class InheritingContainer extends Container
...
-    final concreteInheritenceMap =
+    var concreteInheritanceMap =
         packageGraph.inheritanceManager.getInheritedConcreteMap2(element);
-    final inheritenceMap =
+    var inheritanceMap =
         packageGraph.inheritanceManager.getInheritedMap2(element);
 
-    List<InterfaceElement>? inheritanceChainElements;
+    var inheritanceChainElements =
+        inheritanceChain.map((c) => c.element).toList(growable: false);
```
kallentu, line 164: Also these three maps above are named so similarly. Can we add a comment on what the difference is between them?

kallentu (approved):

lib/src/model/inheriting_container.dart
srawlins: I like it

lib/src/model/inheriting_container.dart
srawlins: Ugh, the doc comment on `inheritanceChain` only describes what it is _not_ ("Not the same as [superChain] as it may include mixins."). OK so I've updated _that_ comment. And some more. And then introduced some comments above these declarations.

I cannot think of better names; they're already long and I think making them longer would hurt readability. But I'm open to new name ideas.

FWIW I don't know why this algorithm was chosen for `_inheritedElements`, but I'm strongly preferring to leave it alone; it is likely depended on by Flutter. It might be documented in tests.

lib/src/model/inheriting_container.dart
srawlins: Oops, wrong name. I am referring to the `getInheritedMap2` method. Updated.

PR #3695, opened 2024-02-29
In PackageBuilder._discoverLibraries, initialize newFiles outside loop.

srawlins:
I noticed with print debugging that `addFilesReferencedBy` was being called way too many times for a given library. It turns out this is because it was being called on different Sets, when really it should mostly be operating on one Set. That one Set is `newFiles`. It was initialized in the do-while loop, but it can be initialized before the do-while loop.

I checked the logic to make sure this change is a no-op.

I also timed documenting flutter. This change seems to reduce the time-to-document by ~100 seconds, but that is only with one before and one after, so I promise nothing.

- Thanks for your contribution! Please replace this text with a description of what this PR is changing or adding and why, list any relevant issues, and review the contribution guidelines below.

---

<details>

</details>

### DIFF
--- a/lib/src/model/package_builder.dart
+++ b/lib/src/model/package_builder.dart
@@ -229,10 +229,10 @@ class PubPackageBuilder implements PackageBuilder {
     // a set of files (starting with the ones passed into the function), resolve
     // them, add them to the package graph via `addLibrary`, and then discover
     // which additional files need to be processed in the next loop. This
-    // discovery depends on various options (TODO: which?), but the basic idea
-    // is to take a file we've just processed, and add all of the files which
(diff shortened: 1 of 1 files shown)


kallentu (approved):

kallentu:
> I also timed documenting flutter. This change seems to reduce the time-to-document by ~100 seconds, but that is only with one before and one after, so I promise nothing.

If this is truly the case, I'm shocked.

srawlins:
> If this is truly the case, I'm shocked.

Haha, I promise nothing! 100 seconds probably isn't within the margin of error, but it could be a much smaller improvement on average.

Issue #3584, opened 2023-11-17
Documentation for functions in `enum`s are linked but not generated


mkorbel1:
If you define an `enum` with a function in it, then the documentation page for that `enum` will have a link to a documentation page about that function (and even the doc comment for the function), however the page for the function will be missing.

In addition, since the target of the link is missing, `--validate-links` will trigger a warning.  This was annoying to debug because of #3331.  I had to run this repo in debug mode and add `print` statements to figure out what was missing since I didn't know what the warning was.

To reproduce, create an `enum` with a `static` function in it, for example:
```dart
enum MyEnum {
  a,
  b,
  c;

  /// Says hello!
  static void myFunction() {
    print("hello");
  }
}
```

The generated documentation will include `MyEnum` and a link to `myFunction`, as expected, however the link (circled) will be dead because the doc page for `myFunction` is not generated.

![image](https://github.com/dart-lang/dartdoc/assets/75506197/7bb46f95-773f-49ca-9b72-05b01f518ce4)

srawlins:
Sorry for the extreme delay (dartdoc has fallen through the gaps in our triage process). I think there is a sizeable blind spot here with enhanced enums. I noticed that dartdoc's Enum doesn't inherit from Class, but it really probably should.

PR #3697, opened 2024-03-01
Generate docs for enum static methods.

srawlins:
Fixes https://github.com/dart-lang/dartdoc/issues/3584

The big function in GeneratorFrontEnd was _really_ big, and I think that made it easy to omit things. I've refactored it to reuse more code. Also, the repeated code had bugs:

* For extensions and extension types, we likely _over_-documented, because we didn't omit things that weren't canonical.
* For enums, we omitted static methods and static fields (#3584).

Also in this change:

* Change Enum to extend Class; while the analyzer's EnumElement does _not_ extend `ClassElement`, behaviorally, it's too good to pass up. I think this also fixes a bug linkifying enum values, as seen in the enum tests.
* Migrate methods_test.dart to use test_reflective_loader.
* Add method tests for methods on classes, enums, mixins, extensions, and extension types.
* Change some `clazz` and `eNum` parameter names to `class_` and `enum_`. More standard, and standard inside the analyzer.
* In the shared testing code, add some code that prints out the in-memory files when a file cannot be read. Here's a snippet of what that looks like:

```none
/temp/method_test/doc/
   ├─ index.html
   ├─ __404error.html
   ├─ search.html
   ├─ lib/
   │  ├─ lib-library.html
   │  ├─ lib-library-sidebar.html
   │  ├─ E-class.html
   │  ├─ E-enum-sidebar.html
   │  └─ E/
   │  │  ├─ values-constant.html
```

---

<details>

</details>

### DIFF
--- a/lib/src/model/container.dart
+++ b/lib/src/model/container.dart
@@ -153,8 +153,16 @@ abstract class Container extends ModelElement
   late final List<Field> publicConstantFieldsSorted =
       publicConstantFields.toList(growable: false)..sort(byName);
 
+  /// The total list of public enum values.
+  ///
+  /// This is defined on [Container] instead of just [Enum], because the
+  /// `ContainerSidebar` Mustache template needs to refer to this field.
   Iterable<Field> get publicEnumValues => const [];
 
+  /// Whether this container has any public enum values.
+  ///
+  /// This is defined on [Container] instead of just [Enum], because the
+  /// `ContainerSidebar` Mustache template needs to refer to this field.
   bool get hasPublicEnumValues => publicEnumValues.isNotEmpty;
 
   Iterable<Accessor> get instanceAccessors =>
--- a/lib/src/model/enum.dart
+++ b/lib/src/model/enum.dart
@@ -8,31 +8,8 @@ import 'package:dartdoc/src/model/model.dart';
 import 'package:dartdoc/src/model_utils.dart' as model_utils;
 import 'package:dartdoc/src/render/enum_field_renderer.dart';
 
-/// The [Enum] class only inherits from [InheritingContainer] because declared
-/// `enum`s inherit methods from [Object].  It can't actually participate
-/// meaningfully in other inheritance or have class modifiers.
-class Enum extends InheritingContainer
-    with Constructable, TypeImplementing, MixedInTypes {
-  @override
-  final EnumElement element;
-
-  Enum(this.element, super.library, super.packageGraph);
-
-  @override
-  late final List<ModelElement> allModelElements = [
-    ...super.allModelElements,
-    ...constructors,
-  ];
-
-  @override
-  late final List<InheritingContainer> inheritanceChain = [
-    this,
-    for (var container in mixedInElements.reversed)
-      ...container.inheritanceChain,
-    for (var container in superChain.modelElements)
-      ...container.inheritanceChain,
-    ...interfaceElements.expandInheritanceChain,
-  ];
+class Enum extends Class {
+  Enum(super.element, super.library, super.packageGraph);
 
   @override
   String get sidebarPath => '${library.dirName}/$name-enum-sidebar.html';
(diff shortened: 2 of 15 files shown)


lib/src/model/enum.dart
```diff
@@ -8,31 +8,8 @@ import 'package:dartdoc/src/model/model.dart';
...
-  @override
-  late final List<InheritingContainer> inheritanceChain = [
-    this,
-    for (var container in mixedInElements.reversed)
-      ...container.inheritanceChain,
-    for (var container in superChain.modelElements)
-      ...container.inheritanceChain,
-    ...interfaceElements.expandInheritanceChain,
-  ];
+class Enum extends Class {
```
kallentu, line 11: I know there's shared behaviour that we want, but this feels a little weird semantically. Could we rename Class to Interface or something similar? I feel like we're making the hierarchy even more confusing.

lib/src/model/container.dart
```diff
@@ -153,8 +153,16 @@ abstract class Container extends ModelElement
   late final List<Field> publicConstantFieldsSorted =
       publicConstantFields.toList(growable: false)..sort(byName);
 
+  /// The total list of public enum values.
+  ///
+  /// This is defined on [Container] instead of just [Enum], because the
+  /// `ContainerSidebar` Mustache template needs to refer to this field.
```
kallentu, line 159: I think this is just the side effect of using Mustache templates, but this doesn't feel great. I wouldn't choose to have `publicEnumValues` on the superclass of enum. 

But alas.

kallentu (approved):

lib/src/model/enum.dart
srawlins: I could back out this change, if it makes the hierarchy more confusing. I wonder if a cleaner hierarchy would be:

* Interface, which maybe aligns with analyzer's InterfaceElement
* Enum extends Interface, aligning with analyzer's EnumElement
* Class extends Interface, aligning with analyzer's ClassElement
* Mixin extends Interface, aligning with analyzer's MixinElement

I'll look into this.

lib/src/model/container.dart
srawlins: Oh yeah. A bummer. There are a few of these. More than being an issue with Mustache templates, it is an issue specifically with our implementation. I haven't looked at this stuff in years, and it might be ripe for fixing. :D

lib/src/model/enum.dart
srawlins: OK so that already existed, it was InheritingContainer 😛  So I just backed out this part of the change, and we'll leave the inheritance like it is. It only reduced some duplication, but at the expense of a correct hierarchy is bad.

lib/src/model/container.dart
srawlins: I looked a bit and remembered. A bit of history 😁 

When I initially wrote Mustachio, our mirrorless Mustache renderer, I envisioned it might be extracted into its own package some day. But I wanted to bake it in dartdoc first. That was a good decision because a few of the little designs are optimized for dartdoc. We have never opened it as a stand-alone Mustache renderer, and that's fine. One less headache 😬 .

It also kind of opens some doors. We can add some custom syntax on Mustache. There are two I've wanted:

* A boolean block based on static typing. `{{ #foo is Foo }}`. There are a handful of places where it would come in handy, and we wouldn't need as much shoe-horning like this.
* An awaited Future value. `{{ await foo }}` maybe. In particular, one reason we have to pre-process all doc comments is so that they are waiting patiently on each ModelElement by the time Mustachio comes by to read them. Mustachio cannot currently ask an object for a value which is Future-typed.

lib/src/model/enum.dart
kallentu: Sounds good to me, thanks.
I was just afraid it would become another situation where we're like "wait what does this mean?"

PR #3730, opened 2024-03-19
mustachio: Separate out the context stack LUB type calculation

srawlins:
I have a bigger change I've had lying around for generated renderer deduplication, and I've teased out this smaller change.

It's a no-op, and I think I've improved the documentation. I need to write docs for the tool that generates the renderers, but I hope this PR is small enough to understand, and how it is a no-op. [Here are some other docs](https://github.com/dart-lang/dartdoc/blob/main/tool/mustachio/README.md#generated-renderer-for-a-specific-type-and-a-pre-compiled-static-template) for these renderers.

---

<details>

</details>


tool/mustachio/codegen_aot_compiler.dart
```diff
@@ -116,27 +119,23 @@ Future<Map<_AotCompiler, String>> _deduplicateRenderers(
       continue;
     }
     var firstCompiler = compilers.first;
```
kallentu, line 121: This part of the codegen is kinda confusing. Why do we use the first compiler's `_syntaxTree` and `_buildData` when making the `lubCompiler`? Also, perhaps move this variable down there, since it's not being used for `contextStacksLength` anymore

All the other changes seem okay to me.

kallentu (approved):

tool/mustachio/codegen_aot_compiler.dart
kallentu: I see! Okay, that makes a lot of sense. Yeah, some extra documentation on that would be great :)

PR #3743, opened 2024-04-05
Deprecate the `missingCodeBlockLanguage` warning.

kallentu:
Deprecates the `missingCodeBlockLanguage` warning. This will be replaced by the `missing_code_block_language_in_doc_comment` lint.

This is a part of the effort to move more Dartdoc logic to the analyzer/linter.

See lint proposal here: https://github.com/dart-lang/linter/issues/4904

---

<details>

</details>

### DIFF
--- a/lib/src/model/documentation_comment.dart
+++ b/lib/src/model/documentation_comment.dart
@@ -1,3 +1,7 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
 import 'dart:io';
 
 import 'package:analyzer/dart/element/element.dart';
--- a/lib/src/model/documentation_comment.dart
+++ b/lib/src/model/documentation_comment.dart
@@ -1,3 +1,5 @@
+import 'dart:io';
+
 import 'package:analyzer/dart/element/element.dart';
 import 'package:args/args.dart';
 import 'package:crypto/crypto.dart' as crypto;
@@ -676,6 +678,8 @@ mixin DocumentationComment
     for (var element in firstOfPair) {
       final result = element.group(2)!.trim();
       if (result.isEmpty) {
+        stderr.writeln("The warning 'missingCodeBlockLanguage' is deprecated. "
+            'Use the `missing_code_block_language_in_doc_comment` lint instead.');
         warn(PackageWarning.missingCodeBlockLanguage,
             message:
                 'A fenced code block in Markdown should have a language specified');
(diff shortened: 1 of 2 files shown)


kallentu:
@srawlins Is there anything else we need to mark for warning deprecations?

srawlins:
Hmm it would be good to report the warning as deprecated, if enabled. It looks like that might be [here](https://github.com/dart-lang/dartdoc/blob/main/lib/src/warnings.dart#L402) (for enabling as an error) and [here](https://github.com/dart-lang/dartdoc/blob/main/lib/src/warnings.dart#L410) (for enabling as a warning).

If you add a `_isDeprecated` field to `PackageWarning`, then it should be pretty easy to check.

However, what does "report" mean? In our advanced warnings system, this is the code that sets up whether various checks are "warning" or "ignore" or "error", so it seems like you'd run into problems quick if you tried to report a warning about enabling a warning while setting up warnings. 😟  So... maybe you can just print to stderr or stdout and call it a day. 😀

kallentu:
Yeah it seemed odd to report on a warning report.
Anyways, added a stderr, feel free to take a look

lib/src/model/documentation_comment.dart
```diff
@@ -676,6 +682,8 @@ mixin DocumentationComment
     for (var element in firstOfPair) {
       final result = element.group(2)!.trim();
       if (result.isEmpty) {
+        stderr.writeln("The warning 'missingCodeBlockLanguage' is deprecated. "
```
srawlins, line 685: Hmm, but this will only alert a developer to the deprecation if a new `missingCodeBlockLanguage` violation is found.

I think we need to alert (print to stderr) any time `dart doc` is invoked, with the missingCodeBlock warning enabled. Then for a codebase like Flutter (though we're holding their hand) will be notified via their CI that they are enabling a deprecated option.

kallentu: Oh you're right. Let me redo this. One sec.

lib/src/model/documentation_comment.dart
```diff
@@ -1,3 +1,7 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
```
srawlins, line 1: Cringe, thanks for that!

lib/src/warnings.dart
```diff
@@ -455,14 +464,21 @@ class PackageWarningOptions {
...
       warningModes[kind] = PackageWarningMode.ignore;
 
   void warn(PackageWarning kind) =>
       warningModes[kind] = PackageWarningMode.warn;
 
-  void error(PackageWarning kind) =>
-      warningModes[kind] = PackageWarningMode.error;
+  void writeErrorOnDeprecation(PackageWarning warning) {
+    if (warning._isDeprecated) {
+      stderr.writeln("The warning 'missingCodeBlockLanguage' is deprecated. "
```
srawlins, line 478: I guess this should refer to `_flagName`. And then we don't need the second sentence; it can just be spelled out in the help text.

srawlins (approved):

PR #3746, opened 2024-04-12
Properly sort unnamed libraries

srawlins:
Found a few little issues with testing:

* The `named` extension getter should only work if there is a _single_ matching element.
* Fixing that showed the issue of re-exported extensions, so I made the `extensions` and `functions` extension getters only return documented members.

---

<details>

</details>

### DIFF
--- a/lib/src/model/nameable.dart
+++ b/lib/src/model/nameable.dart
@@ -97,7 +97,7 @@ mixin Nameable implements Privacy {
 /// Compares [a] with [b] by name.
 int byName(Nameable a, Nameable b) {
   if (a is Library && b is Library) {
-    return a.displayName.compareTo(b.displayName);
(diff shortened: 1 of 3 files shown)


lib/src/model/nameable.dart
```diff
@@ -96,9 +96,14 @@ mixin Nameable implements Privacy {
 
 /// Compares [a] with [b] by name.
 int byName(Nameable a, Nameable b) {
+  if (a is Library && b is Library) {
+    return a.displayName.compareTo(b.displayName);
```
kevmoo, line 100: should we stay consistent and use `compareAsciiLowerCase...` here too?

kevmoo (approved):

lib/src/model/nameable.dart
srawlins: Actually yes, that's a good catch.

PR #3788, opened 2024-06-17
Declare local variables for the 'href' field, to promote and avoid printing 'null'

srawlins:
This change would have more teeth to it if we had a lint rule that guarded against interpolating `null` into a string (https://github.com/dart-lang/linter/issues/3149), but I think it still makes the code a teensy bit safer / more readable.

---

<details>

</details>

### DIFF
--- a/lib/src/markdown_processor.dart
+++ b/lib/src/markdown_processor.dart
@@ -314,15 +314,13 @@ class MarkdownDocument extends md.Document {
     var textContent = _htmlEscape.convert(referenceText);
     var linkedElement = result.commentReferable;
     if (linkedElement != null) {
-      if (linkedElement.href != null) {
+      var href = linkedElement.href;
+      if (href != null) {
         var anchor = md.Element.text('a', textContent);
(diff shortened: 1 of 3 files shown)


srawlins:
CC @dart-lang/analyzer-team

lib/src/markdown_processor.dart
```diff
@@ -314,15 +314,13 @@ class MarkdownDocument extends md.Document {
     var textContent = _htmlEscape.convert(referenceText);
     var linkedElement = result.commentReferable;
     if (linkedElement != null) {
-      if (linkedElement.href != null) {
+      var href = linkedElement.href;
+      if (href != null) {
```
scheglov, line 318: These names are short enough that we could use patterns here.
```dart
if (linkedElement.href case var href?) {}
```
and below
```dart
if (this.href case var href?) {}
```
Maybe even without `this.`?

scheglov (approved):

lib/src/markdown_processor.dart
srawlins: Thanks! Done here

PR #3797, opened 2024-06-26
Store comment reference data on ModelNode

srawlins:
This is work towards https://github.com/dart-lang/dartdoc/issues/3798.

The analyzer is now tagging Comment nodes with `references` with various static elements, all backed by the new `@docImport` syntax, via a new DocImportScope (see https://dart-review.googlesource.com/c/sdk/+/345361 and https://dart-review.googlesource.com/c/sdk/+/353232). But they're not available in the element model. So while we have the syntax nodes, in PackageGraph, we must "side-car" the data, by storing it in separate classes (with no references to any AST nodes).

So in this change we introduce the CommentReferenceData class, attach instances to ModelNode.

---

<details>

</details>

### DIFF
--- a/lib/src/model/package_graph.dart
+++ b/lib/src/model/package_graph.dart
@@ -5,10 +5,11 @@
 import 'dart:collection';
 
 import 'package:analyzer/dart/analysis/analysis_context.dart';
-import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/element/element.dart';
 import 'package:analyzer/file_system/file_system.dart';
 import 'package:analyzer/source/source.dart';
+// ignore: implementation_imports
+import 'package:analyzer/src/dart/ast/ast.dart';
 // ignore: implementation_imports
 import 'package:analyzer/src/dart/element/inheritance_manager3.dart'
     show InheritanceManager3;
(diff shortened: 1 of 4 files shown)


srawlins:
CC @dart-lang/analyzer-team

lib/src/model/package_graph.dart
```diff
@@ -1029,3 +1070,41 @@ class InheritableElementsKey {
...
+        staticElement = identifier.staticElement;
+      } else if (commentReferable case SimpleIdentifier()) {
+        name = commentReferable.name;
+        staticElement = commentReferable.staticElement;
+      } else {
+        continue;
+      }
+
+      if (staticElement != null && !data.containsKey(name)) {
+        data[name] = CommentReferenceData(
```
bwilkerson, line 1100: I haven't thought about this deeply like I'm sure you have, but at first glance this appears to be prone to having information overwritten if it's possible to have two comment references with the same `name` that resolve to different elements.

For example, could a reference to a parameter in one method and a parameter with the same name in a different method cause the map to tell us that both references refer to the second method's parameter? Or a similar question about references to type parameters?

keertip (approved):

bwilkerson (approved):
LGTM, assuming that the possible overwrite issue isn't really an issue.

lib/src/model/package_graph.dart
srawlins: Good question! Since this `get data` is made available on  a single Comment, each reference in square brackets with the same name must refer to the same thing. Like in

```
/// Text [foo] and [bar] and [foo].
```

There is never a change to the scope within a doc comment, so `[foo]` and `[foo]` always refer to the same element.

And then the Map of CommentReferenceData is not merged with any other maps from any other comments. It is just stored in a ModelNode for the documented element.

PR #3805, opened 2024-06-29
Start using docImport resolution when resolving references

srawlins:
Currently dartdoc uses a two-step strategy to resolve doc comment references:

First, for each reference in square brackets, dartdoc asks analyzer's `Scope.lookup` API for what a reference should resolve to. (Note: that API does not support resolution that includes the "doc import" scope which is introduced by `@docImport`.) Second, if `Scope.lookup` could not resolve the reference, dartdoc uses it's own lookup system which searches outward for identifiers (the "universal scope" system).

In this change, we introduce a third step: between the existing two steps, we check the `ModelNode.commentData` references. These references _do_ take the "doc import" scope into account. So here is the new process:

1. Check analyzer's `Scope.lookup` API. If that returns `null`,
2. Check analyzer's earlier resolution, which includes the "doc import" scope. If that returns `null`,
3. Resolve the reference with the "universal scope" system.

This should _not_ change any existing successful resolutions into _different_, _unintended_ resolutions. Introducing the second step should _only_ start resolving references that can be resolved specifically because of an introduced `@docImport`.

The only secondary change here which is needed to facilitate the `@docImport` change is to declare that extension members are "contained" by their enclosing extension. I don't know why this wasn't the case before. It doesn't seem intentional, as the tests continue to pass. I think it was an oversight, or a misunderstanding of extension members.

---

<details>

</details>

### DIFF
--- a/lib/src/model/comment_referable.dart
+++ b/lib/src/model/comment_referable.dart
@@ -10,7 +10,6 @@ import 'dart:core';
 import 'package:analyzer/dart/element/element.dart';
 import 'package:analyzer/dart/element/scope.dart';
 import 'package:collection/collection.dart';
-import 'package:dartdoc/src/model/container.dart';
 import 'package:dartdoc/src/model/library.dart';
 import 'package:dartdoc/src/model/model_element.dart';
 import 'package:dartdoc/src/model/nameable.dart';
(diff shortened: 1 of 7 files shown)


srawlins:
CC @dart-lang/analyzer-team

lib/src/model/comment_referable.dart
```diff
@@ -97,27 +97,40 @@ mixin CommentReferable implements Nameable {
...
-    final resultElement = scope!.lookupPreferGetter(referenceLookup.lookup);
-    if (resultElement == null) return null;
+    Element? resultElement;
+    final scope = this.scope;
+    if (scope != null) {
+      resultElement = scope.lookupPreferGetter(referenceLookup.lookup);
+      if (resultElement == null) return null;
+    }
+
+    if (this case ModelElement(:var modelNode?) when resultElement == null) {
```
scheglov, line 107: Cleaver, maybe too much :-P
Match a pattern, and then check a condition that does not use the pattern.

Ideally, should be a separate `if` enclosing the pattern matching one, but one liner is tempting :-)

scheglov (approved):

lib/src/model/comment_referable.dart
srawlins: Yeah I definitely thought this was a little weird :P Specifically because it seems like the `when resultElement == null` should be first. But we cannot write `if (resultElement == null && this case ModelElement(:var modelNode?))`. I can split it up in a follow-up.

PR #3810, opened 2024-07-02
Fix slash to focus search after package:web migration

parlough:
Fixes https://github.com/flutter/website/issues/10826

These mistakes are the kind that will be caught by the `invalid_runtime_check_with_js_interop_types` lint.


srawlins (approved):
Oh, excellent, thanks!

Issue #3819, opened 2024-07-12
Alert blocks do not render on pub.dev generated docs


JaffaKetchup:
Not sure if this is the best place for the issue, please transfer if necessary.

When generating locally, the markdown alert blocks appear correctly:

![image](https://github.com/user-attachments/assets/4016bd60-2a4b-4050-9263-c4a1bc59382e)

However, in the [generated documentation](https://pub.dev/documentation/flutter_map_tile_caching/latest/flutter_map_tile_caching/FMTCStore-class.html) for a pub.dev package, the blocks do not render:

![image](https://github.com/user-attachments/assets/4abde9e0-6196-4789-b77c-ac42d3a91d7b)

parlough:
\cc @isoos I know pub.dev manually replicates some of dartdoc's styles, is this just something missing from that?

isoos:
We do support them, so this seems to be a bug. Was there any change in the output format or styles that dardoc is using? We've just released a new site version with `dartdoc 8.0.10`...

isoos:
I've found the culprit: `pub.dev` executes `dartdoc` with the `--sanitize-html` flag, and when that is on, the sanitization removes the alert block's `div` and related styles.

srawlins:
Fixed by https://github.com/dart-lang/dartdoc/pull/3820

PR #3820, opened 2024-07-15
Allow alert classnames in HTML sanitization.

isoos:
#3819

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 8.0.12
 
- * Allow `markdown-alert` class names in HTML sanitization.
+ * Allow `markdown-alert` class names in HTML sanitization. (#3820)
 
 ## 8.0.11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 8.0.12
+
+ * Allow `markdown-alert` class names in HTML sanitization.
+
 ## 8.0.11
 
 * Fix the slash key (`/`) not focusing the search box. (#3810)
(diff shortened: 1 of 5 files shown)


isoos:
Note: I have no clue what's going on in test failures, and I also don't have write access to merge this.

pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: dartdoc
-version: 8.0.11
+version: 8.0.12-dev
```
srawlins, line 2: When changing this version, 3 other spots need to change:

* CHANGELOG.md (you did)
* dartdoc_options.yaml
* lib/src/version.dart

isoos: Should I use `-dev`? We use it in pana for intermediate states that precedes release, but keep the final version in changelog. What is the rule here?

srawlins: We stick `-dev` in each of those 4 locations generally. But I can release 8.0.12 right after this is merged so that pub.dev is fixed. So you can use 8.0.12 in each location.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 8.0.12
+
+ * Allow `markdown-alert` class names in HTML sanitization.
```
JaffaKetchup, line 3:
```suggestion
 * Allow `markdown-alert` class names in HTML sanitization. (#3820)
```

JaffaKetchup (commented):
Thanks for getting to this so quickly!

pubspec.yaml
isoos: Please do publish 8.0.12 (not urgent though, just so that we don't forget)

PR #3821, opened 2024-07-16
Add support for package:web v1.0.0

srujzs:
- Expands version range to >=0.5.1 <2.0.0.
- In order to do this, we add an `innerHtml` extension interop member both so we can support both major
versions and so we don't need to do manual conversions of JSAny.
- Also updates `createElement` calls to use the constructors supplied in package:web 0.5.1 onwards.
- Publishes version 8.0.12.


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 8.0.13
```
srawlins, line 1: I actually haven't released 8.0.12 yet, so you can put these notes under 8.0.12

pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: dartdoc
-version: 8.0.12
+version: 8.0.13
```
srawlins, line 2: Can just keep this as 8.0.12.

srawlins (changes requested):

CHANGELOG.md
srujzs: Done, thanks!

pubspec.yaml
srujzs: Done.

srawlins (approved):
Thanks a million, very cool!

PR #3873, opened 2024-09-10
Support multiple partial calls from a single template

srawlins:
There was a bug in the partial-compiler-generating code that cached the results of compiling one partial, keyed only to the partial path. But a partial has a few input parameters that could differ. For example, given two classes, `A` and `B` that each extend `Base`, and a template that iterates over some `A` values and some `B` values, going into a partial for each (see the test case in the PR). The first time the partial is encountered, the compiler would create a partial compiler with an `A` context type. And the second time it is encountered, it would use the cached compiler, seeing the same partial path. But this results in passing `B` values into the partial renderer function where `A` values are expected.

The fix is to instead create multiple compilers for this partial, and let the existing deduplication code collapse them back into one. So a partial compiler for the `A` context type is created, and one is created for the `B` context type (and each compiler is used to create a renderer function). Later, the deduplication code finds a good LUB for `A` and `B` and merges the two renderer functions into one.

---

<details>

</details>

### DIFF
--- a/lib/src/generator/templates.aot_renderers_for_html.dart
+++ b/lib/src/generator/templates.aot_renderers_for_html.dart
@@ -91,7 +91,7 @@ String renderCategory(CategoryTemplateData context0) {
     var context6 = context1.publicMixinsSorted;
     for (var context7 in context6) {
       buffer.write('\n      ');
-      buffer.write(_renderCategory_partial_container_3(context7));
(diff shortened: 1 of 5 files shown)


srawlins:
Ready for review, @dart-lang/analyzer-team

bwilkerson (approved):

Issue #1346, opened 2017-02-17
Support default document (index.html) for cleaner URLs


eseidelGoogle:
Right now to actually get to a library root page you need to know dartdocs pattern of doing name-library.html:
https://docs.flutter.io/flutter/dart-async/dart-async-library.html

https://docs.flutter.io/flutter/dart-async/
would just get you a 404.

Obviously we could fix this in Flutter's doc generation script as well, but seems like something maybe all dartdoc users might want?

sethladd:
Presumably this means there is an index.html under dart-async/. Could there be a thing called "index" inside of a Dart library? If so, what kind of structure would we need to put all the things into sub-directories so that they don't collide with a name called "index" ?

We might be able to teach Firebase what the "index.html" file is actually called. But, that would be a solution specific to Flutter's docs. I think assuming an "index.html" is a safe assumption, so many web servers will look for that file.

Just need to be careful here with collisions... but yeah, +1 for anything we can do for more beautiful URLs

kevmoo:
https://api.dartlang.org/stable/1.24.3/dart-core/

Is soooo much nicer than

https://api.dartlang.org/stable/1.24.3/dart-core/dart-core-library.html

kevmoo:
CC @athomas

srawlins:
CC @kevmoo @gspencergoog would you say this is still very desirable? I have [a PR out](https://github.com/dart-lang/dartdoc/pull/3895).

Re [collisions](https://github.com/dart-lang/dartdoc/issues/1346#issuecomment-280709414), @sethladd, good catch! Most pages are suffixed with a 'kind', like [Ticker-class.html](https://api.flutter.dev/flutter/scheduler/Ticker-class.html), but enums ([AppLifecycleState](https://api.flutter.dev/flutter/dart-ui/AppLifecycleState.html)), properties ([channelBuffers](https://api.flutter.dev/flutter/dart-ui/channelBuffers.html)), functions ([clampDouble](https://api.flutter.dev/flutter/dart-ui/clampDouble.html)), typedefs ([ChannelCallback](https://api.flutter.dev/flutter/dart-ui/ChannelCallback.html)) are not suffixed, so any named `index` would collide with this. With that lower-case, this would particularly be a top-level property or top-level function that could cause a collision. I should probably use new paths for those like `clampDouble-function.html`, and write redirecting files.

kevmoo:
I think this is still VERY desirable.

[...]

PR #3895, opened 2024-10-02
Introduce a new URL for library pages, at index.html

srawlins:
Fixes https://github.com/dart-lang/dartdoc/issues/1346

This "moves" the URL of each library from, e.g.
'package-two_lib2/package-two_lib2-library.html' to 'package-two_lib2/index.html', such that we can have links with 'package-two_lib2'. The old URLs are preserved via redirecting HTML files.

(HTTP redirects are more standard and recommended, but dartdoc does not contain an HTTP server; HTTP redirects could be implemented at the hosting server, which is different for different doc hosts (Google Storage, Firebase, ...). Instead we use simpler HTML redirects. In fact, a given backend host could use the HTML redirecting files as a database to inform HTTP redirect rules, that need to be implemented differently for each backend. So dartdoc now produces redirecting files, but a given backend could instead return an HTTP 301 Permanent Redirect response when requesting one of the redirecting files.)

This adds a negligible number of files (1 per documented library), and a negligible number of bytes (tiny redirect files), but actually is probably a net negative in terms of bytes, as many links are now shortened, removing the HTML file name of library URLs.

---

<details>

</details>


gspencergoog:
Would it make sense to have a mode that doesn't emit redirects at all, and renames the output files for all the symbol types that aren't suffixed, like variable, enum, and function (so the top level "index" variable doesn't collide with "index.html", etc.)? Then that mode could be used for "new" output, and the redirects could be used for "legacy" output.  I'd hate to think that we're stuck emitting ever larger numbers of redirect files every time we rename something in the output tree.

srawlins:
We could have a mode that stops emitting the redirecting files. You'd enable it whenever you're comfortable with some % of broken links coming from StackOverflow, articles, etc.

gspencergoog:
> We could have a mode that stops emitting the redirecting files. You'd enable it whenever you're comfortable with some % of broken links coming from StackOverflow, articles, etc.

I was imagining using it mainly for new packages, so in those cases there wouldn't be any broken links.

srawlins:
Ah, we could do that. It wouldn't be used very much, as a new package author would have to know about it and include it in their dartdoc_options.yaml, or it could be the default and offer the "please don't break old links" mode be opt-in.

gspencergoog:
Okay, maybe not worth the effort, then.  Having it be the default might cause havoc (since people wouldn't know to turn it on and make lots of broken links), and having it not be the default means it won't be used much.

lib/src/generator/templates.aot_renderers_for_html.dart
```diff
@@ -1137,6 +1137,29 @@ String renderLibrary(LibraryTemplateData context0) {
   return buffer.toString();
 }
 
+String renderLibraryRedirect(LibraryTemplateData context0) {
+  final buffer = StringBuffer();
+  buffer.write('''<!DOCTYPE html>
+<html lang="en">
```
kevmoo, line 1143: Might could add a `canonical` tag, too? For the search engines?

kevmoo: https://developers.google.com/search/docs/crawling-indexing/consolidate-duplicate-urls

srawlins: So good! Thanks Kevin, I love this. Done.

srawlins:
Ready for review @dart-lang/analyzer-team . Big feature for dartdoc.

bwilkerson (approved):

PR #3916, opened 2024-10-25
Use scss for styling

parlough:
This PR migrates the existing styles to use SCSS to benefit from nesting and so that some styles can be more easily shared with other Dart and Flutter related sites (particularly pub.dev) in the future.

It also enables minimization :)


parlough:
> Or we could land the change more incrementally.

Sure, happy to do that - https://github.com/dart-lang/dartdoc/pull/3917

parlough:
Most of this has landed in a few separate PRs, with the most recent PR https://github.com/dart-lang/dartdoc/pull/3942 taking advantage of sass's nesting support.

PR #3917, opened 2024-10-25
Generate included css using sass

parlough:
This does not change any styles, it simply changes the extension and adds the generation of a `css` file.

This has the added benefit of providing some CI validation for the file, as only valid syntax will compile successfully.

Follow-up PRs will take advantage of sass features, for a goal of improved maintainability, and eventually sharing more styles (or at least variables) with pub.dev and dart.dev.


srawlins (approved):
Love this! I've wanted to do this for forever.

PR #3961, opened 2025-01-09
Reduce use of generators in finding files to document

srawlins:
The short story of this change is that we're refactoring the "file-finding" code so that `_findFilesToDocumentInPackage` is no longer `async*`, and `_listDir` is no longer `sync*`. Lazy generators like these are nice when the full sequence of elements may not be generated. However, in this code, we definitely iterate over every element returned by these two functions, and `sync*` is known to be slow, and a package set may be thousands of files.

The longer story has more details:

* `findFilesToDocumentInPackage` was async because of the code that finds a package config, and that code is only needed when the `includeDependencies` parameter is `true`. The function only has two callers, one of which always passed `includeDependencies: false`.
  * So we can extract out the code that looks for a package config, and iterates over the listed packages, into a new function, `_findPackagesToDocument`, only called by `_getFilesToDocument`. Now `_findFilesToDocumentInPackage` doesn't need to be async, or a generator.
* Further efficiencies are found in `getFilesToDocument`: This function can actually be split into two cases: are we documenting the SDK, or not?
  * If documenting the SDK, we don't need to consider the `auto-include-dependencies` flag, as it is never used when documenting the SDK. We also don't need to consider the deprecated `include-externals` flag, as it is also never used when documenting the SDK. So we return a very simple calculation in this case. When documenting the SDK, the code is now simpler and does less work.
  * We can then isolate all of the complexity in the "not documenting the SDK" case, taking into account these two flags, and the embedder SDK files.
* Now that `_discoverLibraries` is only called once, we don't need to pass in the `addLibrary` callback. Instead, `_discoverLibraries` can directly call `uninitializedPackageGraph.addLibraryToGraph`.

I benchmarked this both documenting the SDK and googleapis package. There was no visible difference in time-to-document or memory usage when documenting the SDK. For the googleapis package, there was a -1% time-to-document, and -0.5% memory usage, which might well be within the error margins. 🤷 

---

<details>

</details>

### DIFF
--- a/lib/src/model/package_builder.dart
+++ b/lib/src/model/package_builder.dart
@@ -222,13 +222,8 @@ class PubPackageBuilder implements PackageBuilder {
   /// Uses [processedLibraries] to prevent calling [addLibrary] more than once
   /// with the same [LibraryElement]. Adds each [LibraryElement] found to
   /// [processedLibraries].
-  ///
-  /// [addingSpecials] indicates that only [SpecialClass]es are being resolved
-  /// in this round.
-  Future<void> _discoverLibraries(
-      void Function(DartDocResolvedLibrary) addLibrary,
-      Set<LibraryElement> processedLibraries,
-      Set<String> files) async {
+  Future<void> _discoverLibraries(PackageGraph uninitializedPackageGraph,
+      Set<LibraryElement> processedLibraries, Set<String> files) async {
     files = {...files};
     // Discover Dart libraries in a loop. In each iteration of the loop, we take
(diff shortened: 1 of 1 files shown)


srawlins:
CC @dart-lang/analyzer-team

@bwilkerson I was looking at this code while reminding myself how dartdoc looks for files to document. I didn't get far before I decided to simplify this code. :/

srawlins:
This is still ready for review.

pq (approved):

PR #3964, opened 2025-01-15
Elements. Migrate test/typedefs_test.dart

keertip:
Start of migrating to the new Element model.


lib/src/model/typedef.dart
```diff
@@ -5,7 +5,10 @@
 // ignore_for_file: analyzer_use_new_elements
 
 import 'package:analyzer/dart/element/element.dart';
+import 'package:analyzer/dart/element/element2.dart';
 import 'package:analyzer/dart/element/type.dart';
+// ignore: implementation_imports
+import 'package:analyzer/src/utilities/extensions/element.dart';
```
keertip, line 11: Will make changes to the analyzer package so we don't need this ignores.

srawlins:
If other migration PRs are similar to this one, it could be possible that each PR needs that regeneration step.

keertip:
CC @dart-lang/analyzer-team

srawlins (approved):

PR #3972, opened 2025-01-27
Elements. Migrate to Element2

keertip:
Element model migration. Migrate

- lib/src/model/canonicalization.dart
- lib/src/type_utils.dart
- test/mustachio/builder_test_base.dart

### DIFF
--- a/lib/src/model/canonicalization.dart
+++ b/lib/src/model/canonicalization.dart
@@ -2,30 +2,30 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-// ignore_for_file: analyzer_use_new_elements
-
-import 'package:analyzer/dart/element/element.dart';
+import 'package:analyzer/dart/element/element2.dart';
+// ignore: implementation_imports
+import 'package:analyzer/src/dart/element/element.dart';
 import 'package:dartdoc/src/model/model.dart';
 import 'package:dartdoc/src/warnings.dart';
+
+const int _separatorChar = 0x3B;
 
 /// Searches [PackageGraph.libraryExports] for a public, documented library
 /// which exports this [ModelElement], ideally in its library's package.
 Library? canonicalLibraryCandidate(ModelElement modelElement) {
   var thisAndExported =
-      modelElement.packageGraph.libraryExports[modelElement.library.element];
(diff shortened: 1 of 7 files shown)


lib/src/model/canonicalization.dart
```diff
@@ -74,10 +74,56 @@ final class _Canonicalization {
...
+        if (ancestor is LocalFunctionElementImpl) {
+          ancestor = (ancestor.wrappedElement.enclosingElement2
+                  as ExecutableElementImpl)
+              .element;
+        } else {
+          ancestor = ancestor.enclosingElement2;
+        }
+      }
+    }
+     var buffer = StringBuffer();
```
srawlins, line 113: not formatted?

lib/src/model/canonicalization.dart
```diff
@@ -74,10 +74,56 @@ final class _Canonicalization {
...
+    for (var i = 0; i < length; i++) {
+      var currentChar = component.codeUnitAt(i);
+      if (currentChar == _separatorChar) {
+        buffer.writeCharCode(_separatorChar);
+      }
+      buffer.writeCharCode(currentChar);
+    }
+  }
+
+  String _getElementLocation(Element2 element) {
```
srawlins, line 89: Could we add a comment about where this comes from, so that we can look to the other implementation if bugs arise?

lib/src/type_utils.dart
```diff
@@ -2,27 +2,10 @@
...
 // BSD-style license that can be found in the LICENSE file.
 
-// ignore_for_file: analyzer_use_new_elements
-
-import 'package:analyzer/dart/element/element.dart';
 import 'package:analyzer/dart/element/element2.dart';
 import 'package:analyzer/dart/element/type.dart';
 
 extension DartTypeExtension on DartType {
-  /// The static element associataed with this type, where documentable, and
```
srawlins, line 12: Well that's a handy migration!

srawlins (approved):

lib/src/model/canonicalization.dart
keertip: done.

lib/src/model/canonicalization.dart
keertip: added

PR #3976, opened 2025-01-29
Elements. Migrate to Element2

keertip:
Migrate

- lib/src/model/inheritable.dart
- lib/src/model/never.dart
- lib/src/model/parameter.dart
- lib/src/model/prefix.dart
- lib/src/model/typedef.dart

### DIFF
--- a/lib/src/model/never.dart
+++ b/lib/src/model/never.dart
@@ -2,18 +2,23 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-// ignore_for_file: analyzer_use_new_elements
-
 import 'package:analyzer/dart/element/element.dart';
+import 'package:analyzer/dart/element/element2.dart';
+// ignore: implementation_imports
+import 'package:analyzer/src/utilities/extensions/element.dart';
 import 'package:dartdoc/src/model/comment_referable.dart';
 import 'package:dartdoc/src/model/kind.dart';
 import 'package:dartdoc/src/model/model.dart';
 
 class NeverType extends ModelElement with HasNoPage {
   @override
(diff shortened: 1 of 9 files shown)


lib/src/model/never.dart
```diff
@@ -2,18 +2,23 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-// ignore_for_file: analyzer_use_new_elements
-
 import 'package:analyzer/dart/element/element.dart';
+import 'package:analyzer/dart/element/element2.dart';
+// ignore: implementation_imports
```
bwilkerson, line 7: We'll need to remember to search for imports of these extensions before we declare the package fully converted.

Alternatively, you could remove `element` and use `.element2.asElement` in any unconverted files. That might be less error prone.

lib/src/model/parameter.dart
```diff
@@ -2,79 +2,88 @@
...
 // BSD-style license that can be found in the LICENSE file.
 
-// ignore_for_file: analyzer_use_new_elements
-
 import 'package:analyzer/dart/element/element.dart';
+import 'package:analyzer/dart/element/element2.dart';
 // ignore: implementation_imports
 import 'package:analyzer/src/dart/element/member.dart' show ParameterMember;
+// ignore: implementation_imports
+import 'package:analyzer/src/utilities/extensions/element.dart';
```
bwilkerson, line 10: Same here.

bwilkerson (approved):

lib/src/model/never.dart
keertip: Got it. Will look into this in a follow up CL.

PR #3980, opened 2025-02-03
Elements. Migrate to Element2

keertip:
Migrate to Element2

- lib/src/model/class.dart
- lib/src/model/container.dart
- lib/src/model/inheritable.dart
- lib/src/model/inheriting_container.dart


srawlins (approved):
Looks straightforward

PR #3988, opened 2025-02-05
Elements. Migrate lib/src/model/model_element.dart

keertip:
Will do a follow up CL to remove `element` from `ModelElement` and it's hierarchy.


srawlins (approved):
Huge win! Looks mechanical, looks very clean to me.

PR #3993, opened 2025-02-10
Elements. Migrate to Element2

keertip:
Migrate to Element2

- test/end2end/model_test.dart


keertip:
@srawlins , We have an older SDK (3.6.0) to analyze dartdoc on the `ubuntu-latest` bot, and that has `analyzer_use_new_elements` failure in file `model_test.dart` due to older version of the analyzer. It looks like we cannot remove the `ignore for file` until we have a new stable build published. Am I reading this right? Is there a way to land this now?

PR #3998, opened 2025-02-14
Move a test case from e2e tests to unit tests.

srawlins:
This test case is tickled in a refactoring I am attempting. I'm moving it to be a unit test here, for better debuggability, and understandability, before I make the refactoring.


srawlins:
CC @dart-lang/analyzer-team

jwren (approved):

PR #4000, opened 2025-02-17
Rename ModelElement field element2 -> element

keertip:
Also had to rename some local variables for this refactor, to avoid shadowing.


srawlins (approved):
Love it!

PR #4007, opened 2025-02-21
Migrate some linkedName tests for types to be unit tests.

srawlins:
Moving test cases out of the end2end tests consolidates them, and makes them faster.


srawlins:
CC @dart-lang/analyzer-team

pq (approved):

PR #4014, opened 2025-03-10
Place footer below main content

parlough:
Update the site's layout implementation so that the main content can scroll with the entire viewport rather than having a separate scroll view. This allows the footer to be placed after the main content rather than fixed at the bottom and better aligns the site's layout with dart.dev and docs.flutter.dev.

- Don't limit the body to a height of 100%, so that the main content can be scrolled as part of the primary/full viewport. This is what allows the footer to be after the content.
- Make header sticky so they always remain visible at the top of the viewport.
- Make sidebars sticky, but below the header, so that they remain visible when scrolling the main content.
- Remove custom webkit scrollbar styles as they cause part of the header's chrome to go missing and are platform dependent.
  - Add `color-scheme: <light or dark>` to the theme styles so the browser provides properly themed scrollbars.
  - Removing these styles also resolves a TODO relating to the mobile version of the sidebars.

Resolves https://github.com/dart-lang/dartdoc/issues/4013


parlough:
I staged the flutter docs at: https://parlough-api-dart-dev.firebaseapp.com/doc/

srawlins:
Sorry for the delay! I meant to try this out earlier.

I just manually tested with the Dart SDK, and looked at your staged launch, @parlough . I think they look great. I'll do a code review.

CC @athomas @gspencergoog  if you'd like to see the new positioning of the footer (it's nothing surprising, but it is changed from being fixed position).

srawlins:
Thanks @parlough !

gspencergoog:
Thanks, looks good!

Issue #4021, opened 2025-03-21
Crash: Null check operator used on a null value


gs0-debiotech:
When running `dartdoc` on our application, we get the following failure:

```bash
$ dartdoc
Documenting app...
Discovering libraries...
Linking elements...
Precaching local docs for 151614 elements...
Initialized dartdoc with 366 libraries
[...snip...]
Generating docs for library data/dao/app_database_dao.dart from file:///builds/app/lib/data/dao/app_database_dao.dart...
dartdoc 8.3.3 (/root/.pub-cache/global_packages/dartdoc/bin/dartdoc.dart-3.7.0.snapshot) failed: Null check operator used on a null value
#0      Prefix.associatedLibrary (package:dartdoc/src/model/prefix.dart:30:54)
#1      Prefix.associatedLibrary (package:dartdoc/src/model/prefix.dart)
#2      Prefix.canonicalModelElement (package:dartdoc/src/model/prefix.dart:42:41)
#3      Prefix.href (package:dartdoc/src/model/prefix.dart:51:23)
#4      PackageGraph.allHrefs (package:dartdoc/src/model/package_graph.dart:599:33)
#5      new Validator (package:dartdoc/src/validator.dart:38:32)
#6      Dartdoc.generateDocsBase (package:dartdoc/src/dartdoc.dart:209:7)
<asynchronous suspension>
#7      Dartdoc.generateDocs (package:dartdoc/src/dartdoc.dart:241:24)
<asynchronous suspension>
#8      Dartdoc.executeGuarded.<anonymous closure> (package:dartdoc/src/dartdoc.dart:273:9)
<asynchronous suspension>
```

Link to [prefix.dart:30](https://github.com/dart-lang/dartdoc/blob/d40067626b8c419e451897447e4dae74d25bcc37/lib/src/model/prefix.dart#L30) as of today.

I do not know whether it's linked to the last file read. In case it is, here is the code:

```dart
import 'dart:io';

import 'package:app/data/app_database.dart';
import 'package:drift/drift.dart';

part '../../generated/data/dao/app_database_dao.g.dart';

@DriftAccessor()
class AppDatabaseDao extends DatabaseAccessor<AppDatabase> with _$AppDatabaseDaoMixin {
  AppDatabaseDao(super.attachedDatabase);

  Future<void> exportInto(File file) async {
    await file.parent.create(recursive: true);
    await customStatement('VACUUM INTO ?', [file.path]);
  }
}
```

[...]

PR #4022, opened 2025-03-21
Improve failure when import cannot be found

srawlins:
Fixes https://github.com/dart-lang/dartdoc/issues/4021

In the sense that the error state should be more visible :D


srawlins:
CC @dart-lang/analyzer-team

pq (approved):

PR #4023, opened 2025-03-25
Improve the error message when an import cannot be resolved

srawlins:
`DirectiveUri.toString()` turns out to be useless. So try to provide at least information about where the import directive starts in the source.


srawlins:
CC @dart-lang/analyzer-team

pq (approved):

PR #4025, opened 2025-03-27
Remove the deprecated nodoc option.

srawlins:
This should be landed together with other breaking changes for the 9.0.0 release.


szakarias (approved):
LGTM

PR #4031, opened 2025-04-10
Use _category name_ in the URL, instead of the _category displayName_.

jonasfj:
Also clarify documentation, to make it a bit more obvious that a category has two names:
 * _category name_, the thing we reference in source code,
 * _display name_, the thing we display in HTML.

I imagine that the _category name_ lends itself better to be used in a URL (probably using underscore). And the _display name_ is probably something using whitespaces (which don't work nearly as well in a URL).

Also correct `name` -> `displayName` in README.

--------------

I don't know if we should make _redirect pages_ for the old URLs. It's perhaps overkill, since the category feature isn't used by most packages. Does `api.dart.dev` or `api.flutter.dev` have topic pages?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 8.3.4-wip
+* The URL for category pages now uses _category name_ instead of
+  _category `displayName`_.
+
 ## 8.3.3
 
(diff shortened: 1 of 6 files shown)


srawlins (approved):
Very nice, thank you!

PR #4035, opened 2025-04-16
Remove use of deprecated parameter

bwilkerson:
This replaces the use of `updateAnalysisOptions2` with a use of `updateAnalysisOptions3`. The latter callback has one fewer parameter, but since we weren't using that parameter no other changes (other than import cleanup) are required.


bwilkerson:
I'm not sure what's causing the failures.

srawlins (approved):

PR #4037, opened 2025-04-23
Make redirect pages for categories

szakarias:


lib/src/model/category.dart
```diff
@@ -84,7 +84,7 @@ class Category
   String get name => _categoryDefinition.displayName;
```
jonasfj, line 84: Rather than introducing the term `_internalName`, perhaps it would be better we changed this to return `_name`.

And then added a `displayName` property.

Obviously, we'd have to refactor the templates to then use `displayName` where they currently use `name`.

Ofcourse, that might not be reasonable, if `name` is used in inherented templates?

szakarias: It is actually called 'internal name' in the comments already, but I actually think this renaming does not belong in this PR, so I reverted it for now.

lib/templates/category_redirect.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <link rel="canonical" href="{{ #useBaseHref }}{{ #htmlBase }}{{{ htmlBase }}}{{ /htmlBase }}{{ /useBaseHref }}{{ self.href }}" />
+    <meta http-equiv="refresh" content="0; url={{ #useBaseHref }}{{ #htmlBase }}{{{ htmlBase }}}{{ /htmlBase }}{{ /useBaseHref }}{{ self.href }}" />
+  </head>
+  <body>
+    <p><a href="{{ #useBaseHref }}{{ #htmlBase }}{{{ htmlBase }}}{{ /htmlBase }}{{ /useBaseHref }}{{ self.href }}">New URL</a></p>
+  </body>
+</html>
```
jonasfj, line 10: missing empty line

lib/src/model/category.dart
```diff
@@ -105,11 +105,15 @@ class Category
...
-  String get filePath {
+  String get fileName {
     assert(_name != null);
-    return 'topics/$_name-topic.html';
+    return '$_name-topic.html';
   }
 
+  String get filePath => 'topics/$fileName';
+
+  String get redirectFilePath => 'topics/$name-topic.html';
```
jonasfj, line 115:
```suggestion
  /// Prior to dartdoc 8.3.4 the `displayName` was used in the file path
  /// for category pages. We now create a redirect file here instead.
  String get redirectFilePath => 'topics/$name-topic.html';
```

jonasfj (approved):

srawlins (approved):
Looks great, thanks!

lib/templates/category_redirect.html
szakarias: done

lib/src/model/category.dart
szakarias: done

PR #4040, opened 2025-04-28
Stop using ConstructorElementImpl

srawlins:
This simplifies the code and reduces use of private API.


srawlins:
CC @jonasfj @szakarias

PR #4041, opened 2025-04-28
Stop using InterfaceElementImpl2

srawlins:
This simplifies the code and reduces use of private API.


srawlins:
CC @jonasfj @szakarias

PR #4043, opened 2025-04-30
Fix duplicate entries of elements in a category when re-exported.

jonasfj:
If a package has two libraries `foo.dart` and `bar.dart` where
 * `foo.dart` contains `Foo` with an `/// {@category ...}` annotation, and,
 * `bar.dart` re-exports `foo.dart`.

Then the category page shall not list `Foo` twice, which is simply a bug.

This change fixes this issue.

--------

There is an example of the issue in production here:
https://pub.dev/documentation/typed_sql/latest/topics/Exceptions-topic.html

image


srawlins (approved):
Very nice!

szakarias (approved):
LGTM

Issue #2418, opened 2020-11-04
Don't document @internal members


nex3:
https://github.com/dart-lang/sdk/issues/28066 added support for an `@internal` annotation, which can be used to mark members as private to the package that defines them. Because these aren't part of a package's public API, DartDoc shouldn't generate documentation for them.

srawlins:
Similar to https://github.com/dart-lang/dartdoc/issues/1664. I think dartdoc should not generate docs for these.

If we get feedback that some users want docs for these, we can add a flag, but I don't see a strong case for it now.

sigurdm:
This was fixed

nex3:
Finally, a farewell to all these `@nodoc`s!

PR #4048, opened 2025-05-12
Don't document elements with `@internal`

sigurdm:
Makes members with the `@internal` not part of the public interface.

#2418


testing/test_package/lib/example.dart
```diff
@@ -30,6 +30,14 @@ const String COMPLEX_COLOR = 'red' + '-' + 'green' + '-' + 'blue';
 /// @nodoc
 const DO_NOT_DOCUMENT = 'not documented';
 
+/// top level internal variable
+@internal
+final topLevelInternal = 'not documented';
+
+/// top level testing function
+@visibleForTesting
+String testingMethod() => 'not documented'
```
szakarias, line 39:
```suggestion
String testingMethod() => 'not documented';
```

testing/test_package/lib/example.dart
```diff
@@ -122,6 +130,14 @@ class Apple {
 
   /// @nodoc no docs
   int? notDocumented;
+  
+  /// No public docs for this
+  @internal
+  int? internalField
```
szakarias, line 136:
```suggestion
  int? internalField;
```

testing/test_package/lib/example.dart
```diff
@@ -122,6 +130,14 @@ class Apple {
...
+  
+  /// No public docs for this   
+  @visibleForTesting
+  int? testField
```
szakarias, line 140:
```suggestion
  int? testField;
```

szakarias (approved):

rrousselGit:
Honestly I'm a bit skeptical with hiding `visibleForTesting`. Those are visible, just under some conditions

visibleForTesting allows _any_ test suite to access those utils, not just the test folder of the defining package.
A package can define a visibleForTesting element. And apps can use it in their tests.


So it is public API.

sigurdm:
> And apps can use it in their tests.

According to the documentation it only allows test suites from the same package to access it:
image
https://pub.dev/documentation/meta/latest/meta/visibleForTesting-constant.html

Not sure if that is how the warnings actually work

testing/test_package/lib/example.dart
sigurdm: Thanks!

sigurdm:
I have tried out the warnings. And indeed it seems you can import a `@visibleForTesting` from anywhere without getting a warning. So either we need to fix the docs or the implementation. (Or maybe add a separate annotation)

srawlins:
> which is in the test folder of the defining package.

Oops, yeah. That's now how the analyzer runs things. But also, I think I prefer the analyzer's interpretation, over the documented behavior ("of the defining package"). We could change the docs.

rrousselGit:
I like the current behavior of `visibleForTesting`.
If it only worked in the defining package, it'd pretty much be a duplicate of `@internal`


I use `VisibleForTesting` in ways that allow usage outside the package.
For instance, Riverpod has a Notifier class defined as:

```dart
abstract class Notifier<T> {
  @protected
  @visibleForTesting
  T state;
}
```

So within production code, "state" is protected, to promote good separation of concern.
But tests have direct access to "state" so that folks can more easily initialize their notifiers.

___

We could also imagine that a package defining `Foo` would also ship an `isFoo` Matcher for testing sake. And mark that matcher as `visibleForTesting`

This avoids having to make two packages to separate dev dependencies from non-dev ones, without compromising tree shaking.

sigurdm:
Not sure where to go from here.

It seems that both of these annotations have a vague relation re. what it means for the annotatee to the exported.

Sounds like we could use 4 annotations:

* `@visibleForOwnTests`
* `@visibleForOthersTests`
* `@warnIfExportedFromPackage`
* `@warnIfImportedInPackage`

All of these seem to have at least some use cases. And the documentation around `@internal` and `@visibleForTesting` is vague.

Filed https://github.com/dart-lang/sdk/issues/60731
and https://github.com/dart-lang/sdk/issues/60732

rrousselGit:
We're not limited to applying a single annotation.

So given:

> * `@visibleForOwnTests`
> * `@visibleForOthersTests`

We could combine visibleForTesting+internal to achieve both:


```dart
@internal
@visibleForTesting
int a; // only usable in the tests of the defining package

@visibleForTesting
int b; // usable by any test
```


I'm not sure what the current behavior is, but that feels like a good solution

lib/src/model/model_element.dart
```diff
@@ -427,6 +427,12 @@ abstract class ModelElement
       }
     }
 
+    if (element.nonSynthetic2 case Annotatable(:var metadata2)) {
```
sigurdm, line 430: I found the `nonSynthetic2` getter.

Perhaps that can also be used above instead of all the dancing around "PropertyinducingElement"

PR #4060, opened 2025-05-28
Remove Templates.fromContext and as a result, make code more synchronous

srawlins:
Since "runtime template" support was removed, we could remove the `forceRuntimeTemplates` parameter from `Templates.fromContext`. However, the result of this is that calling `Templates.fromContext` would not need the `DartdocGeneratorOptionContext` parameter either, and is then just an alias for the `HtmlAotTemplates` constructor. So we can remove the function altogether.

Since we no longer need to `await` the result of calling this function (which is now a constructor call), we can make more and more functions synchronous (see
https://journal.stuffwithstuff.com/2015/02/01/what-color-is-your-function/): `bin/dartdoc.dart`'s `main`, `Dartdoc.fromContext` (which can new be a factory constructor), `initHtmlGenerator`, `buildDartdoc`, `generatorContextFromArgv`, `createPackageBuilder`.

We can also remove some more code that becomes dead: `RuntimeTemplates._create`, ResourceProvider utilities, and some trivial `buildDartdoc` test helpers.


szakarias (approved):
Cool with all these cleanups! Thanks.

PR #4062, opened 2025-05-31
Elements. element.lookupName can be null, use empty string then.

scheglov:
Found during https://dart-review.googlesource.com/c/sdk/+/432125


szakarias (approved):
Thanks, LGTM

PR #4070, opened 2025-07-01
Fix lints and tests.

isoos:


test/end2end/dartdoc_test.dart
```diff
@@ -110,7 +110,7 @@ void main() {
       expect(
           useSomethingInAnotherPackage.modelType.linkedName,
           matches(
-              r'<a class="deprecated" href="https://pub.dev/documentation/meta/[^"]*/meta/Required-class.html">Required</a>'));
+              r'<a href="https://pub.dev/documentation/meta/[^"]*/meta/Required-class.html">Required</a>\?'));
```
sigurdm: Is this test change a consequence of the change above, or is it an already failing test thats fixed?

lib/src/model/canonicalization.dart
```diff
@@ -89,7 +89,7 @@ final class _Canonicalization {
     Element2? ancestor = element;
     while (ancestor != null) {
       if (ancestor is LibraryElement2) {
-        components.insert(0, ancestor.identifier);
+        components.insert(0, ancestor.uri.toString());
```
sigurdm, line 92: Is this really equivalent?

sigurdm: Ah - this is only for libraries - I guess it makes sense then.

isoos: identifier has a deprecation note that we should use uri instead

test/end2end/dartdoc_test.dart
isoos: updated the regexp PTAL, apparently different test versions output different text.

isoos:
Note: the CI test failure was tracked back to `package:meta`'s recent change where "Mark `Required` and `required` as `@Deprecated`." Fixing the test with the now-new `?` mark, and also bumping the CI's SDK version, so that all tests can and will use  the latest `package:meta` version consistently.

PR #4073, opened 2025-07-03
Formatting Dart code + enforcing it on CI.

isoos:
Note: I have no idea whether this is good/required for the template files with lots of changes, but for the rest of the files it seems somewhat important to make them consistent with other project workflows where the `dart format` is prerequisite of the approval.

I've also extended the CI tool to run it before the `dart analyzer` is run (in the same command block, for simplicity, and also because there is very little reason to run them separately).


tool/task.dart
```diff
@@ -118,6 +121,10 @@ Future<void> analyzeTestPackages() async {
   }
 }
 
+Future<void> _runFormatter() async {
```
szakarias, line 124: maybe update `_runFormatter()` to accept specific paths and then run it only where the builder actually outputs its generated files

tool/task.dart
```diff
@@ -118,6 +121,10 @@ Future<void> analyzeTestPackages() async {
...
+  await SubprocessLauncher('formatter').runStreamedDartCommand(['format', '.']);
```
szakarias:
```suggestion
  await SubprocessLauncher('format').runStreamedDartCommand(['format', '.']);
```

for consistency with `analyzePackage()`

tool/task.dart
isoos: Not sure: where to get this information from?

PR #4074, opened 2025-07-04
Minimal screenshot tool.

isoos:
- The `test_browser.dart` and the `screenshot_util.dart` is a slightly adjusted copy from `dart-lang/pub-dev`. I think in the long term we may want to extract it into a utility package that could be imported by both projects.
- Uses an end2end test and observes its output with a local http server (thus the index.json fetching also works).
- Right now only a few random pages are selected, we should probably handpick a few more, or maybe even make it crawl the entire output, not sure about that part.
- Screenshotting the search dropdown is TBD.
- Comparison tool is TBD, I think we could just start with the tool from `pub-dev` (or again, extract it into a shared package).

### DIFF
--- a/test/end2end/dartdoc_test.dart
+++ b/test/end2end/dartdoc_test.dart
@@ -57,8 +57,6 @@ void main() {
       // Set up the pub metadata for our test packages.
       runPubGet(testPackageToolError.path);
       runPubGet(_testSkyEnginePackage.path);
-
-      if (isScreenshotDirSet) {}
     });
 
     setUp(() async {


test/end2end/dartdoc_test.dart
```diff
@@ -51,6 +57,8 @@ void main() {
       // Set up the pub metadata for our test packages.
       runPubGet(testPackageToolError.path);
       runPubGet(_testSkyEnginePackage.path);
+
+      if (isScreenshotDirSet) {}
```
sigurdm: intentional?

isoos: Yes. We want to run screenshots only when the output directory is set, regular tests should run without if - at least for now.

sigurdm: But the "then" branch is empty...

isoos: Ah, my bad, mobile view of the code needs more attention.

No, this is not intentional, a leftover, will fix it once I'm with a computer.

pubspec.yaml
```diff
@@ -27,7 +27,10 @@ dev_dependencies:
   dart_style: ^3.0.0
   lints: ">=5.0.0 <7.0.0"
   matcher: ^0.12.15
+  puppeteer: ^3.18.0
```
sigurdm, line 30: We might need to discuss if we want a non-dart-team dependency here (even if it is a dev-dependency)
cc @szakarias wdyt?

isoos: I was thinking to implement this as an external screenshotting tool, mostly because there is no natural integration point with browser-based tests like in `pub-dev`'s integration tests, and maybe `pub-dev` could reuse the tool as a library. wdyt?

Issue #4075, opened 2025-07-10
Generated API reference points to invalid links for Flutter/Dart classes leading to 404s


sigurdm:
Copied from: https://github.com/flutter/flutter/issues/170868

Steps to reproduce
The generated API reference for packages seems to have a number 404s referencing invalid pages for Flutter/Dart classes. This was picked up from a review of API ref for arcgis_maps but seems to apply to any generated API ref - here are a few examples from "Flutter Favorite" packages:

https://pub.dev/documentation/path_provider/latest/path_provider/
E.g. click on Future referenced from various functions
https://pub.dev/documentation/provider/latest/provider/StreamProvider-class.html
E.g. click on Stream referenced in the doc
https://pub.dev/documentation/firebase_core/latest/firebase_core/Firebase-class.html
E.g. click on Future referenced in the methods
This doesn't seem to affect all/most classes, but from our review it has at least affected references to Future, Stream, and Color.

References to Future go to https://api.flutter.dev/flutter/dart-core/Future-class.html instead of https://api.flutter.dev/flutter/dart-async/Future-class.html

References to Stream go to https://api.flutter.dev/flutter/dart-core/Stream-class.html instead of https://api.flutter.dev/flutter/dart-async/Stream-class.html

References to Color go to https://api.flutter.dev/flutter/painting/Color-class.html instead of https://api.flutter.dev/flutter/material/Colors-class.html

Has anyone else seen this? Is this likely a doc generator issue? Or anything configuration wise at the package level?

Expected results
All API references go to the correct, corresponding page.

Actual results
Some API references are hitting invalid pages / 404.

Code sample
Code sample
n/a - see link examples above

Screenshots or Video
Screenshots / Video demonstration
n/a

Logs
Logs
n/a

Flutter Doctor output
Doctor output
n/a

sigurdm:
Is also a problem when the reference goes to api.dart.dev https://pub.dev/documentation/retry/latest/retry/

PR #4076, opened 2025-07-10
Canonicalize elements in remote packages

sigurdm:
Attempt at fixing https://github.com/dart-lang/dartdoc/issues/4075, https://github.com/flutter/flutter/issues/170868

Before dartdoc would canonicalize Future to dart:async when generating sdk docs, but not when generating a link from package docs.

### DIFF
--- a/test/end2end/dartdoc_test.dart
+++ b/test/end2end/dartdoc_test.dart
@@ -110,7 +110,7 @@ void main() {
       expect(
           useSomethingInAnotherPackage.modelType.linkedName,
           matches(
-              r'<a class="deprecated" href="https://pub.dev/documentation/meta/1.17.0/meta/Required-class.html">Required</a>'));
+              r'<a class="deprecated" href="https://pub.dev/documentation/meta/[^"]*/meta/Required-class.html">Required</a>\?'));
       var link = RegExp('/dart-core/String-class.html">String</a>');
       expect(useSomethingInTheSdk.modelType.linkedName, contains(link));
     });


sigurdm:
Sorry for the long delay on this one.

The tests seem to fail on windows only. Not sure what is going on.

@srawlins do you understand why we get this? Is it because the file system on windows is not case-sensitive?
https://github.com/dart-lang/dartdoc/actions/runs/18772999367/job/53561163122?pr=4076#step:10:673

srawlins:
Very strange, the PR is so simple, and the affected test is dartdoc_test.dart, I believe, which is looking at the `testing/` packages. My guess is the backslashes: in HTML all of the links should be forward slashes. Maybe there is logic somewhere in there that is using filepaths, with backslashes, in a way that was not used before this PR.

srawlins:
Ah this code suspiciously looks at slashes and has the word `remote` 🤣 

https://github.com/dart-lang/dartdoc/blob/6d5e133826177f8e3dfa2f1842f9427f3fedac33/lib/src/model/package.dart#L194-L196

`ModelElement.href` uses this getter, and `ModelElement.href` is how each element is linked to.

sigurdm:
Hmm, after digging more, I think that is a red herring.

Here are some random notes on my progress

It seems like the same broken links diagnostic get reported on both windows and non-windows.

The diagnostic gets reported for the element `WhatAClass` in the test_package_imported

https://github.com/dart-lang/dartdoc/blob/1ed34221ccf174c73a612d69e08063e8898915ab/testing/test_package_imported/lib/main.dart#L4

On linux in https://github.com/dart-lang/dartdoc/blob/41c34be18a0f5c868d099a814744a6fb5845ac8b/lib/src/warnings.dart#L550 e is set, and the options (of which there are none) from test_package_imported gets used - so it is a warning.

But on windows e is null, so by default the dartdoc_options.yaml file in test_package gets used, that sets the broken links to an error

https://github.com/dart-lang/dartdoc/blob/8e6a95c4aba8def4ef01ccc600b3a3cfb8f3cb9e/testing/test_package/dartdoc_options.yaml#L23

Still not sure why we see that difference.

sigurdm:
I won't find the time to pursue this. The issue is still real.

PR #4086, opened 2025-08-15
Require analyzer ^8.1.1 and migrate off deprecated APIs

fshcheglov:


lib/src/model/inheriting_container.dart
```diff
@@ -179,6 +179,17 @@ abstract class InheritingContainer extends Container {
           elements.firstWhereOrNull((e) => e is GetterElement);
       final setterElement =
           elements.firstWhereOrNull((e) => e is SetterElement);
+
+      final isNullOrTopMergeGetter = getterElement == null ||
```
srawlins, line 183: Thanks for breaking this down into pieces; much easier to understand.

srawlins (approved):
Thanks for the migration! I reviewed this all, and it looks great. But I'll let the Pub team approve and land.

Mostly a simple migration. Anything around Annotatable is simplified.

@fshcheglov , I see all of the migrations from "Member" to "Substituted...", and these are all private Impl classes. It would be great it dartdoc did not depend on so much private analyzer impl. While this is perhaps fresh in your mind, can you say perhaps whether you think the dartdoc is unnecessarily reliant on these Substituted classes, or whether it is reasonable, and the Substituted classes should be made part of analyzer's public API?

lib/src/model/inheriting_container.dart
fshcheglov: Yes, I took a look and it was simple to replace members with API classes.

Thank you for the code review.

fshcheglov:
Sorry for churn in https://github.com/dart-lang/dartdoc/pull/4086/commits/99656458d2c3171b70e3d61dc2707e477bd053dd and https://github.com/dart-lang/dartdoc/pull/4086/commits/a8689640f0c53f0d33573a19bf8f95b72f8b0897. I accidentally sorted `model_element.dart` in the first commit, and then reverted and re-did it without sorting. Unfortunately it looks that GitHub does not allow seeing changes between two commit points, only individual commits. But changes look cleaner in the whole PR changes.

lib/src/model/inheriting_container.dart
srawlins: Wow... I did not realize all of the Member classes were being used totally unnecessarily :/

srawlins:
No worries. Thanks for the cleanup!

fshcheglov:
@szakarias PTAL, I updated the changelog.

PR #4089, opened 2025-08-21
Add test for dot shorthands

szakarias:
Tests for dot shorthands in docs. Will follow up with annotations and parameter default values

### DIFF
--- a/test/dot_shorthands_test.dart
+++ b/test/dot_shorthands_test.dart
@@ -1,0 +1,60 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+import 'package:test/test.dart';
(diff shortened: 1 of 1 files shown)


test/dot_shorthands_test.dart
```diff
@@ -0,0 +1,60 @@
...
+
+    expect(m.hasDocumentationComment, true);
+    expect(
+      m.documentationAsHtml,
+      '<p><a href="$linkPrefix/Color.html">Color</a> can be referenced but '
+      '<code>.red</code> cannot.</p>',
+    );
+  }
+
+  void test_doc_dot_shorthand_method_invocation_success() async {
```
kallentu: Can you add an additional test for constructors as well?

kallentu (approved):
Thanks for working on this!

sigurdm (approved):

test/dot_shorthands_test.dart
szakarias: done.

PR #4095, opened 2025-08-26
Refactor handling of generateDocs:false

sigurdm:
Not sure if anyone is using this option really. But I think the handling can be simplified.

But it seems that having multiple Generator classes just for deciding whether to write files or not is excessive, and likely to get out of sync (does the EmptyGenerator really still do whatever GeneratorFrontend does, just without writing files?)


szakarias (commented):
Overall looks good, will take another pass tonight.

srawlins (approved):
Love it! Definitely simpler!

szakarias:
I have a couple of nits, I will add it in a follow-up PR.

Also, I was thinking of a slightly different design, where we might want to keep the `Generator` class abstract. I'll put it in a draft for discussion. :-)

lib/src/dartdoc.dart
```diff
@@ -144,19 +161,6 @@ class Dartdoc {
   @visibleForTesting
   set generator(Generator newGenerator) => _generator = newGenerator;
 
-  /// Factory method that builds Dartdoc with an empty generator.
-  factory Dartdoc.withEmptyGenerator(
```
devoncarew, line 148: FYI, the `dart` tool is using this in its implementation of `dart doc`; see https://github.com/dart-lang/sdk/blob/main/pkg/dartdev/lib/src/commands/doc.dart#L119.

I'm sure its possible to update the tool for this change, but we'll want to do that as part of the next roll of this repo into the sdk.

sigurdm: Thanks! I was unaware of that. I'll make a migration CL...

devoncarew: Gotcha; tracking the work here so we don't lose track: https://github.com/dart-lang/dartdoc/issues/4100

PR #4096, opened 2025-08-27
Small cleanup around generators

szakarias:
Nits following https://github.com/dart-lang/dartdoc/pull/4095/


sigurdm (commented):
Why do we actually have this init method?
Why not just a constructor?

szakarias:
> Why do we actually have this init method? Why not just a constructor?

Yeah, we could consider just inlining the call...

PR #4099, opened 2025-09-04
Rename LanguageFeature and remove FeatureSet mixin

szakarias:
`LanguageFeature` was introduced in https://github.com/dart-lang/dartdoc/pull/2221/ for chips indicating null safety. Since we now use chips for other purposes, such as class modifiers, the name is no longer descriptive.

This change renames it to `Tag` to better model the general concept of a highlightable tag on an element. This is somewhat related to`Attribute` , and the two could potentially be merged in a future step.

Furthermore, `FeatureSet` is removed and its functionality is inlined into `ModelElement`.

Finally `ContainerModifier` is extended to store the class modifier description and URL directly, instead of in separate top level maps .


test/class_modifiers_test.dart
```diff
@@ -63,21 +62,21 @@ base mixin O {}
...
-    expect(Gclass.languageFeatureChips(), equals('abstract base'));
-    expect(Hclass.languageFeatureChips(), equals('abstract interface'));
-    expect(Iclass.languageFeatureChips(), equals('abstract final'));
-    expect(Jclass.languageFeatureChips(), equals('mixin'));
-    expect(Kclass.languageFeatureChips(), equals('base mixin'));
-    expect(Lclass.languageFeatureChips(), equals('abstract mixin'));
-    expect(Mclass.languageFeatureChips(), equals('abstract base mixin'));
-    expect(Nmixin.languageFeatureChips(), equals(''));
-    expect(Omixin.languageFeatureChips(), equals('base'));
+    expect(Aclass.classModifierChips(), equals(''));
```
sigurdm: I would just inline the method here (without the join):


```suggestion
    expect(Aclass.tags, []);
```

szakarias: Agree!

lib/src/model/tag.dart
```diff
@@ -0,0 +1,48 @@
...
+    if (url != null) {
+      buffer.write('<a href="');
+      buffer.write(url);
+      buffer.write('"');
+    } else {
+      buffer.write('<span');
+    }
+
+    buffer.write(' class="feature feature-');
+    buffer.writeAll(name.toLowerCase().split(' '), '-');
```
sigurdm, line 34: Do we need html escaping here?
Do we have a guarantee that name and description don't contain `> < &`?

sigurdm: I guess that doesn't change in this PR, so let's leave this for now.

szakarias:
cc @srawlins

lib/src/model/tag.dart
srawlins: I don't think we have an actual guarantee. We could put an `assert` in the constructor. Otherwise, yeah I think this is safe. And I think we don't have any user-creatable tags.

PR #4104, opened 2025-09-10
Remove "parsedConstructorHint" from "PrefixResultType"

szakarias:
I stumbled across this which, I believe, is from when we supported leading "new" in comment references.


srawlins (approved):
Good find!

PR #4105, opened 2025-09-17
Undo deprecation of api exports

szakarias:
Preparing for 9.0.0

We will not be removing these exports anytime soon, so we are reversing the deprecation for now.


sigurdm (approved):

sigurdm:
Do you want to also undeprecate options.dart?

szakarias:
> Do you want to also undeprecate options.dart?

Yeah, we can do it in the same round.

PR #4109, opened 2025-09-23
Bump updateAnalysisOptions parameter for deprecation

srawlins:
When analyzer 8.2.0 was published, with a new deprecation, it started this cascade of issues:

1. We currently run two SDKs in CI: one set of tests uses Dart 3.8.0, and another uses Dart dev (3.10-something). Those dev bots pick up analyzer 8.2.0, which triggers a CI failure due to the deprecated `updateAnalysisOptions3` parameter.
2. In order to stop using that parameter and instead use `updateAnalysisOptions4`, we have to always get analyzer 8.2.0, because analyzer 8.1.1 does not have the new parameter.
3. In order to get analyzer 8.2.0, we need Dart 3.9.0. So this PR changes our CI to use Dart 3.9.0 and Dart dev.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -22,7 +22,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        sdk: [3.8.0, dev]
+        sdk: [3.9.0, dev]
         job: [main, flutter, packages, sdk-docs]
         include:
           - os: macos-latest


srawlins:
Thanks for the review, @szakarias ! Unfortunately, I had to widen it a bit and bump the CI to use Dart 3.9.0 instead of 3.8.0, so that we always pick up analyzer 8.2.0. Would you mind approving again?

szakarias:
> Thanks for the review, @szakarias ! Unfortunately, I had to widen it a bit and bump the CI to use Dart 3.9.0 instead of 3.8.0, so that we always pick up analyzer 8.2.0. Would you mind approving again?

LGTM! Thanks!

Issue #4107, opened 2025-09-23
Support for new Deprecated annotations


srawlins:
We've landed 5 new `@Deprecated` annotations in the Dart SDK. See [the docs](https://api.dart.dev/dev/latest/dart-core/Deprecated-class.html).

We probably want priority support in dartdoc for displaying this annotations, similar to how we display `@deprecated` annotations.

These are being released in Dart 3.10.

PR #4115, opened 2025-10-03
Display constructor names in annotations

srawlins:
Two changes in one here:

* I saw in manual testing that a class annotated with `@Deprecated.extend()` was shown in dartdoc's HTML to just be annotated with `@Deprecated`. 🤕 ouch. Dunno how that has been the case for so many years; I guess named constructors are not used too much in annotations 🤷 . So this PR makes dartdoc use the _constructor_ involved (`Deprecated.extend`) instead of the _type_ which is constructed (`Deprecated`).
* _However_, that change makes all regular `@Deprecated("reason")` annotations be printed as `@Deprecated.new("reason")`. That `.new` looks _very non-idiomatic_. Gross. So I coupled another change here where we override `displayName` on `Constructor` to _not_ include `.new` if we're looking at an unnamed constructor.

This should be good to go. We might deprecate extending `RegExp` and `RegExpMatch` in Dart 3.10; it'd be cool if we displayed the deprecated annotations correctly for this release. Note that as per https://github.com/dart-lang/dartdoc/commit/669b15f7da5da04a38e95e8ac0ff6471697a549f, we won't show the element as ~~struck through~~, so the release is safe. But this PR would be a cherry on top.

Fixes https://github.com/dart-lang/dartdoc/issues/4107

### DIFF
--- a/lib/src/model/annotation.dart
+++ b/lib/src/model/annotation.dart
@@ -5,6 +5,7 @@
 import 'dart:convert';
 
 import 'package:analyzer/dart/element/element.dart';
+// ignore: implementation_imports
 import 'package:analyzer/src/dart/element/member.dart';
 import 'package:dartdoc/src/element_type.dart';
 import 'package:dartdoc/src/model/attribute.dart';
--- a/lib/src/model/annotation.dart
+++ b/lib/src/model/annotation.dart
@@ -5,8 +5,7 @@
 import 'dart:convert';
 
 import 'package:analyzer/dart/element/element.dart';
-// ignore: implementation_imports
-import 'package:analyzer/src/dart/element/member.dart';
+import 'package:analyzer/dart/element/type.dart';
 import 'package:dartdoc/src/element_type.dart';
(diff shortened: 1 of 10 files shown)


lib/src/model/annotation.dart
```diff
@@ -36,14 +38,46 @@ final class Annotation extends Attribute {
...
+        ConstructorElement element =>
+          _packageGraph.getModelForElement(element).linkedName,
+        _ => _modelType.linkedName
+      };
+
+  /// The linked type argument text, with `<` and `>`, if there are any type
+  /// arguments.
+  String get _linkedTypeArguments {
+    var element = _annotation.element;
+    if (element is! SubstitutedConstructorElementImpl) {
```
srawlins: Hi @scheglov , I had to use `SubstitutedConstructorElementImpl` here, in order to get the type arguments on the constructor call of an annotation.

I don't see another way to get the information from the ElementAnnotation. This does feel hacky though, walking through the `substitution` field on private API.

PR #4128, opened 2025-11-15
Pass 'withFineDependencies: true' to AnalysisContextCollectionImpl.

scheglov:
I plan to make this formal parameter required.

Although it is sad that internal `AnalysisContextCollectionImpl` is used.

### DIFF
--- a/lib/src/model/package_builder.dart
+++ b/lib/src/model/package_builder.dart
@@ -65,6 +65,7 @@ class PubPackageBuilder implements PackageBuilder {
           analysisOptions
             ..warning = false
             ..lint = false,
+      withFineDependencies: true,
     );
     return PubPackageBuilder._(
       config,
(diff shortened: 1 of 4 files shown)


sigurdm:
> Although it is sad that internal `AnalysisContextCollectionImpl` is used.

I agree. I don't have the full context, but I wonder if we could get away with not using it?

srawlins:
> I don't have the full context, but I wonder if we could get away with not using it?

I don't know if this is the _only_ reason, but one reason is setting those `..warning = false..lint = false` values; at the time they were added, they saved 5% time to generate docs, which I felt was valuable.

I'd like to remove that setting, see https://github.com/dart-lang/sdk/issues/61520, in favor of a more "process-wide" or "Analysis context collection-wide" setting to say, "when I ask to resolve code, please, no need to check lint rules, or warning rules, or even errors, and no need to enter function bodies. Just give me the resolved ASTs and element model."

sigurdm:
I wonder if those settings could be exposed in the public api? They sound generally useful...

PR #4130, opened 2025-11-25
Add dart documentation comment specification version 1.0.0

szakarias:


doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,525 @@
...
+
+[TOC]
+
+## **1\. Introduction**
+
+### **1.1. Purpose**
+
+The purpose of this document is to provide a single specification for documentation comments in the Dart programming language. Its goal is to ensure that documentation is parsed consistently by all tools and to serve as a definitive standard for developers and tool authors.
+
+###
```
sigurdm: What are these empty headlines?

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,525 @@
...
+* The mechanism for *resolving* references between elements.
+* The logic for *associating* documentation with code elements.
+
+###
+
+**Out of Scope**
+
+This document does not cover:
+
+* The implementation details of specific documentation tools (e.g., `dartdoc`).
```
sigurdm: Maybe mention the other elephant in the room...
```suggestion
* The implementation details of specific documentation tools (e.g., `dartdoc`, `analysis server`).
```

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,525 @@
...
+* The syntax and behavior of tool-specific directives (e.g., `@template`, `@canonicalFor`).
```
sigurdm, line 28: Should we here link to the other document we have for the directives?

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,525 @@
...
+* Any details related to the visual styling or HTML layout of generated documentation.
+* Best practices or style guides for writing effective documentation content.
+
```
sigurdm, line 31:
```suggestion
```

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,525 @@
...
+* **Resolution Follows Scope Hierarchy:** Lookup begins relative to the documented element and proceeds outwards through the well-defined Dart scope precedence hierarchy (see [Section 5.2](#5.2.-scope-precedence-hierarchy)). The first valid match found in this search is used.
+
+* **Disambiguation via Qualification:** To prevent ambiguity or to reference an element from a distant scope, a reference should be qualified. This is done by prefixing the name with a class name (e.g., `[ClassName.memberName]`) or an import prefix (e.g., `[prefix.elementName]`).
+
+* **Handling Ambiguity:** If a reference could resolve to multiple declarations within the same scope, or if no resolution is found after checking all scopes, documentation tools should issue a warning.
+
+### **5.2. Scope Precedence Hierarchy**
+
+The resolution process for a reference \[name\] follows the standard Dart scope of the documented element with the extension of the doc imported scope at the end. Search is done in a specific order of precedence from the narrowest (most local) scope to the broadest (globally available).
```
sigurdm:
```suggestion
The resolution process for a reference `[name]` follows the standard Dart scope of the documented element with the extension of the doc imported scope at the end. Search is done in a specific order of precedence from the narrowest (most local) scope to the broadest (globally available).
```

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,525 @@
...
+
+This applies to doc comments on fields within a class, enum, or mixin. Since fields have no parameters, the search starts at the member level, but otherwise follows the same route as instance methods.
+
+* **Starting Scope**: Class Member Scope
+
+#### **5.3.3. Top-Level Functions**
+
+A top-level function operates like a method but without any enclosing class-like scopes.
+
+* **Starting Scope**:  Formal Parameter Scope
```
sigurdm, line 310: Perhaps explain top-level functions before methods, as they are simpler?

doc/dart_documentation_comment_specification.md
szakarias: done

doc/dart_documentation_comment_specification.md
szakarias: done

doc/dart_documentation_comment_specification.md
szakarias: done

doc/dart_documentation_comment_specification.md
szakarias: removed.

doc/dart_documentation_comment_specification.md
szakarias: I prefer the other one first and in more details, because it has more layers, and is like the diagram.

doc/dart_documentation_comment_specification.md
szakarias: Maybe later, I feel that doc is a bit incomplete compared to this one.

szakarias:
cc @johnniwinther @srawlins @bwilkerson @athomas @munificent

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,478 @@
...
+
+### **4.3. Members**
+
+* Methods (e.g., `[myMethod]`, `[MyClass.myMethod]`,` `[MyClass.myMethod()]`)
+* Fields (constants and variables) (e.g., `[myField]`, `[MyClass.myField]`)
+* Getters and Setters  (See [Section 6.2.2](#6.2.2.-getters-and-setters) for full details)
+* Constructors (e.g., `[MyClass.new]`, `[MyClass.named]`, `[MyClass.named()]`)
+* Enum constants (e.g., `[MyEnum.value]`)
+
+**4.4. Local Scope Parameters (within a member's doc comment):**
```
sgrekhov: Missing ##?

szakarias: Thanks!

sgrekhov:
The questions below are mainly questions about future versions of this spec.

1) If a declaration is [augmented](https://github.com/dart-lang/language/blob/main/working/augmentations/feature-specification.md) how to write doc comment in this case?
```dart 
/// Some doc comment about class C
class C {
  /// Some doc comments about the constructor
  C([int p]);
}

/// Another doc comment about class C
augment class C {
  /// Comment about the constructor
  augment C([int p = 0]) {
  // Augmentation added the defaulf value and the body
  }
}
```

2) Is there a way to add doc coments for a [primary constructor](https://github.com/dart-lang/language/blob/main/accepted/future-releases/primary-constructors/feature-specification.md)?

```dart
/// Some doc comment. Is it related to the class or to the constructor?
class C(var int x);
```

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,478 @@
...
+* The rules governing where documentation comments can be *placed*.
+* The mechanism for *resolving* references between elements.
+* The logic for *associating* documentation with code elements.
+
+**Out of Scope**
+
+This document does not cover:
+
+* The implementation details of specific documentation tools (e.g., `dartdoc`, `analysis server`).
+* The syntax and behavior of tool-specific directives (e.g., `@template`, `@canonicalFor`).
```
srawlins, line 28: Would you consider these in-scope in a later version? It would be very beneficial for DAS and dartdoc to agree on how these function. And new ones can be specified like `@snippet`, `@sample`.

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,478 @@
...
+* Any details related to the visual styling or HTML layout of generated documentation.
+* Best practices or style guides for writing effective documentation content.
+
+### **1.3. Terminology**
+
+* **Doc Comment:** A comment intended to be processed by documentation generation tools, like  dartdoc.
+* **Element**: A specific, declared element in the Dart code, such as a class, method, function, variable, or type parameter.
```
srawlins: Maybe worth it to go into this here, or later, but what about "induced" elements (that is the analyzer term; not sure if it is in the spec)?

So this would be `int x = 1;` inducing a getter and a setter, and `final x = 1;` inducing only a getter. Is one the element? Are both?

If a class A declares `int x = 1;` and then a subclass B declares `set x(int value) {}`, then the property `B.x` is declared half in A and half in B. If I link to `[B.x]`, does that link to the getter indueced by the property `A.x`? Etc. etc.

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,478 @@
...
+* **Identifier:** An individual name in the code (e.g., `MyClass`, myMethod, prefix).
+* **Qualified Name:** A name composed of two or more *identifiers* separated by dots, used to access an element within a specific namespace (e.g., MyClass.myMethod, prefix.MyClass).
+* **Reference:** A string enclosed in square brackets within a doc comment (e.g., `[MyClass]` or `[MyClass.myMethod]`) that links to an element. The text inside the brackets is either a single *identifier* or a *qualified name*.
+
+## **2\. Syntax of Documentation Comments**
+
+### **2.1. Line-Based Doc Comments (Recommended standard)**
+
+* Starts with `///`.
+* All consecutive lines starting with `///` are treated as part of the same comment block.
```
srawlins: Actually, I think even a comment block with a line starting with `// ` is considered one comment block (which does not include the non-doc comment text).

Not sure about `///` lines separated by a blank line...

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,478 @@
...
+```
+/**
+ *
+ *
+ */
+```
+
+### **2.3. Content Format (Markdown)**
+
+The text within a documentation comment block is parsed as Markdown, allowing for rich text formatting. This includes headings, lists, code blocks, and emphasis, which are converted for instance to HTML in the generated documentation.
```
srawlins: It's probably worth mentioning that they are parsed as CommonMark markdown, as "markdown" is not a well-specified syntax.

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,478 @@
...
+* Type aliases (Typedefs) (e.g., `[MyTypedef]`)
+
+### **4.2. Top-Level Declarations:**
+
+* Functions (e.g., `[myTopLevelFunction]`)
+* Variables and constants (e.g., `[myTopLevelVar]`)
+
+### **4.3. Members**
+
+* Methods (e.g., `[myMethod]`, `[MyClass.myMethod]`,` `[MyClass.myMethod()]`)
```
srawlins: There is a stray backtick after the 2nd example.

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,478 @@
...
+* Fields (constants and variables) (e.g., `[myField]`, `[MyClass.myField]`)
+* Getters and Setters  (See [Section 6.2.2](#6.2.2.-getters-and-setters) for full details)
+* Constructors (e.g., `[MyClass.new]`, `[MyClass.named]`, `[MyClass.named()]`)
+* Enum constants (e.g., `[MyEnum.value]`)
+
+### **4.4. Local Scope Parameters (within a member's doc comment):**
+
+* Parameters of the documented method/function (e.g., `[parameterName]`)
+* Type parameters of the documented element (e.g., `[T]`)
```
srawlins: I think also type parameters of the enclosing element, right? Can be referenced with `[T]`?

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,478 @@
...
+| | | | +----------------------------------------------------------+ | | | |
+| | | +--------------------------------------------------------------+ | | |
+| | +------------------------------------------------------------------+ | |
+| +----------------------------------------------------------------------+ |
++--------------------------------------------------------------------------+
+
+```
+### **5.3. Detailed Lookup Process**
+
+The lookup process begins at a specific "starting scope"  determined by the context of the doc comment and then follows the scope hierarchy.
```
srawlins: nit: two spaces after the word "scope"

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,478 @@
...
+
+#### **5.3.3. Top-Level Functions**
+
+A top-level function operates like a method but without any enclosing class-like scopes.
+
+* **Starting Scope**:  Formal Parameter Scope
+
+#### **5.3.4. Top-Level Variables**
+
+For top-level variables, the search has no local context and must begin at the file's library scope.
```
srawlins, line 314: It sounds silly, but can a doc comment on a top-level variable refer to itself? (Or can a doc comment on any element refer to itself?)

I ask because it's not too uncommon. You might see "Instead of calling [thisFunction], consider calling [somethingElse]."

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,478 @@
...
+
+**Namespace-Providing Elements**
+
+If an Identifier resolves to one of the following elements, it establishes a new namespace for resolving the next identifier in the chain.
+
+*Case 1: Import Prefix*
+
+* **Namespace:** The export scope of the imported library, as filtered by any show or hide combinators on the import directive.
+* **Example:** In `[math.pi]`, the identifier `math` resolves to an import prefix (e.g., from `import dart:math' as math;`). The tool then searches the public namespace of dart:math for the identifier pi.
+* **Combinator Example:** If the import was import 'dart:math' as math show sin;, the namespace for math would *only* contain sin. A reference to `[math.pi]` would fail to resolve, as `pi` was not included in the show list.
```
srawlins: It would be nice if the code here was in backticks.

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,478 @@
...
+  * Static Members
+  * Constructors
+
+* **Notes on Generics:**
+  * The namespace does not include the element's own type parameters. For a class `MyClass<T>`, a reference like `[MyClass.T]` is invalid because T is not a member of `MyClass`'s namespace.
+  * References are also made to the generic declaration, not a specific instantiation (e.g., write `[List.filled]`, not `[List<int>.filled]`).
+* **Example:** To resolve the reference `[collection.BoolList.empty]`:
+  * The identifier collection resolves to an import prefix.
+  * The identifier BoolList resolves to a class element within the collection library's public namespace.
+  * The identifier empty resolves to a named constructor element within the BoolList class's member namespace.
```
srawlins: It would be nice if code like 'empty' was in backticks here.

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,478 @@
...
+
+In some cases, the effective documentation for an element is determined by context, inheritance, or combination rules rather than by a direct preceding comment.
+
+#### **6.2.1. Overridden Members**
+
+When a member overrides an ancestor, its documentation is determined by a set of inheritance and precedence rules.
+
+* **Explicit Documentation:** If an overriding member does have its own doc comment, that comment takes full precedence. The documentation from any base members is ignored.
+
+* **Unambiguous Documentation Inheritance** If a member overrides a member from an ancestor and does not have its own doc comment, it inherits the documentation from the base member. This rule applies when the source of the documentation is unambiguous (e.g., overriding a single concrete method).  Documentation tools should display this inherited comment, ideally noting its origin (e.g., "Copied from BaseClass").
```
srawlins: It's interesting to use the 'concrete' attribute here. I don't think the abstractness/concreteness of any method is used anywhere else in this doc; it probably does not come into play here.

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,478 @@
...
+
+* **Ambiguous Documentation from Multiple Ancestors** The behavior for resolving documentation when a member inherits conflicting documentation from multiple ancestors is *reserved for future standardization*. Currently, tools may handle this case at their discretion, or ignore inherited documentation in ambiguous scenarios. It is recommended that tools *issue a warning* when this scenario is encountered, prompting the user to provide explicit documentation to resolve the ambiguity.
+
+#### **6.2.2. Getters and Setters**
+
+A getter and a setter that share the same name are treated as a single *conceptual property* rather than the individual getter or setter.
+
+Documentation tools should handle this property with the following rules:
+
+* **Display:** The getter and setter should be presented as a single item in the final documentation.
```
srawlins: Interesting, this is _not_ how DAS works today.

If you write `A().x = 1`, then the hover text in the IDE for `x =` is the doc comment for the setter, `set x`.

I think this is a good UX, and maybe this rule should instead be left to the discretion of the tool.

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,478 @@
...
+*Case 1: Import Prefix*
+
+* **Namespace:** The export scope of the imported library, as filtered by any show or hide combinators on the import directive.
+* **Example:** In `[math.pi]`, the identifier `math` resolves to an import prefix (e.g., from `import dart:math' as math;`). The tool then searches the public namespace of dart:math for the identifier pi.
+* **Combinator Example:** If the import was import 'dart:math' as math show sin;, the namespace for math would *only* contain sin. A reference to `[math.pi]` would fail to resolve, as `pi` was not included in the show list.
+
+*Case 2: Class-like top-level declaration*
+
+* **Namespace:** The set of all members declared within that element, including:
+  * Instance Members (fields, methods, etc.)
```
srawlins: I'm not sure if it is mentioned anywhere, but what about inheritance? So if `B` extends `A`, and does not re-implement a method, `A.foo`, then can we refer to the method on `B` with `B.foo`? `foo` is not declared on `B`.

doc/dart_documentation_comment_specification.md
srawlins: I think a missing colon in `Inheritance**`

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,478 @@
...
+
+* **Explicit Documentation:** If an overriding member does have its own doc comment, that comment takes full precedence. The documentation from any base members is ignored.
+
+* **Unambiguous Documentation Inheritance** If a member overrides a member from an ancestor and does not have its own doc comment, it inherits the documentation from the base member. This rule applies when the source of the documentation is unambiguous (e.g., overriding a single concrete method).  Documentation tools should display this inherited comment, ideally noting its origin (e.g., "Copied from BaseClass").
+
+* **Ambiguous Documentation from Multiple Ancestors** The behavior for resolving documentation when a member inherits conflicting documentation from multiple ancestors is *reserved for future standardization*. Currently, tools may handle this case at their discretion, or ignore inherited documentation in ambiguous scenarios. It is recommended that tools *issue a warning* when this scenario is encountered, prompting the user to provide explicit documentation to resolve the ambiguity.
```
srawlins: I think a missing colon in `Ancestors**`.

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,478 @@
...
+
+Documentation tools should handle this property with the following rules:
+
+* **Display:** The getter and setter should be presented as a single item in the final documentation.
+* **Precedence:**
+  * Documentation for the property should only be placed on the getter.
+  * The tooling should *issue a warning* if both a getter and its corresponding setter have unique doc comments.
+  * The only situation the documentation of the setter is considered is in the case of no getter existing.
+* **Reference:** A reference to a property name (e.g., `[value]`) resolves to the conceptual property rather than the individual getter or setter. This applies in all cases where a getter and setter share the same name, including:
+  * When an explicit getter (get value) and setter (set value(...)) are declared.
```
srawlins: I think the code samples here should be in backticks.

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,478 @@
...
+#### **6.2.2. Getters and Setters**
+
+A getter and a setter that share the same name are treated as a single *conceptual property* rather than the individual getter or setter.
+
+Documentation tools should handle this property with the following rules:
+
+* **Display:** The getter and setter should be presented as a single item in the final documentation.
+* **Precedence:**
+  * Documentation for the property should only be placed on the getter.
+  * The tooling should *issue a warning* if both a getter and its corresponding setter have unique doc comments.
```
srawlins, line 541: How does inheritance factor in here?

If I override a documented field with a setter, can I document the setter?

If a supertype has a documented setter, and an undocumented getter, and I override the getter, can I document the overriding getter (thus making both the getter and setter documented)?

doc/dart_documentation_comment_specification.md
munificent: It would be good to specify here how the comment text is extracted from these lines like you say for 2.2. In particular, that the leading `///` and one subsequent space is removed.

doc/dart_documentation_comment_specification.md
szakarias: done

doc/dart_documentation_comment_specification.md
szakarias: done

doc/dart_documentation_comment_specification.md
szakarias: Yes!

doc/dart_documentation_comment_specification.md
szakarias: done

doc/dart_documentation_comment_specification.md
szakarias: done

doc/dart_documentation_comment_specification.md
szakarias: done

doc/dart_documentation_comment_specification.md
szakarias: We definitely could! Or we can clean up `directives.md` and add new ones there.

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,525 @@
...
+* Top-level variables
+* Top-level getters or setters (See [Section 6.2.2](#6.2.2.-getters-and-setters) for details)
+
+### **3.3. Member Declarations**
+
+* Constructors (including factory and named constructors)
+* Methods (instance, static)
+* Operators
+* Fields (instance, static)
+* Getters or setters (Not both, prefer getter)
```
johnniwinther: Instead of `(Not both, prefer getter)` maybe just refer to Section 6.2.2 as above.

doc/dart_documentation_comment_specification.md
szakarias: done

doc/dart_documentation_comment_specification.md
szakarias: Yeah you are right. Changed the wording.

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,525 @@
...
+
+###
+
+### **2.3. Content Format (Markdown)**
+
+The text within a documentation comment block is parsed as Markdown, allowing for rich text formatting. This includes headings, lists, code blocks, and emphasis, which are converted for instance to HTML in the generated documentation.
+
+### **2.4. References**
+
+A reference is a special directive within the Markdown content that creates a hyperlink to a Dart element. It is written by enclosing a Dart identifier in square brackets (e.g., `[foo]`).  See [Section 4](#4.-referenceable-elements) for detailed information about which elements can be referenced.
```
johnniwinther: From this is sounds like only *identifiers* in square brackets are supported. I guess it could also be qualified names.

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,525 @@
...
+###
+
+### **4.2. Top-Level Declarations:**
+
+* Functions (e.g., `[myTopLevelFunction]`)
+* Variables and constants (e.g., `[myTopLevelVar]`)
+
+### **4.3. Members**
+
+* Methods (e.g., `[myMethod]`, `[MyClass.myMethod]`,` `[MyClass.myMethod()]`)
```
johnniwinther: The use of `()` within a reference is not mentioned in 2.4 and its meaning is not defined anywhere. Do we want/need this? Should we remove it from the spec, at least for now, pending a decision on the matter?

doc/dart_documentation_comment_specification.md
szakarias: I think a doc comment should be able to refer to itself and we already support that in dartdoc. Do you think this is unclear from the way it is formulated in the spec now?

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,525 @@
...
+* Any details related to the visual styling or HTML layout of generated documentation.
+* Best practices or style guides for writing effective documentation content.
+
+
+### **1.3. Terminology**
+
+* **Doc Comment:** A comment intended to be processed by documentation generation tools, like  dartdoc.
+* **Element**: A specific, declared element in the Dart code, such as a class, method, function, variable, or type parameter.
+* **Identifier:** An individual name in the code (e.g., `MyClass`, myMethod, prefix).
+* **Qualified Name:** A name composed of two or more *identifiers* separated by dots, used to access an element within a specific namespace (e.g., MyClass.myMethod, prefix.MyClass).
```
johnniwinther, line 37: Maybe **Name** should be defined as either an identifier or a qualified name and used in section 2.4 when describing the content of a reference. Name is already used elsewhere, seemingly with this meaning, for instance in the beginning of section 5.

doc/dart_documentation_comment_specification.md
johnniwinther: The resolution starts with the first *identifier* of a name.

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,525 @@
...
+
+  /// A method.
+  ///
+  /// Lookup examples:
+  /// * [param]: Resolves to the parameter (Formal Parameter Scope).
+  /// * [R]: Resolves to the method type parameter (Method Type Parameter Scope).
+  /// * [value]: Resolves to the class member (Class Member Scope).
+  /// * [myStaticMethod]: Resolves to the static class member (Class Member Scope).
+  /// * [T]: Resolves to the class type parameter (Class Type Parameter Scope).
+  /// * [AnotherClass]: Resolves to the library member (Library Scope).
```
johnniwinther, line 275: Maybe add a cases for `[myMethod]` and `[MyClass]` just to make it clear that parent declarations are in scope (and not just siblings to parents).

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,525 @@
...
+  /// * [List]: Resolves from 'dart:core' (Imported Scopes).
+  /// * [JsonCodec]: Resolves from 'dart:convert' (Imported Scope).
+  /// * [Random]: Resolves from 'dart:math' (Doc Imported Scope).
+  /// * [named] is not in scope (constructors are not in Class Member Scope)
+  void myMethod<R>(T param) {
```
johnniwinther, line 280: Do we support references to import prefixes as a way to reference libraries?

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,525 @@
...
+/// @docImport 'dart:math';
+
+/// A top-level function.
+///
+/// Lookup examples:
+/// * [input]: Resolves to the parameter (Formal Parameter Scope).
+/// * [R]: Resolves to the function type parameter (Function Type Parameter Scope).
+/// * [globalConstant]: Resolves to the library member (Library Scope).
+/// * [String]: Resolves from 'dart:core'(Imported Scope).
+R topLevelFunction<R>(String input) {
```
johnniwinther, line 336: Should this example be moved to 5.3.3 ?

johnniwinther, line 336: Add `[topLevelFunction]` to the example.

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,525 @@
...
+  // ...
+}
+
+/// A top-level variable.
+///
+/// Lookup examples:
+/// * [topLevelFunction]: Resolves to the library member (Library Scope).
+/// * [Duration]: Resolves from 'dart:core' (Imported Scope).
+/// * [Random]: Resolves from 'dart:math' (Doc Imported Scope).
+const int globalConstant = 10;
```
johnniwinther, line 347: Should this example be moved to 5.3.4 ?

johnniwinther, line 347: Add `[globalConstant]` to the example.

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,525 @@
...
+
+/// A class declaration.
+///
+/// Lookup examples:
+/// * [myMethod]: Resolves to the class member (Class Member Scope).
+/// * [T]: Resolves to the class type parameter (Class Type Parameter Scope).
+/// * [topLevelFunction]: Resolves to the library member (Library Scope).
+/// * [List]: Resolves from 'dart:core' (Imported Scopes).
+/// * [Random]: Resolves from 'dart:math' (Doc Imported Scope).
+class MyClass<T> {
```
johnniwinther, line 358: Add `[MyClass]` to the example.

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,525 @@
...
+
+```
+/// @docImport 'dart:math';
+
+class AnotherClass {}
+
+/// A typedef for a Map.
+///
+/// Lookup examples:
+/// * [T]: Resolves to the typedef type parameter (typedef type parameter scope).
```
johnniwinther: Use capital initials, "Typedef Type Parameter Scope", as in other examples.

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,525 @@
...
+/// * [AnotherClass]: Resolves to the library member (Library Scope).
+/// * [Map]: Resolves from 'dart:core' (Imported Scope).
+/// * [String]: Resolves from 'dart:core' (Imported Scope).
+/// * [Random]: Resolves from 'dart:math' (Doc Imported Scope).
+typedef JsonMap<T> = Map<String, T>;
```
johnniwinther, line 392: Add `[JsonMap]` to the example.

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,525 @@
...
+```
+class A {
+  /// A named constructor.
+  A.foo();
+
+  /// An instance method.
+  void foo() {}
+}
+
+/// Usage in documentation:
```
johnniwinther, line 499: Maybe move this inside the class since `[foo]` is only valid here.

doc/dart_documentation_comment_specification.md
johnniwinther: I disagree. The hover should know that it is a setter, but display the singular documentation for the property, and that documentation should cover both is read and write aspect. This would also avoid having documentation that isn't shown; if you painstakingly write documentation on the setter it would be unfortunate that is *only* shown when you hover and not elsewhere.

As a side note, DAS is a bit inconsistent with this. For instance if you ask for usage on a setter, you still get all the results that read the property, indicating that it *is* using the *conceptual property* in this case.

doc/dart_documentation_comment_specification.md
szakarias: done

doc/dart_documentation_comment_specification.md
szakarias: Yes, you are right. Updated.

doc/dart_documentation_comment_specification.md
szakarias: We can remove it for now, I'll make a note that we need to decide on this.

doc/dart_documentation_comment_specification.md
szakarias: done

doc/dart_documentation_comment_specification.md
szakarias: done.

doc/dart_documentation_comment_specification.md
szakarias: done.

doc/dart_documentation_comment_specification.md
szakarias: I think we want to, but not 100% decided. It is tracked here https://github.com/dart-lang/dartdoc/issues/1621

doc/dart_documentation_comment_specification.md
szakarias: I split out and updated the example.

szakarias: Actually I prefer keeping it in one 'top-level' example so we don't need to have all the top-level declarations copied in each example.

doc/dart_documentation_comment_specification.md
szakarias: done

doc/dart_documentation_comment_specification.md
szakarias: done

doc/dart_documentation_comment_specification.md
szakarias: done

doc/dart_documentation_comment_specification.md
szakarias: done

doc/dart_documentation_comment_specification.md
szakarias: Updated all the examples, hopefully it should be clear now.

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,478 @@
...
+
+### **2.1. Line-Based Doc Comments (Recommended standard)**
+
+* Starts with `///`.
+* All consecutive lines starting with `///` are treated as part of the same comment block.
+
+### **2.2. Block-Based Doc Comments (Historic, not recommended)**
+
+* Starts with `/**` and ends with the matching `*/`.
+* Block comments can be nested (e.g., `\*\* \* inner */ */`). Documentation tools must respect this, ensuring that the comment block only ends at the closing  \*/ that matches the opening delimiter.
```
lrhn: `\` -> `/` in the example.

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,478 @@
...
+* The content within the delimiters is treated as the documentation.
+* Leading whitespace followed by a single asterisk and a single following space on subsequent lines are considered stylistic and are stripped by doc generators.
```
lrhn: (What if there is a `*` that is not followed by a space? Is the `*` removed or not?
Particularly relevant for empty comment lines like:
````dart
/**
 * ```dart
 * const x = """
 *
 * """;
 * assert(x == '\n');
 * ```
 */
````
Here the empty line has no trailing space, and it would be wrong to include the `*`. )

File: doc/dart_documentation_comment_specification.md (hunk 10 above)
Comment (lrhn) [inline doc/dart_documentation_comment_specification.md]: The block-doc below says that the single space following the `*` is omitted. Is a single space following `///` also omitted?

(It probably is. That would be consistent, but then it should also be stated consistently. The only place where it may make a difference is withing code-fences, where the leading space becomes visible as indentation.)

File: doc/dart_documentation_comment_specification.md [hunk 38]
```diff
@@ -0,0 +1,478 @@
...
+* The implementation details of specific documentation tools (e.g., `dartdoc`, `analysis server`).
+* The syntax and behavior of tool-specific directives (e.g., `@template`, `@canonicalFor`).
+* Any details related to the visual styling or HTML layout of generated documentation.
+* Best practices or style guides for writing effective documentation content.
+
+### **1.3. Terminology**
+
+* **Doc Comment:** A comment intended to be processed by documentation generation tools, like  dartdoc.
+* **Element**: A specific, declared element in the Dart code, such as a class, method, function, variable, or type parameter.
+* **Identifier:** An individual name in the code (e.g., `MyClass`, myMethod, prefix).
```
lrhn: Code-quote `myMethod` too, probably also `prefix`.

doc/dart_documentation_comment_specification.md
lrhn: (For the record, "MarkDown" is also specified, https://daringfireball.net/projects/markdown/, it's just that nobody actually uses that specification. That's why CommonMark was created.)

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,478 @@
...
+
+### **2.3. Content Format (Markdown)**
+
+The text within a documentation comment block is parsed as Markdown, allowing for rich text formatting. This includes headings, lists, code blocks, and emphasis, which are converted for instance to HTML in the generated documentation.
+
+### **2.4. References**
+
+A reference is a special directive within the Markdown content that creates a hyperlink to a Dart element. It is written by enclosing a Dart identifier in square brackets (e.g., `[foo]`).  See [Section 4](#4.-referenceable-elements) for detailed information about which elements can be referenced.
+
+Conceptually, these behave like [reference-style links](https://www.markdownguide.org/basic-syntax/#reference-style-links) in Markdown. The documentation generator resolves the name against the available source code to create the link's destination. See [Section 5](#5.-reference-lookup-and-resolution) for detailed resolution rules.
```
lrhn, line 113: Worth saying that a CommonMark links, or other valid CommonMark uses of square brackets, are not parsed as a DartDoc reference:
* `[Banana](https://example.com "Bananas are fun")` - A link, not to the Banana class.
* `An image` - An image link.
* `[Link with reference][banana]` + `[banana]: https://example.com "Bananas are fun"
* `[^footnote]` + `[^footnote]: This note is afoot.`


CommonMark allows a stand-alone `[text]` that is not preceded or followed by something specific, as a [shorthand link](https://spec.commonmark.org/0.25/#shortcut-reference-link) equivalent to `[text][]`, which is again equivalent to `[text][text]`. That's the syntax that DartDoc takes over and gives a new meaning to.

Or rather, if the text inside `[...]` is not defined as a reference by the document, then that text has no meaning to CommonMark, and DartDoc can use it for its own. (We should consider doing the same thing for a `[some text][MyClass.someMethod]` if possible.)

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,478 @@
...
+* Fields (instance, static)
+* Getters or setters (Not both, prefer getter)
+
+### **3.4. Enum Constants**
+
+* A doc comment can be placed before an individual enum value.
+
+### **3.5. Invalid Placements**
+
+While not strictly disallowed by the language, any other placement of a comment with the /// syntax, is not considered a doc comment, and hence should be ignored by documentation tools.
```
lrhn, line 158: Do state explicitly whether metadata is considered part of the declaration. That is, can you do:
```dart
@override
/// This function overrides!
void foo() {}
```
?
(I'm guessing "no", the DartDoc must precede any metadata before the declaration.)

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,478 @@
...
+
+### **1.2. Scope**
+
+**In Scope**
+
+This specification covers the following topics:
+
+* The *syntax* for writing documentation comments.
+* The rules governing where documentation comments can be *placed*.
+* The mechanism for *resolving* references between elements.
```
lrhn, line 20: What is an "element"?

doc/dart_documentation_comment_specification.md
lrhn: (Cyclic definition: An element is a (specific declared) element. Then some examples, but one has to guess what they have in common.
The concept of "element" is an implementation detail in our tools, it's not a term that the language specification uses. It's not a word we can assume a reader will understand by themselves, so if we want to use it here, it needs to be defined, non-circularly. Or just use "declaration" or maybe "named declaration".)

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,478 @@
...
+## **4\. Referenceable Elements**
+
+A reference in a doc comment (e.g., `[name]`) can link to any Dart element that is visible from the Dart scope of the documented element. See [Section 5](#5.-reference-lookup-and-resolution) for more details about scoping.   This includes:
+
+### **4.1. Types**
+
+* Classes (e.g., `[MyClass]`)
+* Mixins (e.g., `[MyMixin]`)
+* Enums (e.g., `[MyEnum]`)
+* Named extensions (e.g., `[MyExtension]`)
```
lrhn, line 179: Nit: Not a type. Another reason to just use "top-level declarations". 😉

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,478 @@
...
+
+### **3.5. Invalid Placements**
+
+While not strictly disallowed by the language, any other placement of a comment with the /// syntax, is not considered a doc comment, and hence should be ignored by documentation tools.
+
+## **4\. Referenceable Elements**
+
+A reference in a doc comment (e.g., `[name]`) can link to any Dart element that is visible from the Dart scope of the documented element. See [Section 5](#5.-reference-lookup-and-resolution) for more details about scoping.   This includes:
+
+### **4.1. Types**
```
lrhn: These are also top-level declarations. In 3.2 they were included in the "top-level declarations" section. 
Any reason to put them by themselves here? ("Type declarations" would be more consistent with the next section's name.)

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,478 @@
...
+* Mixins (e.g., `[MyMixin]`)
+* Enums (e.g., `[MyEnum]`)
+* Named extensions (e.g., `[MyExtension]`)
+* Extension types (e.g., `[MyExtensionType]`)
+* Type aliases (Typedefs) (e.g., `[MyTypedef]`)
+
+### **4.2. Top-Level Declarations:**
+
+* Functions (e.g., `[myTopLevelFunction]`)
+* Variables and constants (e.g., `[myTopLevelVar]`)
```
lrhn, line 183: Getters and setters?

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,478 @@
...
+
+### **4.3. Members**
+
+* Methods (e.g., `[myMethod]`, `[MyClass.myMethod]`,` `[MyClass.myMethod()]`)
+* Fields (constants and variables) (e.g., `[myField]`, `[MyClass.myField]`)
+* Getters and Setters  (See [Section 6.2.2](#6.2.2.-getters-and-setters) for full details)
+* Constructors (e.g., `[MyClass.new]`, `[MyClass.named]`, `[MyClass.named()]`)
+* Enum constants (e.g., `[MyEnum.value]`)
```
lrhn, line 192: Maybe worth mentioning that there is no syntactic distinction between references to static and instance members.

(There is also an ambiguity, since `[Class.name]` can refer both to a constructor and an instance method, and a class declaration can have *both*. Need to say how it's disambiguated. Prefix `new` may still work.)

doc/dart_documentation_comment_specification.md
lrhn: And type parameters of surrounding declaration.
```dart
class ValueHolder<T> {
   final T value;
   ValueHolder(this.value);
   /// Converts a [T] to an [R] using [function].
   R callWith<R>(R Function(T) function) => function(value);
}
```

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,478 @@
...
+
+### **4.4. Local Scope Parameters (within a member's doc comment):**
+
+* Parameters of the documented method/function (e.g., `[parameterName]`)
+* Type parameters of the documented element (e.g., `[T]`)
+
```
lrhn, line 200: (Could this section simply say "any declaration whose name is in scope other than import prefixes"? Plus qualified names where the first identifier must denote "any declaration in scope".)

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,478 @@
...
+When a name is enclosed in square brackets (e.g., `[MyClass.myMethod]`), documentation tools attempt to resolve it to a specific API member. This section details the principles of that resolution process, the scope hierarchy it follows, and how to handle special cases.
+
+### **5.1. General Principles**
+
+* **Resolution Follows Scope Hierarchy:** Lookup begins relative to the documented element and proceeds outwards through the well-defined Dart scope precedence hierarchy (see [Section 5.2](#5.2.-scope-precedence-hierarchy)). The first valid match found in this search is used.
+
+* **Disambiguation via Qualification:** To prevent ambiguity or to reference an element from a distant scope, a reference should be qualified. This is done by prefixing the name with a class name (e.g., `[ClassName.memberName]`) or an import prefix (e.g., `[prefix.elementName]`).
+
+* **Handling Ambiguity:** If a reference could resolve to multiple declarations within the same scope, or if no resolution is found after checking all scopes, documentation tools should issue a warning.
```
lrhn: And resolve to one of them, or refuse to resolve?

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,478 @@
...
+The resolution process for a reference `[name]` follows the standard Dart scope of the documented element with the extension of the doc imported scope at the end. Search is done in a specific order of precedence from the narrowest (most local) scope to the broadest (globally available).
+
+The hierarchy is searched from the inside out. Below is an example for an instance method:
+```
++--------------------------------------------------------------------------+
+| 7. Doc Imported Scope (documentation-only imports with @docImport )      |
+| +----------------------------------------------------------------------+ |
+| | 6. Imported Scopes (all 'import' directives + implicit 'dart:core')  | |
+| | +------------------------------------------------------------------+ | |
+| | | 5. Library Scope (all other declarations in the file)            | | |
```
lrhn: (Including prefixes.)

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,478 @@
...
+| | | +--------------------------------------------------------------+ | | |
+| | | | 4. Class Type Parameter Scope (e.g., <T>)                    | | | |
+| | | | +----------------------------------------------------------+ | | | |
+| | | | | 3. Class Member Scope (e.g., static/instance members)    | | | | |
+| | | | | +------------------------------------------------------+ | | | | |
+| | | | | | 2. Method Type Parameter Scope (e.g., <R>)           | | | | | |
+| | | | | | +--------------------------------------------------+ | | | | | |
+| | | | | | | 1. Formal Parameter Scope (e.g., parameter names)| | | | | | |
```
lrhn, line 232: (And for initializing constructors, use the initializer list scope instead, since that contains the names of initializing formals and super parameters too.)

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,478 @@
...
+```
+### **5.3. Detailed Lookup Process**
+
+The lookup process begins at a specific "starting scope"  determined by the context of the doc comment and then follows the scope hierarchy.
+
+The following sections detail the starting scope for each type of declaration.
+
+#### **5.3.1. Instance Methods and Constructors**
+
+This applies to doc comments on methods, constructors, and operators within a class, enum, mixin, extension, or extension type. The search begins with the method's own parameters.
```
lrhn, line 250: (But not setters. Maybe say that explicitly.
They have a parameter list too, but I can see that's inconsistent with treating them as properties. If I do:
```dart
class C {
   void cake() {}
   /// Is it [cake]?
   set bar(int cake) {}
}
```
it looks like DartDoc refuses to make `[cake]` a link. It probably resolves to the setter parameter, but there is no target for a setter parameter to link to. If that's the case, a setter doc *does* start at the parameter scope.
I guess that answers my "anything but import prefixes" question above in the negative. "Anything but import prefixes and setter parameters."?)

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,478 @@
...
+/// @docImport 'dart:math';
+
+import 'dart:convert';
+
+class AnotherClass {}
+
+class MyClass<T> {
+  T? value;
+
+  MyClass.named();
```
lrhn: (Consider:
```dart

  /// A non-redirecting generative constructor.
  ///
  /// Lookup examples:
  /// * [value]: Resolves to the parameter.
  /// * [other]: Resolves to the parameter.
  /// * [MyClass.value]: Resolves to the instance variable.
  /// ... (more?) ...
  MyClass.value(this.value, int other);
```
)

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,478 @@
...
+
+  static void myStaticMethod() {
+    // ...
+  }
+
+}
+
+```
+
+#### **5.3.2. Instance Fields**
```
lrhn: Same for static fields?

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,478 @@
...
+
+This applies to doc comments on fields within a class, enum, or mixin. Since fields have no parameters, the search starts at the member level, but otherwise follows the same route as instance methods.
+
+* **Starting Scope**: Class Member Scope
```
lrhn, line 304: (And getters and setters are treated like fields?)

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,478 @@
...
+
+* **Starting Scope**:  Formal Parameter Scope
+
+#### **5.3.4. Top-Level Variables**
+
+For top-level variables, the search has no local context and must begin at the file's library scope.
+
+* **Starting Scope**: Library Scope
+
+#### **5.3.5. Top-Level Declarations**
```
lrhn: The two prior sections are also top-level declarations.

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,478 @@
...
+  /// A class member.
+  void myMethod() {}
+
+}
+
+```
+
+#### **5.3.6. Typedefs**
+
+The starting scope for a typedef doc comment depends on what it is an alias for. For function and record types, the scope starts at the parameter/field level. For all other types, it starts at the typedef's own type parameter scope.
```
lrhn, line 368: (This may need to be accounted for in the definition of what an "element" is. Type aliases for structural types do not introduce any *declarations*, their field/parameter names are not what I would usually consider expect to be an "element".)

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,478 @@
...
+/// [fun] : Resolves to parameter
+/// [p] : is not in scope
+typedef F<T> = void Function(void Function<S>(void Function(T p)) fun);
+```
+
+**Note on Nested Scopes:** The lookup only applies to the *immediate* parameters or fields of the typedef. It does not extend into parameters of nested function types.
+
+### **5.4. Resolving Qualified Names**
+
+When a reference contains a qualified name (e.g., `[prefix.ClassName.member]`), the resolution process is an iterative, left-to-right evaluation of each Identifier.
```
lrhn: "identifier", lower-case.

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,478 @@
...
+The namespace available depends on the type of element the previous identifier resolved to. Below are the primary cases.
+
+**Namespace-Providing Elements**
+
+If an Identifier resolves to one of the following elements, it establishes a new namespace for resolving the next identifier in the chain.
+
+*Case 1: Import Prefix*
+
+* **Namespace:** The export scope of the imported library, as filtered by any show or hide combinators on the import directive.
+* **Example:** In `[math.pi]`, the identifier `math` resolves to an import prefix (e.g., from `import dart:math' as math;`). The tool then searches the public namespace of dart:math for the identifier pi.
```
lrhn: ... searches the prefixed import namespace for the identifier `pi`.

If you have multiple imports with the same prefix and/or `show`/`hide` modifiers, the public namespace of the imported library isn't the whole truth.

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,478 @@
...
+*Case 2: Class-like top-level declaration*
+
+* **Namespace:** The set of all members declared within that element, including:
+  * Instance Members (fields, methods, etc.)
+  * Static Members
+  * Constructors
+
+* **Notes on Generics:**
+  * The namespace does not include the element's own type parameters. For a class `MyClass<T>`, a reference like `[MyClass.T]` is invalid because T is not a member of `MyClass`'s namespace.
+  * References are also made to the generic declaration, not a specific instantiation (e.g., write `[List.filled]`, not `[List<int>.filled]`).
```
lrhn, line 451: Or else? 
Currently it seem like the type argument is ignored, but the linking works.

doc/dart_documentation_comment_specification.md
lrhn: Today `List.List.empty` also resolves to the `List.empty` constructor in the `List` namespace. (Which can be convenient?)

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,478 @@
...
+* **Namespace:** The export scope of the imported library, as filtered by any show or hide combinators on the import directive.
+* **Example:** In `[math.pi]`, the identifier `math` resolves to an import prefix (e.g., from `import dart:math' as math;`). The tool then searches the public namespace of dart:math for the identifier pi.
+* **Combinator Example:** If the import was import 'dart:math' as math show sin;, the namespace for math would *only* contain sin. A reference to `[math.pi]` would fail to resolve, as `pi` was not included in the show list.
+
+*Case 2: Class-like top-level declaration*
+
+* **Namespace:** The set of all members declared within that element, including:
+  * Instance Members (fields, methods, etc.)
+  * Static Members
+  * Constructors
```
lrhn, line 447: Also type-aliases (`typedef`) probably works like the type they alias, if they alias a specific type.

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,478 @@
...
+
+### **5.5. Disambiguation and Special Cases**
+
+* **Constructors**: To refer to a named constructor, the syntax is `[ClassName.constructorName]`. Referencing the default, unnamed constructor requires the special syntax `[ClassName.new]`.
+
+* **Constructor vs. Member Conflicts** A naming conflict can occur if a class declares a named constructor (e.g., `MyClass.foo`) and also contains an instance member with the same name (e.g., a method `foo()` or a field `foo`).
+  * **Unqualified References** (`[foo]`): There is no ambiguity. A reference to `[foo]` always resolves to the member. A named constructor cannot be referenced by its name alone, it is not in scope as a simple identifier.
+  * **Qualified References** (`[MyClass.foo]`): This syntax is ambiguous because it could validly refer to either the named constructor or the member (via the class namespace). In this specific conflict scenario, `[MyClass.foo]` resolves to the *named constructor*. It is recommended that documentation tools *issue a warning* when this shadowing occurs. To reference the member explicitly in this context, users should be advised to use the unqualified `[foo]` if within the class context.
```
lrhn, line 487: That's a change, the current behavior is to refer to the `foo` method (as far as I can see).

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,478 @@
...
+  * The tooling should *issue a warning* if both a getter and its corresponding setter have unique doc comments.
+  * The only situation the documentation of the setter is considered is in the case of no getter existing.
+* **Reference:** A reference to a property name (e.g., `[value]`) resolves to the conceptual property rather than the individual getter or setter. This applies in all cases where a getter and setter share the same name, including:
+  * When an explicit getter (get value) and setter (set value(...)) are declared.
+  * When a final field (which acts as an implicit getter) is paired with an explicit setter of the same name.
+  * When a getter and setter for the same property are brought into scope, even if they are imported from different libraries.
+
+#### **6.2.3. Parameters**
+
+Parameters of functions, methods, and constructors do not have their own preceding doc comments. They are documented within the doc comment of their enclosing element using square brackets `[p]`.
```
lrhn: "... and referenced using an element reference \`[p]\`."

szakarias: done

doc/dart_documentation_comment_specification.md
szakarias: Yes it is, I think the current behavior is wrong.

FMorschel: Yes. The current spec mentions being able to differentiate between the instance member and the constructor by adding `()` for specifying the constructor (see https://github.com/dart-lang/sdk/issues/47553).

doc/dart_documentation_comment_specification.md
DanTup: @srawlins is it CommonMark or GFM? I was just looking up the difference (aware that VS Code supports a lot of GitHub's constructs), and found https://github.com/dart-lang/dartdoc/pull/1557 which looks like GFM table support in dartdoc, but AFAICT that's not part of CommonMark.

srawlins: Ah you're correct. It is [GitHub-flavored Markdown](https://github.github.com/gfm/), an extension of CommonMark.

doc/dart_documentation_comment_specification.md
szakarias: It is covered in the terminology section, now with updated definition. I guess I could swap 1.2 and 1.3. It feels a bit backwards to define terms before explaining what the doc is about.

doc/dart_documentation_comment_specification.md
szakarias: done

doc/dart_documentation_comment_specification.md
szakarias: Good catch, I would like to keep element for this doc. Updated the definition.

kevmoo:
Did I miss anything about actual HTTPS style links? That's been a problem w/ the analyzer lint for a while!

doc/dart_documentation_comment_specification.md
```diff
@@ -0,0 +1,485 @@
...
+```
+/**
+ *
+ *
+ */
+```
+
+### **2.3. Content Format (Markdown)**
+
+The text within a documentation comment block is parsed as CommonMark markdown, allowing for rich text formatting. This includes headings, lists, code blocks, and emphasis, which are converted for instance to HTML in the generated documentation.
```
DanTup: @szakarias in case you missed the other comments (I can't find them now, I think they got collapsed somewhere when this was updated), this is actually GFM rather than CommonMark.

Edit: https://github.com/dart-lang/dartdoc/pull/4130#discussion_r2577521545

doc/dart_documentation_comment_specification.md
szakarias: Good catch. Yes, that is true. Both for blank lines and lines starting with `//`

szakarias: done

szakarias: Updated this section, hopefully it should be clear enough now.

doc/dart_documentation_comment_specification.md
szakarias: done

doc/dart_documentation_comment_specification.md
szakarias: Yes, updated the text.

doc/dart_documentation_comment_specification.md
szakarias: done

doc/dart_documentation_comment_specification.md
szakarias: Ah great thanks, I was looking for the other comment since I could see that you had left one!

doc/dart_documentation_comment_specification.md
szakarias: updated it to say GFM! Thanks!

doc/dart_documentation_comment_specification.md
szakarias: Actually it could be both places. Not that pretty...

doc/dart_documentation_comment_specification.md
szakarias: Collapsed the sections.

doc/dart_documentation_comment_specification.md
szakarias: done

doc/dart_documentation_comment_specification.md
szakarias: done

doc/dart_documentation_comment_specification.md
szakarias: Noted. 

The disambiguation part is handled further in Section 5.5

doc/dart_documentation_comment_specification.md
szakarias: done

doc/dart_documentation_comment_specification.md
szakarias: It should not resolve, like we would handle it in the code, clarified this in the text.

doc/dart_documentation_comment_specification.md
szakarias: I would say it applies to setters too. (Although we don't consider documentation above the setter - except in the case where no getter exists.)

doc/dart_documentation_comment_specification.md
szakarias: Added example for the constructor but this spec proposes that `[MyClass.value]` will resolve to the constructor and not the field.

doc/dart_documentation_comment_specification.md
szakarias: Can't we get away with still calling it formal parameter scope? I mean do you see any examples where we can refer to a name in the initializer list?

doc/dart_documentation_comment_specification.md
szakarias: Yes. It just says 'Fields' now.

doc/dart_documentation_comment_specification.md
szakarias: No, lookup-wise I would say like methods. But getters/setters are a bit of a special case, that's why there is a dedicated section for those (section 6.2.2)

doc/dart_documentation_comment_specification.md
szakarias: Yes, changed it to "class-like top-level declaration" as used also in section 5.4

doc/dart_documentation_comment_specification.md
szakarias: Yes, and then they fall under the "class-like" category, no?

doc/dart_documentation_comment_specification.md
szakarias: This is the core idea yes. I still think it makes sense to have this section with the examples of the references. Especially for those that have a bit of special syntax, like `[MyClass.new]` or if we decide to support references to methods with trailing parentheses.

doc/dart_documentation_comment_specification.md
szakarias: done

doc/dart_documentation_comment_specification.md
szakarias: Added it both to the element definition and to section 4.3.

doc/dart_documentation_comment_specification.md
szakarias: The linking works in dartdoc but not in the IDE. I think we should not support it and it is kind of a bug that we resolve it in dartdoc.

doc/dart_documentation_comment_specification.md
szakarias: > If I override a documented field with a setter, can I document the setter?


I would say no. The mental model here is that a property (whether defined via a field or a getter/setter pair) is a single semantic entity. The getter is the "anchor" for that entity's documentation.

If a class inherits a getter and overrides the setter, the property's documentation remains anchored to the getter. Documenting the overriding setter would probably have something to do with the new mutable behavior but the original documentation would still be relevant, so this new doc shouldn't replace the old. Otherwise we create scenario where the documentation is split with no single source of truth. To prevent ambiguity, tools should ignore documentation on the setter if a getter exists.

> If a supertype has a documented setter, and an undocumented getter, and I override the getter, can I document the overriding getter (thus making both the getter and setter documented)?


In this system, an undocumented getter takes precedence over a documented setter. The presence of a getter determines the documentation slot. If the getter is undocumented, the property is effectively undocumented.

doc/dart_documentation_comment_specification.md
szakarias: I agree that showing context-specific information in an IDE (like hover) is good UX. However, I distinguish between interactive context and defining the property's contract.

Having a single source of truth for the property's documentation makes it clear where to go look for the "main" description. I think tools like DAS are free to provide extra implementation-level details during interaction, but for the purpose of the API reference we should have the documentation unified and anchored to the getter.

doc/dart_documentation_comment_specification.md
szakarias: done

doc/dart_documentation_comment_specification.md
szakarias: done

doc/dart_documentation_comment_specification.md
szakarias: In my opinion this is a bug in dartdoc. The correct reference is `[List.empty]`. (And it also doesn't resolve in the IDE)

szakarias:
> The questions below are mainly questions about future versions of this spec.
> 
> 1. If a declaration is [augmented](https://github.com/dart-lang/language/blob/main/working/augmentations/feature-specification.md) how to write doc comment in this case?
> 
> ```dart
> /// Some doc comment about class C
> class C {
>   /// Some doc comments about the constructor
>   C([int p]);
> }
> 
> /// Another doc comment about class C
> augment class C {
>   /// Comment about the constructor
>   augment C([int p = 0]) {
>   // Augmentation added the defaulf value and the body
>   }
> }
> ```
>
> 2. Is there a way to add doc coments for a [primary constructor](https://github.com/dart-lang/language/blob/main/accepted/future-releases/primary-constructors/feature-specification.md)?
> 
> ```dart
> /// Some doc comment. Is it related to the class or to the constructor?
> class C(var int x);
> ```


The exact details for doc comments for augmentations and primary constructors are not completely decided. This spec will be updated once the design is final.

szakarias:
Thanks so much for all the questions and comments! I have now addressed all of them. PTAL @lrhn @sigurdm @srawlins @johnniwinther @bwilkerson

PR #4139, opened 2025-12-19
Fix 404 error page formatting and broken links

Franklyn-R-Silva:
This PR addresses the issues reported in dart-lang/sdk#62262 regarding the confusing 404 error page.

### Changes:
- **Removed inert search boxes:** Deleted the search form in the body and the `{{>search_sidebar}}` tag, as search scripts typically fail to load on 404 pages, leaving these boxes stuck in a "Loading..." state.
- **Fixed formatting:** Simplified the 404 template to provide a cleaner user experience.
- **Improved links:** Although I preserved the original text, I updated the structure to ensure the page remains functional.

### Testing:
- Manual verification: Generated documentation using `dart bin/dartdoc.dart` and verified the resulting `404error.html`. The inert elements are gone and the layout is much cleaner.
- Automated tests: Existing HTML generator tests were skipped locally as they are currently not supported on Windows, but the template changes were verified in the generated output.

Fixes https://github.com/dart-lang/sdk/issues/62262

### DIFF
--- a/lib/templates/404error.html
+++ b/lib/templates/404error.html
@@ -8,25 +8,11 @@
       You've tried to visit a page that doesn't exist. Luckily this site has
       other <a href="index.html">pages</a>.
     </p>
-    <div>
-      If you were looking for something specific, try searching:
-      <form class="search-body" role="search">
-        <input
-          type="text"
-          id="search-body"
-          autocomplete="off"
-          disabled
-          class="form-control typeahead"
-          placeholder="Loading search..."
-        />
-      </form>
-    </div>
   </section>
 </div>
 <!-- /.main-content -->
 
 <div id="dartdoc-sidebar-left" class="sidebar sidebar-offcanvas-left">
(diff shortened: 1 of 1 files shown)


szakarias (commented):
Thanks for the PR. 

I think for the example URL this still leaves many broken links on the page. Mabye a better short-term fix would be to only have one link to the front page of the latest version....

The more comprehensive fix would be to resolve all relative paths even when the base URL is broken.

Franklyn-R-Silva:
> Thanks for the PR.
> 
> I think for the example URL this still leaves many broken links on the page. Mabye a better short-term fix would be to only have one link to the front page of the latest version....
> 
> The more comprehensive fix would be to resolve all relative paths even when the base URL is broken.

In the future I think someone can implement this problem, but I'll see if I can do it for everyone

szakarias:
@Franklyn-R-Silva , are you planning to work on this anytime soon?

PR #4148, opened 2025-12-27
Deprecate unused useCategories option

### LINKED ISSUE #4046: Investigate `useCategories` option
It appears to me that `useCategories` defaults to `true`, and that nobody reads it.

```dart
    DartdocOptionArgOnly<bool>('useCategories', true, resourceProvider,
        help: 'Display categories in the sidebar of packages'),
```

```dart
   bool get useCategories => optionSet['useCategories'].valueAt(context);
```

I certainly can't find any references of it in the templates or repository at large.

Perhaps, we should:
 * Allow the option to be passed, but ignore it.
 * Print a warning that it is deprecated when the option is passed.
 * Remove `useCategories => optionSet['useCategories'].valueAt(context);`

nathannewyen:
Removes the unused `useCategories` getter and hides the option from help output.

The option is kept for backwards compatibility but marked as deprecated.

Fixes #4046

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 * Fix a bug where experiment flags passed via command-line were ignored.
 * Support specifying experimental features in `dartdoc_options.yaml`.
 * Document using the public name for private named parameters.
+* Deprecate the `--use-categories` option.
 
 ## 9.0.3
 * Require analyzer 10.1.0 APIs.


lib/src/dartdoc_options.dart
```diff
@@ -1548,7 +1546,8 @@ List<DartdocOption> createDartdocOptions(
       return packageMeta;
     }, resourceProvider, help: 'PackageMeta object for the default package.'),
     DartdocOptionArgOnly<bool>('useCategories', true, resourceProvider,
-        help: 'Display categories in the sidebar of packages'),
+        help: '(deprecated, ignored) Display categories in the sidebar of packages',
```
szakarias:
```suggestion
        help: 'Deprecated, has no effect.',
```

PR #4155, opened 2026-01-05
Remove `Library.sentinel`.

lrhn:
Use `null` instead of a sentinel value that throws on all member accessses.

That lets the type system help with avoiding spurious runtime errors
if forgetting to check for a sentinel. 
It makes it explicit, as a `.library!`, where code assumes that there is a valid library.

(Some small drive-by tweaks.)

### DIFF
--- a/lib/src/model/library.dart
+++ b/lib/src/model/library.dart
@@ -114,6 +114,7 @@ class Library extends ModelElement
       }
     }
 
+    if (name.isEmpty) return false; // Not how library privacy is determined!
     if (element.nonSynthetic.metadata.hasInternal) return false;
     if (element.hasPrivateName || hasNodoc) return false;
 
--- a/lib/src/model/model_element.dart
+++ b/lib/src/model/model_element.dart
@@ -542,7 +542,7 @@ abstract class ModelElement
     return self.enclosingElement == self.canonicalEnclosingContainer;
   }
 
-  /// The documentaion, stripped of its comment syntax, like `///` characters.
+  /// The documentation, stripped of its comment syntax, like `///` characters.
   @override
   String get documentation => injectMacros(
       documentationFrom.map((e) => e.documentationLocal).join('<p>'));


lib/src/element_type.dart
```diff
@@ -24,15 +24,15 @@ abstract class ElementType with CommentReferable, Nameable {
   @override
   final PackageGraph packageGraph;
   @override
-  final Library library;
+  final Library? library;
```
lrhn, line 27: I left `Library` as nullable anywhere a sentinel value could potentially flow (as well as I could deduce that, which isn't much - using `null` makes things easier to reason about).

lib/src/generator/generator_utils.dart
```diff
@@ -97,7 +97,7 @@ int _packageRank(List<String> packageOrder, ModelElement element) {
   var index = packageOrder.indexOf(packageName);
   if (index == -1) return packageOrder.length * 10;
   if (packageName == 'Dart' &&
-      !_dartCoreLibraries.contains(element.library.name)) {
+      !_dartCoreLibraries.contains(element.library!.name)) {
```
lrhn, line 100: Most of the places where a `!` was added was in this file, where an `element.library.property` would already have thrown if the value was a sentinel.
The rest are where calling constructors for classes that expect a non-`null` library value, ensuring that we only get "no library" where a sentinel was expected to be used.

lib/src/model/library.dart
```diff
@@ -488,7 +470,7 @@ class Library extends ModelElement
   late final Set<String> _allOriginalModelElementNames = () {
     // Instead of using `allModelElements`, which includes deeper elements like
     // methods on classes, gather up only the library's immediate members.
-    var libraryMembers = [
+    var libraryMembers = <ModelElement>[
```
lrhn, line 496: (Explicitly typed to avoid inferring `hasLibrary`, which will prevent `is! GetterSetterCombo` below from promoting.)

lib/src/model/model_element.dart
```diff
@@ -59,14 +59,28 @@ import 'package:path/path.dart' as p show Context;
 abstract class ModelElement
     with CommentReferable, Warnable, Nameable, SourceCode, DocumentationComment
     implements Comparable<ModelElement>, Documentable, HasLocation {
+  /// The [Library] of a model can be `null` in three cases:
+  ///
+  /// * the library for `dynamic` and `Never`.
+  /// * the library for type parameters. <!-- Is this true? -->
```
lrhn, line 66: It seems `Parameter` was also called without a library in some tests.

lib/src/model/model_element.dart
```diff
@@ -269,52 +283,52 @@ abstract class ModelElement
...
       MethodElement(isOperator: false)
           when e.enclosingElement is ExtensionElement =>
         Method.providedByExtension(
-            e, enclosingContainer, library, packageGraph),
+            e, enclosingContainer, library!, packageGraph),
       MethodElement(isOperator: false) => Method.inherited(
-          e, enclosingContainer, library, packageGraph,
+          e, enclosingContainer, library!, packageGraph,
           originalElement: originalMember as ExecutableElement?),
       FormalParameterElement() => Parameter(e, library, packageGraph,
```
lrhn, line 328: Accepts `null` library too. Was called with it in some tests.

It may just have been called with the sentinel value, and never *used it* for anything, but it means there is a path that creates such an element.

lib/src/model/model_element.dart
```diff
@@ -373,21 +387,24 @@ abstract class ModelElement
...
     if (name.isEmpty) {
       return false;
     }
     if (this is! Library) {
+      var library = this.library;
+      if (library == null) return false;
       final canonicalLibrary = this.canonicalLibrary;
       var isLibraryOrCanonicalLibraryPrivate = !library.isPublic &&
           (canonicalLibrary == null || !canonicalLibrary.isPublic);
-      if (library == Library.sentinel || isLibraryOrCanonicalLibraryPrivate) {
```
lrhn: The test was bugged. The code does `library.isPublic` above *before* testing for the sentinel, which means that if it was a sentinel value, it would throw before this test.

lib/src/model/model_element.dart
szakarias: Yes, I think under this design `Parameter`s are also allowed to have no library.

szakarias, line 66: > Is this true? 

I'm not sure if this is actually the case for type parameters. Regardless, I wasn't a fan of the sentinel pattern either, so I think it's a good idea to review the design around library now that we're cleaning this up.

srawlins (approved):
This looks great to me!

PR #4158, opened 2026-01-07
Handle private named parameters.

lrhn:
Uses the corresponding public name for private named parameters. (If the program is valid, there won't be any conflicts.)

@munificent 
I had a look at the code, and it looked "easy" to make it work with private-named-parameters.

(But feel free to ignore, may want to wait if a later analyzer version will expose the correct name for you directly.)

### DIFF
--- a/lib/src/model/parameter.dart
+++ b/lib/src/model/parameter.dart
@@ -2,7 +2,9 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/element/element.dart';
+import 'package:dartdoc/src/charcode.dart' show $z, $a, $dollar, $A;
 import 'package:dartdoc/src/element_type.dart';
 import 'package:dartdoc/src/model/comment_referable.dart';
 import 'package:dartdoc/src/model/kind.dart';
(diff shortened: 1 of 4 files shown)


lib/src/model/parameter.dart
```diff
@@ -78,6 +80,27 @@ class Parameter extends ModelElement with HasNoPage {
...
+    if (element.isNamed &&
+        name != null &&
+        Identifier.isPrivateName(name) &&
+        name.length > 1) {
+      // Check if the remainder is a valid non-private identifier.
+      var nextChar = name.codeUnitAt(1);
+      var nextCharLower = nextChar | ($a ^ $A);
+      if (nextCharLower >= $a && nextCharLower <= $z || nextChar == $dollar) {
+        name = name.substring(1);
+      }
```
munificent, line 99: Technically, this also needs to test whether the corresponding public name is not a reserved word (i.e. `_for` -> `for`). Fortunately, analyzer does all that logic for you now. When it creates a `FieldFormalParameterElement`, the `name` will be the corresponding public name if it comes from a private named parameter. I believe `lookupName` defaults to returning `name`, so will do the same thing.

test/parameters_test.dart
```diff
@@ -157,6 +157,27 @@ class A {
...
+        <span class="parameter" id="privateNamed-param-f">
+          <span class="type-annotation">
+            <a href=".*/dart-core/int-class\.html">int</a>\?
+          </span>
+          <span class="parameter-name">f</span>
+        </span>
+        \}
+      '''));
+  }
+
```
munificent, line 180: You'll also want to test references to private named parameters in doc comments.

munificent (commented):
Thanks for hacking on this!

test/parameters_test.dart
lrhn: ACK.

lib/src/model/parameter.dart
lrhn: I was hoping that some of this would be unnecessary if the analyzer already did the job.

szakarias:
Replaced by https://github.com/dart-lang/dartdoc/pull/4202

PR #4171, opened 2026-01-13
Remove PackageConfigProvider classes; simplify with one function

srawlins:
This is a giant cleanup, but it's quite mechanical, and easily explained :D

PackageConfigProvider was created with two concrete classes, one that used the file system to find PackageConfigs, and one that was totally fake, where Packages were manually added to the provider.

But we no longer need the Fake ones; this might stem all the way back from when we started using analyzer's ResourceProvider to test with in-memory file systems. So we can remove the hierarchy. And then we're left with the singular method that was defined, `findPackageConfig`. We can make this method a top-level function, and then tear-down the numerous places where PackageConfigProvider was instantiated, stored in fields, and passed around. So much deleted code.

### DIFF
--- a/lib/src/package_config_provider.dart
+++ b/lib/src/package_config_provider.dart
@@ -13,8 +13,6 @@ import 'package:package_config/package_config.dart' as package_config;
 /// See https://github.com/dart-lang/tools/issues/1536 for any progress on
 /// synchronous APIs in the package_config package.
 package_config.PackageConfig? findPackageConfig(Folder folder) {
-  //var directory = io.Directory(baseDirectory.path);
-  //if (!directory.isAbsolute) directory = directory.absolute;
   if (!folder.exists) return null;
 
   do {
@@ -35,7 +33,6 @@ package_config.PackageConfig? findPackageConfig(Folder folder) {
 /// with `checkForPackageConfigJsonFile` inlined.
 package_config.PackageConfig? findPackageConfigInDirectory(Folder folder) {
   var packageConfigFile =
-      // io.File(path.join(folder.path, '.dart_tool', 'package_config.json'));
       folder
           .getChildAssumingFolder('.dart_tool')
           .getChildAssumingFile('package_config.json');


lib/src/package_config_provider.dart
```diff
@@ -2,85 +2,51 @@
...
 ///
-/// This provides an abstraction around package_config, which can only work
-/// with the physical file system.
-abstract class PackageConfigProvider {
-  package_config.PackageConfig? findPackageConfig(Folder dir);
+/// See https://github.com/dart-lang/tools/issues/1536 for any progress on
+/// synchronous APIs in the package_config package.
+package_config.PackageConfig? findPackageConfig(Folder folder) {
+  //var directory = io.Directory(baseDirectory.path);
+  //if (!directory.isAbsolute) directory = directory.absolute;
```
szakarias: Should these two lines stay?

lib/src/package_config_provider.dart
```diff
@@ -2,85 +2,51 @@
...
-  /// The minimum "Package Config" version which we can parse.
-  static const _minVersion = 1;
+/// Finds a `.dart_tool/package_config.json` file in [folder].
+///
+/// This is a synchronous version of
+/// [`findPackageConfigInDirectory`](https://github.com/dart-lang/tools/blob/52cc9b5bb10e2293a094678815c2b4d13b1a3acf/pkgs/package_config/lib/src/discovery.dart#L119C24-L119C52).
+/// with `checkForPackageConfigJsonFile` inlined.
+package_config.PackageConfig? findPackageConfigInDirectory(Folder folder) {
+  var packageConfigFile =
+      // io.File(path.join(folder.path, '.dart_tool', 'package_config.json'));
```
szakarias: And this?

szakarias (approved):
Very cool!!

lib/src/package_config_provider.dart
srawlins: Yikes good catch.

lib/src/package_config_provider.dart
srawlins: Yikes another good catch 😅

Issue #4134, opened 2025-12-16
Add or update tests to check conformance to the specification


sgrekhov:
Add/update tests to check that the current implementation follows the [Dart Documentation Comments Specification](https://github.com/dart-lang/dartdoc/blob/main/doc/dart_documentation_comment_specification.md)

cc @szakarias @athomas  Please assign me as Assignee

PR #4172, opened 2026-01-14
#4134. Add doc comments tests for metadata

sgrekhov:


test/co19/annotations_test.dart
```diff
@@ -0,0 +1,271 @@
...
+/** 
+ * Line 3
+ */
+class C {}
+''');
+    var c = library.classes.named('C');
+    expect(c.documentation, 'Line 3');
+  }
+
+  /// Check that if there is an annotation in doc comments it is treated as part
```
szakarias:
```suggestion
  /// Check that if there is an annotation in a doc comment it is treated as part
```

test/co19/annotations_test.dart
```diff
@@ -0,0 +1,271 @@
...
+/// `@Meta` Line 2  
+class C {}
+''');
+    var c = library.classes.named('C');
+    expect(c.documentation, '''
+@Meta Line 1
+`@Meta` Line 2''');
+  }
+
+  /// Check that if there is an annotation in doc comments it is treated as part
```
szakarias:
```suggestion
  /// Check that if there is an annotation in a doc comment it is treated as part
```

szakarias (approved):
LGTM

sgrekhov:
Updated. PTAL.

PR #4174, opened 2026-01-15
Prepare 9.0.2 release

szakarias:
Migrate to analyzer 10.0.0

### DIFF
--- a/lib/src/model/accessor.dart
+++ b/lib/src/model/accessor.dart
@@ -31,9 +31,9 @@ class Accessor extends ModelElement with HasLibrary {
       {ExecutableElement? super.originalElement});
 
   @override
-  CharacterLocation? get characterLocation => !element.isOriginDeclaration
-      ? enclosingCombo.characterLocation
-      : super.characterLocation;
+  CharacterLocation? get characterLocation => element.isOriginDeclaration
+      ? super.characterLocation
+      : enclosingCombo.characterLocation;
 
   @override
   ExecutableElement? get originalMember =>


lib/src/model/accessor.dart
```diff
@@ -31,7 +31,7 @@ class Accessor extends ModelElement with HasLibrary {
       {ExecutableElement? super.originalElement});
 
   @override
-  CharacterLocation? get characterLocation => element.isSynthetic
+  CharacterLocation? get characterLocation => !element.isOriginDeclaration
       ? enclosingCombo.characterLocation
       : super.characterLocation;
```
sigurdm: Slight simplification
```suggestion
  CharacterLocation? get characterLocation => element.isOriginDeclaration
? super.characterLocation
: enclosingCombo.characterLocation;
```

lib/src/model/getter_setter_combo.dart
```diff
@@ -114,16 +115,21 @@ mixin GetterSetterCombo on ModelElement {
...
-    if (element.isSynthetic) {
-      if (hasExplicitGetter) return getter!.characterLocation;
-      if (hasExplicitSetter) return setter!.characterLocation;
-      assert(false, 'Field and accessors can not all be synthetic: $element');
+    if (hasExplicitGetter) {
+      return getter!.characterLocation;
     }
-    return super.characterLocation;
+    assert(hasExplicitSetter);
+    return setter!.characterLocation;
```
sigurdm, line 132: The null check is an implicit assertion. That should be enough.

```suggestion
    if (getter != null) {
      return getter.characterLocation;
    }
    return setter!.characterLocation;
```

It seems you removed the condition on `element.isSynthetic` is that intended? In that case I think the comment above needs to be adjusted.

lib/src/model/accessor.dart
szakarias: done

lib/src/model/getter_setter_combo.dart
szakarias: This is not quite right. There can be an implicit getter and in that case it is non-null.

sigurdm: If you are sure the logic is equivalent, then fine. I'm not sure I follow completely.

sigurdm: Ah, I see now, hasExplicitGetter is different from `getter != null`. Ok never mind - this should work!

PR #4179, opened 2026-01-20
Move all '@canonicalFor', '@category', and '@subCategory' processing into DocumentationComment

srawlins:
This is a bit of a doozy. In order to get rid of using `documentationLocal` for its side-effects, we have to get rid of the `buildDocumentationAddition` overrides; this is not so bad as there were only three:

* `Library` overrode `buildDocumentationAddition` only to maybe report some warnings, if there were bad `{@canonicalFor}` directives. So we just rename this to be `checkCanonicalForIsValid` and call from DocumentationComment.
* The only class that mixed in `CanonicalFor` was `Library`. So `Library.buildDocumentationAddition` calling `super.buildDocumentationAddition` was the only call site. This code becomes `DocumentationComment._stripCanonicalFor`.
* Many classes mixed in `Categorization` (see below), but it really just boils down to ModelElement. So this implementation moves to `DocumentationComment._stripAndSetDartdocCategories`.

Then, parsing all of the doc comment directives is located in one mixin, DocumentationComment. There are still side-effect issues, but they're all contained in this file now, and I think we're better positioned to clean it up after this change. There are also many late or nullable fields that are a little precarious. I have ideas for fixing these.

This led to some other cleanups:

* Remove all `buildDocumentationAddition` declarations.
* `_processCommentWithoutTools` only ever accepted the `documentationComment` field so... just remove the parameter.
* `_processCommentWithoutTools` did all the same things that `processCommentDirectives`, except for processing macros. So change `processCommentDirectives` to optionally process macros, and deduplicate the code.
* `_documentationLocalIsSet` and `_docsHaveBeenBuilt` were mostly redundant, revealed by the refactoring. So delete the former.
* Fold `_buildDocumentationBase` into its only caller, `precacheLocalDocs`.
* Fold `_buildDocumentationBaseSync` into its only caller, `documentationLocal`.
* `Container`, `Library`, `ModelFunction`, `TopLevelVariable` and `Typedef` all mixed in Categorization; no longer needed.
* Many `Categorization` type annotations are just `ModelElement` now.

### DIFF
--- a/lib/src/model/documentation_comment.dart
+++ b/lib/src/model/documentation_comment.dart
@@ -699,7 +699,7 @@ mixin DocumentationComment implements Warnable, SourceCode {
   /// `/// {@canonicalFor libname.ClassName}`
   String _stripCanonicalFor(String docs) {
     if (this is Library) {
-      docs = docs.replaceAll(_canonicalRegExp, '');
+      docs = docs.replaceAll(_canonicalForRegExp, '');
 
       (this as Library).checkCanonicalForIsValid();
     }
@@ -832,7 +832,6 @@ mixin DocumentationComment implements Warnable, SourceCode {
     return _subCategoryNames!;
   }
 
-  // todo change template
   bool get hasCategoryNames => categoryNames.isNotEmpty;
   List<String>? _categoryNames;
 
@@ -861,9 +860,10 @@ mixin DocumentationComment implements Warnable, SourceCode {
 
   /// The set of libraries which this [Library] is canonical for.
   late final Set<String> canonicalFor = {
-    for (var match in _canonicalRegExp.allMatches(documentationComment))
+    for (var match in _canonicalForRegExp.allMatches(documentationComment))
       match.group(1)!,
   };
 
-  static final _canonicalRegExp = RegExp(r'{@canonicalFor\s([^}]+)}');
+  static final _canonicalForRegExp =
+      RegExp(r'[ ]*{@canonicalFor\s([^}]+)}[ ]*\n?');
 }


lib/src/model/documentation_comment.dart
```diff
@@ -805,8 +789,81 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+  List<String>? _subCategoryNames;
+
+  /// A set of strings containing all declared subcategories for this element.
+  List<String> get subCategoryNames {
+    // TODO(srawlins): avoid side-effect dependency.
+    documentationLocal;
+    return _subCategoryNames!;
+  }
+
+  // todo change template
```
szakarias: Leftover todo?

lib/src/model/documentation_comment.dart
```diff
@@ -805,8 +789,81 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+    return categoryNames.isNotEmpty || subCategoryNames.isNotEmpty;
+  }
+
+  /// The set of libraries which this [Library] is canonical for.
+  late final Set<String> canonicalFor = {
+    for (var match in _canonicalRegExp.allMatches(documentationComment))
+      match.group(1)!,
+  };
+
+  static final _canonicalRegExp = RegExp(r'{@canonicalFor\s([^}]+)}');
```
szakarias: Should we consider making the two `RegExp`s more consistent, so that we handle spaces and newlines the same way? Maybe:

`RegExp(r'[ ]*{@(keyword)\s+([^}]+)}[ ]*\n?')`

szakarias (approved):
Love it!

lib/src/model/documentation_comment.dart
srawlins: Good catch; done.

lib/src/model/documentation_comment.dart
srawlins: Done

PR #4181, opened 2026-01-21
#4134. Add tests for block-comments placement

sgrekhov:


test/co19/placement_test.dart
```diff
@@ -191,8 +366,45 @@ late final int x; /// Line 2
...
+ */
+late final int x = 0; /** Line 2 */
+/** Line 3
+ */
+''');
+    var v = library.properties.named('x');
+    expect(v.documentation, 'Line 1');
+  }
+
+  /// Check that doc comments can be placed before a late final variable.
```
szakarias: The 'late final' test is above.

sgrekhov:
Thank you. Fixed. PTAL.

test/co19/placement_test.dart
```diff
@@ -191,8 +366,45 @@ late final int x; /// Line 2
     expect(v.documentation, 'Line 1');
   }
 
+  /// Check that doc comments can be placed before a variable.
```
szakarias: Maybe you want to update the doc here to say late final?

sgrekhov: Of course. Thank you. Updated.

PR #4190, opened 2026-01-28
Make documentationComment nullable

srawlins:
This one isn't a slam dunk, but I do think it is an improvement. Feedback welcome 😁 

I found it strange that `DocumentationComment.documentationComment` (forgiving the naming for now) returned the element's `documentationComment ?? ''`. Then we have lots of code that asks if that comment is empty. So we're replacing one sentinel value (`null`) for another (`''`), without any gain, I believe. In my mind it makes things weird as well, when a real String value might not be an actual documentation comment; it might be that `?? ''` fallback.

So this was an experiment to make `DocumentationComment.documentationComment` nullable. I think it reads much nicer; instead of many cases of `hasDocumentationComment &&`, we can use `documentationComment ??`.

This also opens up a very nice refactoring for `GetterSetterCombo._getterSetterDocumentationComment`. The previous code is pretty complex with many conditions and local variables slowly building to the answer of "does my getter have a doc comment? does my setter? do both?" In the refactoring, we gather up the getter portion of the comment (which may be a nice tidy `null`), and then the setter portion (again, may be `null`). And then we return one, the other, or a combination.

(Also, regarding `DocumentationComment.documentationComment` naming: the mixin should be called `HasDocumentationComment`; the mixin does not merely represent a doc comment; it contains utilities and data for a thing that _has_ a documentation comment.)

### DIFF
--- a/lib/src/model/getter_setter_combo.dart
+++ b/lib/src/model/getter_setter_combo.dart
@@ -191,30 +191,22 @@ mixin GetterSetterCombo on ModelElement {
   late final String? documentationComment =
       _getterSetterDocumentationComment ?? element.documentationComment;
 
-  @override
-  bool get hasDocumentationComment =>
-      _getterSetterDocumentationComment != null ||
-      element.documentationComment != null;
-
   /// Derives a documentation comment for the combo by copying documentation
   /// from the [getter] and/or [setter].
   late final String? _getterSetterDocumentationComment = () {
-    // Check for synthetic before public, always, or stack overflow.
-    var getterComment = switch (getter) {
-      Accessor(isSynthetic: false, isPublic: true) && var g =>
-        g.documentationFrom.first.documentationComment,
-      _ => null,
-    };
-    var setterComment = switch (setter) {
-      Accessor(isSynthetic: false, isPublic: true) && var s =>
-        s.documentationFrom.first.documentationComment,
-      _ => null,
-    };
+    String? getComment(Accessor? a) => switch (a) {
+          Accessor(isSynthetic: false, isPublic: true) =>
+            a.documentationFrom.firstOrNull?.documentationComment,
+          _ => null,
+        };
+
+    var getterComment = getComment(getter);
+    var setterComment = getComment(setter);
 
     if (setterComment == null) return getterComment;
-    return getterComment == null
-        ? setterComment
-        : '$getterComment\n\n$setterComment';
+    if (getterComment == null) return setterComment;
+
+    return '$getterComment\n\n$setterComment';
   }();
 
   @override


lib/src/model/documentation_comment.dart
```diff
@@ -65,16 +65,15 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+  bool get hasNodoc => documentationComment?.contains('@nodoc') ?? false;
 
   /// Processes a [documentationComment], performing various actions based on
   /// `{@}`-style directives (except tool directives), returning the processed
   /// result.
   String _processCommentWithoutTools() {
     // We must first strip the comment of directives like `@docImport`, since
     // the offsets are for the source text.
-    var docs = _stripDocImports(documentationComment);
+    var docs = _stripDocImports(documentationComment ?? '');
```
szakarias, line 76: Would it maybe be cleaner to just do an early bail-out in this method if `documentationComment` is `null`? It seems a bit wrong to pass the empty comment for processing.

lib/src/model/documentation_comment.dart
```diff
@@ -88,7 +87,7 @@ mixin DocumentationComment implements Warnable, SourceCode {
   Future<String> processComment() async {
     // We must first strip the comment of directives like `@docImport`, since
     // the offsets are for the source text.
-    var docs = _stripDocImports(documentationComment);
+    var docs = _stripDocImports(documentationComment ?? '');
```
szakarias, line 90: The same here.

lib/src/model/documentation_comment.dart
```diff
@@ -859,7 +858,8 @@ mixin DocumentationComment implements Warnable, SourceCode {
 
   /// The set of libraries which this [Library] is canonical for.
   late final Set<String> canonicalFor = {
-    for (var match in _canonicalForRegExp.allMatches(documentationComment))
+    for (var match
+        in _canonicalForRegExp.allMatches(documentationComment ?? ''))
```
szakarias: Again to avoid passing the empty string, maybe this?

```
  late final Set<String> canonicalFor = {
    if (documentationComment case var comment?)
      for (var match in _canonicalForRegExp.allMatches(comment))
        match.group(1)!,
  };
```

lib/src/model/getter_setter_combo.dart
```diff
@@ -188,48 +188,33 @@ mixin GetterSetterCombo on ModelElement {
...
-          ? element.documentationComment ?? ''
-          : _getterSetterDocumentationComment;
+  late final String? documentationComment =
+      _getterSetterDocumentationComment ?? element.documentationComment;
 
   @override
   bool get hasDocumentationComment =>
-      _getterSetterDocumentationComment.isNotEmpty ||
+      _getterSetterDocumentationComment != null ||
       element.documentationComment != null;
```
szakarias: Can we just do 

`@override
bool get hasDocumentationComment => documentationComment != null;
`

lib/src/model/getter_setter_combo.dart
```diff
@@ -188,48 +188,33 @@ mixin GetterSetterCombo on ModelElement {
...
-    }
-
-    if (!hasSetter) {
-      return getterComment;
-    }
-
-    final setter = this.setter!;
-    if (setter.isSynthetic || !setter.isPublic) return getterComment;
-
-    assert(setter.documentationFrom.length == 1);
```
szakarias, line 224: We lose this assertion. Is it needed? In that case we should maybe consider doing `documentationFrom.firstOrNull?.documentationComment`

lib/src/model/getter_setter_combo.dart
```diff
@@ -188,48 +188,33 @@ mixin GetterSetterCombo on ModelElement {
...
+    var setterComment = switch (setter) {
+      Accessor(isSynthetic: false, isPublic: true) && var s =>
+        s.documentationFrom.first.documentationComment,
+      _ => null,
+    };
+
+    if (setterComment == null) return getterComment;
+    return getterComment == null
+        ? setterComment
+        : '$getterComment\n\n$setterComment';
```
szakarias: To avoid duplicating the code we could do:  

```
late final String? _getterSetterDocumentationComment = () {

    String? getComment(Accessor? a) => switch (a) {
      Accessor(isSynthetic: false, isPublic: true) => 
        a!.documentationFrom.firstOrNull?.documentationComment,
      _ => null,
    };

    var getterComment = getComment(getter);
    var setterComment = getComment(setter);

    if (setterComment == null) return getterComment;
    if (getterComment == null) return setterComment;
    
    return '$getterComment\n\n$setterComment';
  }();
```

szakarias (approved):
I really like the idea of making `documentationComment` nullable instead of using an empty string pattern! Good find!

lib/src/model/getter_setter_combo.dart
szakarias: Question: In general, should it be the case that the documentation string is non-empty for it to count?

lib/src/model/documentation_comment.dart
srawlins: Yes these points sort of show that I am just "passing the buck" of empty strings down into other pieces of code.

I tried to do exactly as you suggest. The (very annoying) problem is the side-effects of searching for categories etc, which initializes the `_categoryNames` and `_subCategoryNames` fields. See:

```dart
List<String> get categoryNames {
  // TODO(srawlins): avoid side-effect dependency.
  documentationLocal;
  return _categoryNames!;
}
```

I could put a TODO on line 76 here noting this.

Alternatively, if I have a big if/else for the nullable `documentationComment`, I could initialize `_categoryNames` and `_subCategoryNames` to empty lists in the else branch. I would apply the same logic on line 90.

I decided for this PR that that was getting ugly, and I just need to simplify the initialization of `_categoryNames` etc to avoid these three positions where we call `documentationLocal;` for the side effects.

lib/src/model/getter_setter_combo.dart
srawlins: Very elegant, done!

lib/src/model/getter_setter_combo.dart
srawlins: Done, at which point its the same as the code it is overriding, and can just be removed!

lib/src/model/getter_setter_combo.dart
srawlins: I don't think it is helpful. I think there are many overly defensive "we should not get into this state" assertions in dartdoc that just aren't justified. But we could keep it if you like.

PR #4193, opened 2026-01-28
 Refactor Package to use MarkdownFileDocumentation mixin

szakarias:
* Refactors `Package` to use the `MarkdownFileDocumentation` mixin 
* Updates the mixin's `documentation` getter to return `null` instead of `''` when documentation is missing. 

This removes redundant code and aligns file-based documentation with the null-safety improvements in #4190

### DIFF
--- a/lib/src/model/package.dart
+++ b/lib/src/model/package.dart
@@ -3,8 +3,8 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'package:analyzer/dart/element/element.dart';
+import 'package:analyzer/file_system/file_system.dart';
 import 'package:dartdoc/src/dartdoc_options.dart';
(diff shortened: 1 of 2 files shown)


lib/src/model/package.dart
```diff
@@ -30,7 +30,7 @@ const String htmlBasePlaceholder = r'%%__HTMLBASE_dartdoc_internal__%%';
 /// A [LibraryContainer] that contains [Library] objects related to a particular
 /// package.
 final class Package extends LibraryContainer
-    with Nameable, Warnable, CommentReferable {
+    with Nameable, Warnable, CommentReferable, MarkdownFileDocumentation {
```
srawlins, line 33: Ha, I was a little confused about this. I think I assumed that Package _did_ mixin MarkdownFileDocumentation 😅 .

PR #4194, opened 2026-01-29
Introduce unit tests for "properties"

srawlins:
In trying to reason about "properties" (fields, getters, setters), I needed some unit tests. We didn't have any that cleanly explore how property-_inducing_ things (fields) and property _accessors_ (getters and setters) work together, especially with inheritance.

We had the old-school end-to-end tests, which are hard to understand as the test code is very distant from the test. This starts to clean that up a bit.

### DIFF
--- a/test/properties_test.dart
+++ b/test/properties_test.dart
@@ -338,7 +338,7 @@ class C {
 ''');
     var x = library.instanceField('C', 'x');
     expect(x.fullyQualifiedName, 'properties.C.x');
-    // TODO(srawlins): Should this include the getter part from the super-field?
+    // TODO(srawlins): Is this correct?
     expect(x.documentationAsHtml, '<p>Comment 2.</p>');
   }
 


srawlins:
In the first commit here, I have some 'model_test.dart:' comments. These indicate when a test case is a replacement for a deleted model_test.dart test case, so that they can be tracked. Those comments are just for code review, and are deleted in the second commit.

test/properties_test.dart
```diff
@@ -0,0 +1,392 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
```
szakarias: nit
```suggestion
// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
```

test/properties_test.dart
```diff
@@ -0,0 +1,392 @@
...
+  set x(int value) {}
+}
+
+class D extends C {
+  @override
+  int x = 1;
+}
+''');
+    var x = library.instanceField('D', 'x');
+    // TODO(srawlins): Why does this not have the setter?
```
szakarias: I believe this is the wanted behavior, and it aligns with what is written in the spec. However, the current implementation might have intended something else. I have a TODO to revisit the getter/setter setup to ensure the behavior is sane and correctly documented in the spec.

test/properties_test.dart
```diff
@@ -0,0 +1,392 @@
...
+}
+
+class D extends C {
+  /// Comment 2.
+  set x(int value) {}
+}
+''');
+    var x = library.instanceSetter('D', 'x=');
+    expect(x.fullyQualifiedName, 'properties.D.x=');
+    expect(x.documentationAsHtml, '<p>Comment 2.</p>');
```
szakarias, line 310: If we agree on what is in the spec, this might actually not be the wanted behavior.... We could add a todo here.

test/properties_test.dart
```diff
@@ -0,0 +1,392 @@
...
+class C {
+  int get x => 0;
+
+  /// Comment 2.
+  set x(int value) {}
+}
+''');
+    var x = library.instanceField('C', 'x');
+    expect(x.fullyQualifiedName, 'properties.C.x');
+    expect(x.documentationAsHtml, '<p>Comment 2.</p>');
```
szakarias, line 342: same here...

szakarias (approved):
These tests are so much easier to read. I love it!

test/properties_test.dart
srawlins: Done!

test/properties_test.dart
srawlins: Ah excellent, thanks!

test/properties_test.dart
srawlins: Done.

PR #4195, opened 2026-01-29
Refactor categorization to be lazy and side-effect-free

szakarias:
Decouples category parsing from the full documentation build to fix side-effect dependencies.

 Changes:
 - Categories are now parsed lazily from the raw comment when requested, instead of forcing a full documentation build.
 - Added early bail-outs for missing documentation to stop passing/processing empty strings.
 - Switched documentationLocal to use a cleaner nullable backing field pattern.
 - Added tests for parse_categorization

### DIFF
--- a/lib/src/model/documentation_comment.dart
+++ b/lib/src/model/documentation_comment.dart
@@ -71,29 +71,33 @@ mixin DocumentationComment implements Warnable, SourceCode {
   /// `{@}`-style directives (except tool directives), returning the processed
   /// result.
   String _processCommentWithoutTools() {
-    // We must first strip the comment of directives like `@docImport`, since
-    // the offsets are for the source text.
-    var docs = _stripDocImports(documentationComment ?? '');
-    docs = stripCommentDelimiters(docs);
-    // TODO(srawlins): Processing templates here causes #2281. But leaving
-    // them unprocessed causes #2272.
-    docs = _processCommentDirectives(docs, processMacros: false);
-    return docs;
+    if (documentationComment case var comment?) {
+      // We must first strip the comment of directives like `@docImport`, since
+      // the offsets are for the source text.
+      var docs = _stripDocImports(comment);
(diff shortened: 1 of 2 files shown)


lib/src/model/documentation_comment.dart
```diff
@@ -111,6 +115,57 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+      categories: categorySet == null
+          ? const []
+          : (categorySet.toList(growable: false)..sort()),
+      subCategories: subCategorySet == null
+          ? const []
+          : (subCategorySet.toList(growable: false)..sort()),
+    );
+  }
+
+  final Set<String> _categoryNamesSet = {};
```
srawlins, line 154: This works, and is a fine solution. I've tried to avoid this solution because it adds thousands of empty sets to the vast majority of elements, as the vast majority have no categories.

The old way would use `const []` which is nicely just one constant object.

However, this always may have been too defensive on my part; maybe these empty Sets really do no harm in a short-lived process like dartdoc.

lib/src/model/documentation_comment.dart
```diff
@@ -792,52 +836,21 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
 
   bool get hasCategoryNames => categoryNames.isNotEmpty;
-  List<String>? _categoryNames;
 
   /// A set of strings containing all declared categories for this element.
   List<String> get categoryNames {
-    // TODO(srawlins): avoid side-effect dependency.
-    documentationLocal;
-    return _categoryNames!;
+    _ensureInitialCategorization();
```
srawlins, line 852: Very nice :D

lib/src/model/documentation_comment.dart
szakarias: Good point! Let's try this out for now because of the benefits, but keep it in mind.

PR #4197, opened 2026-01-30
Convert some "mixed-in field" tests into unit tests

srawlins:
These "mixed-in field" tests are a little complex, involving four types and a diamond-shaped inheritance graph. But they are meaningful; sort of the smallest diamond-shape of inheritance. I struggled to make sense of the end-to-end tests (as a change I was making was breaking them) and just had to convert them to unit tests to make them make sense 😅 .

### DIFF
--- a/test/properties_test.dart
+++ b/test/properties_test.dart
@@ -191,7 +191,7 @@ class E extends D with M {}
   }
 
   void
-      test_field_inherited_fromExtened_andFromTransitiveExtended_andMixin() async {
+      test_field_inherited_fromExtended_andFromTransitiveExtended_andMixin() async {
     var library = await bootPackageWithLibrary('''
 class C {
   /// Comment.


test/properties_test.dart
```diff
@@ -99,6 +100,134 @@ class D extends C {
...
+    expect(x.canonicalEnclosingContainer, library.mixins.named('M'));
+    expect(x.documentationFrom, hasLength(1));
+    expect(
+      x.documentationFrom.single,
+      library.classes.named('C').instanceFields.named('x').getter,
+    );
+  }
+
+  void
+      test_field_inherited_fromExtened_andFromTransitiveExtended_andMixin() async {
```
szakarias:
```suggestion
      test_field_inherited_fromExtended_andFromTransitiveExtended_andMixin() async {
```

test/end2end/model_test.dart
```diff
@@ -1623,79 +1611,6 @@ void main() async {
...
     });
-
-    test('Verify overrides & documentation inheritance work as intended', () {
-      expect(overrideByEverything.canonicalEnclosingContainer,
-          equals(TypeInferenceMixedIn));
-      expect(overrideByGenericMixin.canonicalEnclosingContainer,
-          equals(GenericMixin));
-      expect(overrideByBoth.canonicalEnclosingContainer, equals(GenericMixin));
-      expect(overrideByModifierClass.canonicalEnclosingContainer,
-          equals(ModifierClass));
```
szakarias, line 1634: I am not sure how important this is, but are we maybe missing this test case in the new (and much nicer) setup?  

If I understand it correctly we are testing that a field in a subclass identifies the correct enclosing container even though there is a mixin (that does not override the field)? Something like:

```
    class C {
      int x = 0;
    }
    class D extends C {
      @override
      int x = 1;
    }
    mixin M on C {
      // M does not override x
    }
    class E extends D with M {}
    ```

srawlins: Hmm, you're right... I replaced each test one-by-one, I should have covered everything, but yes it looks like I'm missing this case. I think I am getting a cold; maybe my brain is foggy 😅 .

PR #4201, opened 2026-02-11
Add support for the @example directive

szakarias:
Introduces the `{@example path [lang=LANGUAGE] [indent=keep|strip]}` directive, enabling developers to inject content from external source files directly into documentation comments as fenced code blocks. 
   * **Block-Level Requirement**: The directive must appear on its own line to be processed. If used inline, it will be ignored.
   * **Code Block Customization**:
       * `lang`: Specify the language for the fenced code block (defaults to the file extension).
       * `indent`: Choose whether to strip common indentation (default) or keep the original file formatting. To avoid broken formatting, it issues a warning and skips stripping if non-space characters (e.g. tabs) are detected in the indentation.
   * **Path Resolution** :
       * Supports paths relative to the current file or absolute paths starting from the package root (e.g., `{@example /examples/hello.dart}`). 
       * Resolution logic ensures that requested files stay within the package root and handles URI-encoded characters.
   * **Error Handling**: Adds a new missing-example-file warning for broken paths and validates against disallowed URI schemes.

  Note: Support for extracting specific `#region` segments from files is planned for a follow-up PR.

### DIFF
--- a/lib/src/model/documentation_comment.dart
+++ b/lib/src/model/documentation_comment.dart
@@ -354,8 +354,7 @@ mixin DocumentationComment implements Warnable, SourceCode {
   ///   prevent broken formatting.
   ///
   /// The `lang` parameter specifies the language for the fenced code block. If
-  /// not provided, it defaults to the file extension of `path`, or `dart` if
-  /// the extension is empty.
+  /// not provided, it defaults to the file extension of `path`.
   ///
   /// Example:
   ///


srawlins:
Just FYI it would be nice to implement the parsing work in the analyzer, so that this isn't a dartdoc-only feature. The analyzer could render `@example`s in hovers, can provide diagnostics (typos etc).

lib/src/model/documentation_comment.dart
```diff
@@ -320,6 +323,178 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+  /// Syntax:
+  ///
+  ///     &#123;@example PATH [lang=LANGUAGE] [indent=keep|strip]&#125;
+  ///
+  /// The `PATH` is resolved relative to the package root or the current file;
+  /// see [resolveExamplePath] for details.
+  ///
+  /// The `lang` parameter specifies the language for the fenced code block. If
+  /// not provided, it defaults to the file extension of `PATH`, or `dart` if
+  /// the extension is empty.
```
jonasfj: nit:
```suggestion
  ///     &#123;@example <path> [lang=LANGUAGE] [indent=keep|strip]&#125;
  ///
  /// The `path` is resolved relative to the package root or the current file;
  /// see [resolveExamplePath] for details.
  ///
  /// The `lang` parameter specifies the language for the fenced code block. If
  /// not provided, it defaults to the file extension of `path`, or `dart` if
  /// the extension is empty.
```

lib/src/model/documentation_comment.dart
```diff
@@ -320,6 +323,178 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+  ///     &#123;@example ../subdir/utils.dart lang=dart indent=strip&#125;
+  String _injectExamples(String rawDocs) {
+    return rawDocs.replaceAllMapped(_exampleRegExp, (match) {
+      var argsAsString = match[1];
+      if (argsAsString == null) {
+        warn(PackageWarning.invalidParameter,
+            message: 'Must specify a file path for the @example directive.');
+        return '';
+      }
+      var args = _parseArgs(argsAsString, _exampleArgParser, 'example');
```
jonasfj, line 372: Not your crime I see this done else where, but this is going to convert `lang=dart indent=strip` to `['--lang=dart', '--indent=strip']` and then you're going to parse it with `ArgParser`.

I'm not a 100% sure, but reading the code I'm getting the impression that users will be able to write:
`{@example path/to/file.dart --lang=dart --indent=strip}`

Do we want to allow both `--lang=dart` and `lang=dart`? We risk wanting to break that behavior later.

lib/src/model/documentation_comment.dart
```diff
@@ -320,6 +323,178 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+      if (resolved == null) return '';
+
+      var file = packageGraph.resourceProvider.getFile(resolved.path);
+      if (!file.exists) {
+        warn(PackageWarning.missingExampleFile, message: filepath);
+        return '';
+      }
+
+      var extension = resolved.extension;
+      var language = args['lang'] ?? (extension.isEmpty ? 'dart' : extension);
```
jonasfj: if there is no `lang` option and no file extension I'd suggest we don't add a language specifier.

If people embed `/AUTHORS` or  `/LICENSE` let's not give them language marker dart :D

Also who writes Dart code in a file without a `.dart` extension? :see_no_evil:  Surely, someone does, but I'd suggest we don't encourage it :rofl:

lib/src/model/documentation_comment.dart
```diff
@@ -320,6 +323,178 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+    for (var line in lines) {
+      if (line.trim().isEmpty) continue;
+      var indent = line.length - line.trimLeft().length;
+      if (minIndent == null || indent < minIndent) {
+        minIndent = indent;
+      }
+    }
+
+    if (minIndent == null || minIndent == 0) return content;
+
```
jonasfj, line 574: We could consider only stripping in increments of 2.

So if `minIndent = 3` then round it to `2`.

Just because most of use 2 spaces for indentation. But then again, it'd probably be a problem for someone using tabs. And having a single space infront of each line seems like something you wouldn't do intentionally.

lib/src/model/documentation_comment.dart
```diff
@@ -320,6 +323,178 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+      var file = packageGraph.resourceProvider.getFile(resolved.path);
+      if (!file.exists) {
+        warn(PackageWarning.missingExampleFile, message: filepath);
+        return '';
+      }
+
+      var extension = resolved.extension;
+      var language = args['lang'] ?? (extension.isEmpty ? 'dart' : extension);
+      var indent = args['indent'] ?? 'strip';
```
jonasfj, line 413: We could consider letting `indent` default to `keep` if embedding a full file, and defaulting `indent` to `strip` if embedding a region of a file.

I don't know anyone who creates files where every line starts with a whitespace. But there is probably an ascii-artist out there doing things like that 🤣

I could also be convinced to not special case this behavior.

szakarias:
> Just FYI it would be nice to implement the parsing work in the analyzer, so that this isn't a dartdoc-only feature. The analyzer could render `@example`s in hovers, can provide diagnostics (typos etc).

I agree!  I will start by landing it here, and then we can move it in a follow-up.

lib/src/model/documentation_comment.dart
```diff
@@ -320,6 +323,178 @@ mixin DocumentationComment implements Warnable, SourceCode {
         .cast<String, String>();
   }
 
+  /// Replace &#123;@example ...&#125; in API comments with a fenced code block
```
lrhn, line 333: Should you be able to escape the block in some way, either `\{@example banana}` or `{\@example banana}`, so it isn't interpreted.

A `\` is usually used for markdown escape, so you can write \[code\] that isn't interpreted.
(But if a `\@` gets converted to `@` by markdown parsing, then you can't use that signal.)

Could you code-escape here:
```dart
/// Replace `{@example ...}` in ...
```
(Probably require the `{@example ...}` to be on a line of its own. It is concerted to markdown block content, so it should't be used inline.)

lib/src/model/documentation_comment.dart
```diff
@@ -320,6 +323,178 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+  /// containing the contents of the specified file.
+  ///
+  /// Syntax:
+  ///
+  ///     &#123;@example PATH [lang=LANGUAGE] [indent=keep|strip]&#125;
+  ///
```
lrhn, line 341: Don't think this works. A four-space indent is equivalent to a code fence, so this should be the same as:
```dart
  /// Syntax:
  /// ```
  /// &#123;@example PATH [lang=LANGUAGE] [indent=keep|strip]&#125;
  /// ```
```
which is shown as the literal escapes, not the `{...}` braces.

Maybe ignore `{@...}` inside a code fence too.

lib/src/model/documentation_comment.dart
```diff
@@ -320,6 +323,178 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+  /// The `PATH` is resolved relative to the package root or the current file;
+  /// see [resolveExamplePath] for details.
```
lrhn, line 344:
```
... is resolved relative to the current file's path, using the package's root as
the root path.
```
(or something to that effect. Because pedantically, the PATH is resolved relative to the current file's path. If the PATH is absolute, it throws away the current file's path and starts at its root. That's not "relative", it's "absolute using a specific root" ... which there really is not a good way to phrase.)

lib/src/model/documentation_comment.dart
lrhn: Actually disagree. Use the same word in the example and the reference.
Using "PATH" in both places is fine.
Using "\<path\>" in both places is also fine.

Using \<path\> and `` `path` `` is confusing, and using code quotes suggests that `path` is a concrete source name, not a meta-variable.

lib/src/model/documentation_comment.dart
```diff
@@ -320,6 +323,178 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+  ///
+  /// The `lang` parameter specifies the language for the fenced code block. If
+  /// not provided, it defaults to the file extension of `PATH`, or `dart` if
+  /// the extension is empty.
+  ///
+  /// The `indent` parameter specifies whether to `strip` common indentation
+  /// from the file contents (the default) or `keep` it exactly as is.
+  ///
+  /// Example:
+  ///
```
lrhn, line 360: (Use a code fence instead of indentation. Make sure the braces show correctly.)

lib/src/model/documentation_comment.dart
```diff
@@ -320,6 +323,178 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+  ///     &#123;@example /examples/my_example.dart&#125;
+  ///     &#123;@example ../subdir/utils.dart lang=dart indent=strip&#125;
+  String _injectExamples(String rawDocs) {
+    return rawDocs.replaceAllMapped(_exampleRegExp, (match) {
```
lrhn, line 365: I'd *consider* working on parsed Markdown instead of raw strings.
And then not allow `{@example ...}` in inline content, and not recognize it inside code fences,
basically so it has to be `/// {@example ladi.dart}` where markdown block content is allowed.

(Don't know how easy that is, if all other tags are parsed before markdown.)

lib/src/model/documentation_comment.dart
```diff
@@ -320,6 +323,178 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+      var extension = resolved.extension;
+      var language = args['lang'] ?? (extension.isEmpty ? 'dart' : extension);
+      var indent = args['indent'] ?? 'strip';
+
+      var content = file.readAsStringSync();
+      if (indent == 'strip') {
+        content = _stripIndentation(content);
+      }
+
+      return '\n```$language\n$content\n```\n';
```
lrhn: (I'd check that content doesn't already end with a `\n`. If it does, probably shouldn't add another one.
Consider changing content's line ending to `\n` if they aren't already.)

lib/src/model/documentation_comment.dart
```diff
@@ -320,6 +323,178 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+  }
+
+  /// Resolves the given [filepath] as a URI reference relative to the current
+  /// package root.
+  ///
+  /// The [filepath] is resolved as follows:
+  /// - If it starts with a leading `/`, it is resolved relative to the package
+  ///   root.
+  /// - Otherwise, it is resolved relative to the directory containing the file
+  ///   with the documentation comment.
```
lrhn: Or just the file itself, but more importantly, that path *rooted at the package root*.

I'd say "the file". For resolving, there is a difference between `/lib/src` and `lib/src/`, and both can be argued to be "the directory containing the file". If it's resolved relative to the file's path, then there is guaranteed to be a `/` after the last directory.)

lib/src/model/documentation_comment.dart
```diff
@@ -320,6 +323,178 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+  /// - Resolution follows URI reference rules, ensuring that `..` segments
+  ///   stop at the package root and never escape it.
+  /// - [filepath] can include URI-encoded characters (like `%20` for spaces).
+  ///
```
lrhn, line 454: So: The PATH here *must* be a _valid URI path_ (so allows %-escapes and must use `/` as path separator), and it's resolved against a URI path that is the containing file's path with the package root as path root.

lib/src/model/documentation_comment.dart
```diff
@@ -320,6 +323,178 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+  /// Returns the absolute path and extension of the resolved file, or `null` if
+  /// the path could not be resolved or is outside the package root.
```
lrhn: It shouldn't be able to be outside the package root.
(If you only allow it to be a URI path. Then anything that is not a relative or absolute path will be invalid and "cannot be resolved", so still the first case.)

lib/src/model/documentation_comment.dart
szakarias: done

lib/src/model/documentation_comment.dart
szakarias: done

lib/src/model/documentation_comment.dart
szakarias: hmm, I think this is not worth the complication.

lib/src/model/documentation_comment.dart
szakarias: That's a fair point. I'm using the shared `_parseArgs` utility here to be consistent with how other directives (like {@youtube}) are parsed.

I agree that supporting both syntaxes isn't ideal, but I'd prefer to keep the behavior consistent across all directives for now. If we want to strictly enforce `key=value` and disallow the `--` prefix, we should probably update `_parseArgs` and clean it up for all directives in a separate PR.

We can document that `key=value` is the supported syntax. The fact that the underlying parser "accidentally" supports `--` is an implementation detail that users are (hopefully) unlikely to stumble upon if we don't advertise it.

lib/src/model/documentation_comment.dart
szakarias: I think I'd prefer to stick with `strip` as the single default to keep the mental model simple.

For a standard, well-formatted file (where top-level code starts at column 0), `strip` calculates a minimum indentation of 0 and effectively acts like `keep` anyway. So it "just works" for full files.

I'd worry that changing the default based on whether a region is used or not might be confusing for users.

lib/src/model/documentation_comment.dart
```diff
@@ -320,6 +323,178 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+  @visibleForTesting
+  static ({String path, String extension})? resolveExamplePath(
+    String filepath, {
+    required String packagePath,
+    required String? sourceFileName,
+    required p.Context pathContext,
+    required void Function(PackageWarning, {String? message}) warn,
+  }) {
+    // Resolve the path as a URI reference.
```
lrhn, line 466: You're using a `package:path` `Context`, which may not do URI resolution.
(If it's always`path.url`, it doesn't need to be a parameter.)

Either use `path.url` or use `Uri` for resolution. Then convert to a file system path only at the end.

lib/src/model/documentation_comment.dart
```diff
@@ -320,6 +323,178 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+  /// - Resolution follows URI reference rules, ensuring that `..` segments
+  ///   stop at the package root and never escape it.
+  /// - [filepath] can include URI-encoded characters (like `%20` for spaces).
+  ///
+  /// Returns the absolute path and extension of the resolved file, or `null` if
+  /// the path could not be resolved or is outside the package root.
+  @visibleForTesting
+  static ({String path, String extension})? resolveExamplePath(
+    String filepath, {
+    required String packagePath,
```
lrhn, line 458: Not documented.  What is it?

Guessing:
```dart
/// The [packagePath] is the file system path of the root of the current package,
/// in the form used by [pathContext].
///
/// The [sourceFilePath] is the file system path of the file containing the example
/// link in the form used by [pathContext].
/// If absent, the [filepath]
/// Always inside the package root given by [packagePath].
```
(Then start by `assert(sourceFilePath == null || pathContext.isWithin(packagePath, sourceFilePath));`, just for the sanity checking.)

lib/src/model/documentation_comment.dart
```diff
@@ -320,6 +323,178 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+    required String? sourceFileName,
+    required p.Context pathContext,
+    required void Function(PackageWarning, {String? message}) warn,
+  }) {
+    // Resolve the path as a URI reference.
+    // 1. Get the relative path of the current file from the package root.
+    var relativeBase = sourceFileName != null
+        ? pathContext.relative(sourceFileName, from: packagePath)
```
lrhn: Is the `sourceFileName` a path? Should it be named `sourceFilePath`?

lib/src/model/documentation_comment.dart
```diff
@@ -320,6 +323,178 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+        : '';
+
+    // 2. Convert to POSIX-style for URI parsing (using '/' as separator).
+    // We start with a '/' so that Uri.resolve() stops at the package root
+    // when it sees too many '..' segments.
+    var relativeBasePosix = '/${pathContext.split(relativeBase).join('/')}';
```
lrhn: Don't do path conversion yourself, use `pathContext`. See below.

lib/src/model/documentation_comment.dart
```diff
@@ -320,6 +323,178 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+  ///   with the documentation comment.
+  /// - Resolution follows URI reference rules, ensuring that `..` segments
+  ///   stop at the package root and never escape it.
+  /// - [filepath] can include URI-encoded characters (like `%20` for spaces).
+  ///
+  /// Returns the absolute path and extension of the resolved file, or `null` if
+  /// the path could not be resolved or is outside the package root.
+  @visibleForTesting
+  static ({String path, String extension})? resolveExamplePath(
+    String filepath, {
```
lrhn: Be consistent with capitalization, `packagePath` has capital `P`, so `filePath` probably should too.

Maybe call it `exampleFilePath`, there are more than one file path here, and it's positional so the length doesn't matter.

lib/src/model/documentation_comment.dart
```diff
@@ -320,6 +323,178 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+    // 3. Resolve using [Uri.resolve].
+    // This handles:
+    // - Leading '/' (relative to root)
+    // - Relative paths (relative to base)
+    // - '..' (stops at root)
+    // - URI encoding (%20 etc)
+    String resolvedPath;
+    String extension;
+    try {
+      var baseUri = Uri.parse(relativeBasePosix);
```
lrhn: Skip the `\` &rarr; `/` conversion above and let `pathContext` do it:
```dart
  var baseUri = pathContext.toUri(pathContext.join(pathContext.pathSeparator, relativeBase));
```
That gives you a `file:` URI, but that's OK, we can always ignore the `file:` later.

(Better to not do any path/uri logic yourself, let the well-tested helper code in `package:path` and `Uri` do all the logic. It won't forget any cases. Or if it does, we'll fix that bug at the root!)

Using `parse` on something that is not known to be a URI is dangerous. If the file system path contains a `?` or `#` those will be escaped by `pathContext.toUri`, or even `Uri(path: ...)`, but the `.parse` will treat the first two as a query or fragment starter. And if the path contains `%`s, it's a coin-toss whether `Uri` will escape it or not. (Meaning that *I* don't remember which constructors do or don't, or if they only do if what follows isn't two hex digits. And if it is two hex digits, it may be reinterpreted as an escape, and even capitalize the letters, so `project/its100%efficient/foo.dart` becomes `file:///projects/its100%EFficent/foo.dart`, which won't on a case sensitive file system when converted back, whether unescaped or not. It really is turtles all the way down!)

lib/src/model/documentation_comment.dart
```diff
@@ -320,6 +323,178 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+      var resolvedUri = baseUri.resolve(filepath);
+
+      // 4. Convert back to a package-relative file path.
+      resolvedPath = pathContext.joinAll([
+        packagePath,
+        ...resolvedUri.pathSegments,
+      ]);
+      extension = pathContext
+          .extension(resolvedUri.path)
```
lrhn: Use `resolvedPath` as argument here.
If the extension contains URI escapes in the `Uri`, you want them unescaped when treating it as a path.

(In general, never take something from `Uri` and directly treat it as a file system path, or something from `pathContext` and treat it as a `Uri`, always use a conversion first. Probably one in `pathContext` since it's more knowledgable about file system paths.)

lib/src/model/documentation_comment.dart
```diff
@@ -320,6 +323,178 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+    String extension;
+    try {
+      var baseUri = Uri.parse(relativeBasePosix);
+      var resolvedUri = baseUri.resolve(filepath);
+
+      // 4. Convert back to a package-relative file path.
+      resolvedPath = pathContext.joinAll([
+        packagePath,
+        ...resolvedUri.pathSegments,
+      ]);
```
lrhn: Let `pathContext` do the work here too. (Sometimes, if the path separator is `/`, there is no work, so it's also cheaper.)
```dart
resolvedPath = pathContext.join(packagePath, pathContext.fromUri(resolvedUri));
```
(I think using `pathSegments` here is otherwise the correct choice, those should be properly unescaped. But safer to just use `pathContext`.)


(Do check what `fromUri` throws if its input is not valid. It *should* be valid, since the URI has the same scheme as the one created by `pathContext.toUri`, but try to see if you can break it somehow.)

lib/src/model/documentation_comment.dart
```diff
@@ -320,6 +323,178 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+      extension = pathContext
+          .extension(resolvedUri.path)
+          .replaceFirst('.', '')
+          .toLowerCase();
+    } on FormatException catch (e) {
```
lrhn, line 505: (Where can this be thrown? if it's only in `Uri.parse`, which should be `pathContext.toUri`, check if it still throws that. If in more than one place, make sure the message is correct for all of them.)

lib/src/model/documentation_comment.dart
```diff
@@ -320,6 +323,178 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+    }
+
+    resolvedPath = pathContext.normalize(resolvedPath);
+
+    if (!pathContext.isWithin(packagePath, resolvedPath) &&
+        !pathContext.equals(packagePath, resolvedPath)) {
+      warn(PackageWarning.invalidParameter,
+          message:
+              'Example file $filepath must be within the package root ($packagePath).');
+      return null;
```
lrhn, line 521: (Any coverage for this? Can it be hit at all?)

lib/src/model/documentation_comment.dart
```diff
@@ -320,6 +323,178 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+    }
+
+    return (path: resolvedPath, extension: extension);
```
lrhn: (You can compute the extension from the resolved path, so you could do it afterwards in a separate function, and not need to return a pair.)

lib/src/model/documentation_comment.dart
```diff
@@ -320,6 +323,178 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+  }
+
+  static String _stripIndentation(String content) {
+    var lines = LineSplitter.split(content).toList();
+    if (lines.isEmpty) return content;
+
+    int? minIndent;
+    for (var line in lines) {
+      if (line.trim().isEmpty) continue;
```
lrhn: Just use `trimLeft` once. If the line is all whitespace, that'll be an empty string as well.
And consider replacing all-whitespace lines with empty strings.
```dart
for (var i = 0; i < lines.length; i++) {
  var line = lines[i];
  var leftTrimmedLine = line.trimLeft();
  if (leftTrimmedLine.isEmpty) {
    if (line.isNotEmpty) lines[i] = '';
    continue;
  }
  var indent = line.length - leftTrimmedLine.length;
```

lrhn: Should this care about tabs?
```
␣␣Bananas are awesome
⇥⇥And highly indented.
⇥And pineapples too.
```
Two spaces have length 2, so do two tabs.


Rather than _counting_ whitespace characters, maybe consider finding _equal whitespace sequences_. If one line starts with space and another with tab, that's not shared whitespace. And only accept space and tab.
```dart
static String _stripIndentation(String content) {
  var lines = LineSplitter.split(content).toList();
  if (lines.isEmpty) return '';
  int firstNonWhitspaceChar(String line, int index) {
    for (;index < line.length; index++) {
      var char = line.codeUnitAt(index);
      if (char != 0x20 && char != 0x09) break;
    }
    return index;
  }
  var index = 0;
  // Find first non-empty line.
  int sharedWhitespaceLength;
  String line;
  firstNonEmptyLine:
  while (true) {
    line = lines[index++]; 
    if (line.isEmpty) continue;
    var nonWhitespaceIndex = firstNoWhitespaceChar(line, 0);
    if (nonWhitespaceIndex < line.length) {
      sharedWhitespaceLength = nonWhitespaceIndex;
      break;
    } else {
      lines[index - 1] = ''; // Canonicalize whitespace-only lines to empty.
    }
    if (index == lines.length) return '';
  }
  while (index < lines.length) {
    var otherLine = lines[index++];
    if (otherLine.isEmpty) continue;
    for (var i = 0; i < sharedWhitespaceLength; i++) {
      if (i == otherLine.length || otherLine.codeUnitAt(i) != line.codeUnitAt(i)) {
        // Check if line is all-whitespace, then don't count it.
        if (firstNonWhitespaceChar(otherLine, i) == otherLine.length) {
          lines[index - 1] = '';
        } else {
          sharedWhitespaceLength = i;
        }
        break;
      }
    }
  }
  if (sharedWhitespaceLength > 0) {
    for (var i = 0; i < lines.length; i++) {
      var line = lines[i];
      if (line.isNotEmpty) {
        lines[i] = line.substring(sharedWhitespaceLength);
      }
    }
  }
  return lines.join('\n');
}
```
Yeah, that got complicated. (I try to not create unnecessary allocations. But even the line-splitting is unnecessary, one can find the shared whitespace directly in the original string, then replace that sequence with `""` when at start of a line as a single replace operation. No extra allocations needed. And here I go micro-optimizing again 😢 )

Maybe *only* accept space characters, and just complain if the shared whitespace length contains any non-spaces. (Better to get a warning so you can fix an accidental tab, than silently getting a weird output.)

lib/src/model/documentation_comment.dart
lrhn: Agree that nobody ever wanted to end up with a one-space indent if they ask to remove shared indent.
If they want to control indent, do it in the source, and don't remove it.

lib/src/model/documentation_comment.dart
```diff
@@ -320,6 +323,178 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+
+    if (minIndent == null || minIndent == 0) return content;
+
+    return lines.map((line) {
+      if (line.length < minIndent!) return '';
+      return line.substring(minIndent);
+    }).join('\n');
+  }
+
+  static final _exampleRegExp = RegExp(r'{@example(?:\s+([^\s}][^}]*))?}');
```
lrhn: Escape `{`, it's significant in RegExp. (RegExp parsers are too lenient, if it doesn't parse as `{digits}`, they treat the `{` as literal, but that's technically not part of the ECMAScript RegExp specification proper, it's an "Annex B compatibility extension".)
```dart
 static final _exampleRegExp = RegExp(r'\{@example(?:\s+([^\s}][^}]*))?\}');
```
This staill makes the file path optional. It matches `{@example}` and `{@example x}`, _but not_ `{@example }`. Should probably match that too, so consider:
```dart
 static final _exampleRegExp = RegExp(r'\{@example\b(?:[^}]*)\}');
```

The `\s+` in the original, allows newlines. Should it?

The `[^}]` allows *every character except `}`*, including `\n  /// {@example otherExample`,
so if you forget a `}`, it'll continue until the next `}` *anywhere* later.
You'll probably get an error about the content, but it's still not good parsing recovery.
(It's run on the entire doc, not just individual lines, right?)

I'd consider only allowing `{@example...}` on a line of its own, with no newlines inside,
and nothing before or after.
So maybe `RegExp(r'^[ \t]*@\{example(?:[ \t]+([^}\r\n]*?)\}[ \t]*$', multiLine: true)`.

Can also give you a way to escape the predicate if you don't want it taken literally.
Say allowing `\{@....}` to not be taken as a dartdoc directive. Which requiring only whitespace before the `{` will do, otherwise it can use `(?<!(?<!\\)(\\\\)*\\)\{` to match the `{`-without-`\`-before, counting an even number of `\`s as escaped `\`s. 

`RegExp(r'^[ \t]*@\{example(?:[ \t]+([^}\r\n]*?)\}[ \t]*$', multiLine: true)`.

(Can still look for `{@example ...}` that doesn't satisfy that, and give a warning for them.
So an alternative could be:
`RegExp(r'(?<pre>^.*)\{@example([^}]*)\}(?<post>.*$)'), and then check if `(match as RegExpMatch).namedGroup('pre')!.trim().isNotEmpty()` or same for `post`, and complain if they are not all whitespace. And also complain if `match[1]` contains a line terminator.)

Or, as mentioned before, I'd do this *after* markdown parsing, replacing a single block content containing the `{@example...}` text with a code fence. (If parsed markdown is mutable, otherwise I can see why you want to do it first.)

Allowing `/// foo {@example bar.dart} baz` to be replaced into:
```dart
/// foo
/// ```
/// conten of bar.dart
/// ```
/// baz
```
will work (and the extra newlines are there to ensure it), but will also trigger inside markdown "literals" like
```dart
/// ```javascript
/// var value = "{@example banana}";
/// ```
```
and there is no way to escape that. (That's my main reason to prefer doing this after markdown parsing: Treating `{@example...}` as a *markdown macro*, not an "ASCII text macro".
But that will prevent recognizing `\{@`, since markdown removes the `'\` 😢.)

But this can work, and can be made more precise later. Nobody should *want* to put the `{@example}` where it's not on a line by itself.

lib/src/model/documentation_comment.dart
jonasfj: I'm down with us solving this later.

test/example_directive_test.dart
```diff
@@ -0,0 +1,171 @@
...
+
+@reflectiveTest
+class ExampleDirectiveTest extends DocumentationCommentTestBase {
+  Future<void> _bootPackage(String comment,
+      {Map<String, String> files = const {},
+      String libraryPath = 'lib/a.dart'}) async {
+    projectRoot = utils.writePackage(packageName, resourceProvider);
+    var pathContext = resourceProvider.pathContext;
+    for (var entry in files.entries) {
+      var pathParts = pathContext.split(entry.key);
```
lrhn: I _think_ all path contexts split on `/`, even `windows`, so this probably works.

But, pedantically, the `entry.key` here is a file system path provided by `pathContext`, it's a `/`-separated sequence of path segments defined in the "test input domain", so the safest thing would be to do `entry.key.split('/')`.

You have a string with a format, so who defined the format?
Only use tools applicable to that format, and explicit conversions to other formats, rather than trying to interpret one format using a tool for another.
Here the format is defined by the author of this file, as `/`-separated, not by `pathContext`. It would be more precise to use `path.posix.split(entry.key)`, because the format of the string is the Posix path format, not necessarily the `pathContext` format.

test/example_directive_test.dart
```diff
@@ -0,0 +1,171 @@
...
+}
+
+@reflectiveTest
+class ExampleDirectiveTest extends DocumentationCommentTestBase {
+  Future<void> _bootPackage(String comment,
+      {Map<String, String> files = const {},
+      String libraryPath = 'lib/a.dart'}) async {
+    projectRoot = utils.writePackage(packageName, resourceProvider);
+    var pathContext = resourceProvider.pathContext;
+    for (var entry in files.entries) {
```
lrhn: (Can do `(var MapEntry(key: path, value: content) in files.entries)`, just so the following is easier to read.)

test/example_directive_test.dart
```diff
@@ -0,0 +1,171 @@
...
+      var pathParts = pathContext.split(entry.key);
+      var currentFolder = projectRoot;
+      for (var i = 0; i < pathParts.length - 1; i++) {
+        if (pathParts[i] == '' || pathParts[i] == pathContext.separator) {
```
lrhn: A `var pathPart = pathParts[i];` would save some repetition.

test/example_directive_test.dart
```diff
@@ -0,0 +1,171 @@
...
+        currentFolder = currentFolder.getChildAssumingFolder(pathParts[i]);
+      }
+      currentFolder
+          .getChildAssumingFile(pathParts.last)
+          .writeAsStringSync(entry.value);
+    }
+
+    var pathParts = pathContext.split(libraryPath);
+    var currentFolder = projectRoot;
+    for (var i = 0; i < pathParts.length - 1; i++) {
```
lrhn, line 44: Same loop and following `getChildAssumingFile` as above, could be a helper function?

test/example_directive_test.dart
lrhn: Use `.isEmpty`, not `== ''`.

Can it ever be empty? (The `split` function isn't that well documented, but a quick check suggets that it never gives an empty segment. I the input has `//`, it omits that segment.
(Even in the `url` context, which isn't really correct URI behavior.)

test/example_directive_test.dart
```diff
@@ -0,0 +1,171 @@
...
+        .getChildAssumingFile(pathParts.last)
+        .writeAsStringSync('$comment\nlibrary;');
+
+    packageGraph =
+        await utils.bootBasicPackage(projectRoot.path, packageMetaProvider);
+
+    libraryModel = packageGraph.defaultPackage.libraries.firstWhere((l) =>
+        pathContext
+            .normalize(l.sourceFileName)
+            .endsWith(pathContext.normalize(libraryPath)));
```
lrhn: Can you use `pathContext` to decide if something ends with something else?
Because if the `libraryPath` doesn't start with a separator, then
`"foo/bar.dart"' is a suffix of `"path/lib/src/barfoo/bar.dart"`, but the former is not a suffix _path_ of the latter. (So does `libraryPath` start with a path separator?)

(Normalize the library path outside of the loop.)


Maybe:
```dart
var normalizedLibraryPath = pathContext.normalize(libraryPath);
assert(pathContext.pathSeparator.length == 1);
libraryModel = ...firstWhere((l) {
   var path = pathContext.normalize(l.sourceFileName);
   return path.endsWith(normalizedLibraryPath) &&
       (pathContext.isAbsolute(normalizedLibraryPath) ||
        path.length == normalizedLibraryPath ||
        path.startsWith(pathContext.pathSeparator, path.length - normalizedLibraryPath.length - 1));
 });
```
So ends with the *text*, and either the normalized library path starts with path separator, or path starts a path segment where the normalized library path suffix starts.

(Alternative is to split both and check that the *segments* match. Safer, but more work.
Can still split the `normalizedLibraryPath` once outside of the loop, and start by checking if the path string is a suffix.)

test/example_directive_test.dart
```diff
@@ -0,0 +1,171 @@
...
+    expect(
+      libraryModel,
+      hasWarning(
+          PackageWarning.missingExampleFile, '/examples/non_existent.dart'),
+    );
+  }
+
+  void test_exampleDirective_missingFilePath() async {
+    await _bootPackage('''
+/// {@example}
```
lrhn, line 210: Consider adding tests for tricky or invalid stuff, to check that it is invalid:
```dart
/// Banana {@example examples/hello.txt} Split
```
```dart
/// Banana {@example
///    examples/hello.txt
///    lang=golfball
///  } Split
```

```dart
/// {@example }
```

```dart
/// {@example
// what's this?
/// example/hello.txt}
```
(I don't know what that means, but I want to see what happens!)

```dart
/// {@example example/hello.txt lang
/// {@example}
```
(Whoops, someone deleted the end of my line! 😎 )

Consider indent-stripping with non-space characters, just as documentation for what it does.
(One tab=one whitespace character, a test will prove that it's deliberate, and that changing it is breaking.)

test/example_path_resolution_test.dart
```diff
@@ -0,0 +1,154 @@
...
+import 'package:path/path.dart' as p;
+import 'package:test/test.dart';
+
+void main() {
+  group('DocumentationComment.resolveExamplePath', () {
+    late p.Context pathContext;
+    late String packagePath;
+
+    setUp(() {
+      pathContext = p.Context(style: p.Style.posix, current: '/project');
```
lrhn: The code in the resolution doesn't seem to assume that `current` is anything specific. 
Making it equal to a relevant path here could mask a problem if it does use it accidentally.

Consider making `current` be `/completelydifferent/and/far/away` or something.

test/example_path_resolution_test.dart
```diff
@@ -0,0 +1,154 @@
...
+      var result = resolve('examples\\hello.dart',
+          sourceFileName: '/project/lib/a.dart');
+      expect(result, '/project/lib/examples/hello.dart');
+    });
+
+    test('unusual path inputs (no crash)', () {
+      var inputs = [
+        '\x00',
+        ' ',
+        '%%',
```
lrhn, line 172: Can try `100%effective`, `a%25b`, `a%5cb` and `a%20b` too. 😎 
And `?foo` and `#foo` too.

And all these nasty characters can appear *validly* in the project and file paths as well, and should be preserved.
(So probably a similar test where special such characters occur in the project path, and in the file path inside the project.)

test/example_path_resolution_test.dart
```diff
@@ -0,0 +1,154 @@
...
+        '',
+        '..',
+        '//',
+        '\\..\\..\\',
```
lrhn: (Can use raw string, `r'\..\..\'`

test/example_path_resolution_test.dart
```diff
@@ -0,0 +1,154 @@
...
+            pathContext: pathContext,
+            warn: (kind, {message}) {},
+          );
+        } catch (e) {
+          fail('resolveExamplePath crashed on input "$input": $e');
+        }
+      }
+    });
+
+    test('invalid URI inputs (trigger warnings)', () {
```
lrhn: (Can move the `test` call inside the loop, to get one test per input, so you can see if more than one is failing.
```dart
  group('invalid Uri inputs (trigger warnings)', () {
     var inputs = [
        ': [', // Invalid URI
        'http://[::1]]', // Invalid IPv6
        '/%ff%fe%fd', // Invalid UTF-8 encoding
      ];

      for (var input in inputs) {
        test(': $input', () {
          var warned = false;
          DocumentationComment.resolveExamplePath(
            input,
            packagePath: packagePath,
            sourceFileName: '/project/lib/a.dart',
            pathContext: pathContext,
            warn: (kind, {message}) => warned = true,
          );
          expect(warned, isTrue, reason: 'Expected a warning for input "$input"');
        });
      }
    });
```
Same above with the weird paths.)

lib/src/model/documentation_comment.dart
```diff
@@ -320,6 +323,178 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+    required void Function(PackageWarning, {String? message}) warn,
+  }) {
+    // Resolve the path as a URI reference.
+    // 1. Get the relative path of the current file from the package root.
+    var relativeBase = sourceFileName != null
+        ? pathContext.relative(sourceFileName, from: packagePath)
+        : '';
+
+    // 2. Convert to POSIX-style for URI parsing (using '/' as separator).
+    // We start with a '/' so that Uri.resolve() stops at the package root
```
sigurdm: Reading this makes me think that we should actually only allow `/` in the path argument.
No reason to allow windows filenames that would not resolve on other platform
Or maybe even specify that it should be a path-only URL.

lib/src/model/documentation_comment.dart
szakarias: About escaping the directive itself (e.g. \{@example}): I haven't implemented that yet. I think it's a good idea, but since this directive is converted to a markdown block, I think requiring it to be on its own line (as it is now) is a reasonable enough signal for now. I can add escaping in a follow-up if we find people are hitting this by accident.

szakarias: I agree. We should be able to escape. And I actually started to add escaping, enforcing line boundaries, and respecting Markdown fences, but I reverted it again -- for now. This is probably better to have as an overall directive rule set that applies to all block-level `dartdoc` directives not just example. I will create an issue to summarize these points, but will leave it out of this PR.

szakarias: https://github.com/dart-lang/dartdoc/issues/4205

lib/src/model/documentation_comment.dart
szakarias: Issue: https://github.com/dart-lang/dartdoc/issues/4205

lib/src/model/documentation_comment.dart
szakarias: This is also something to be considered later, I think.

lib/src/model/documentation_comment.dart
szakarias: done

lib/src/model/documentation_comment.dart
szakarias: done

lib/src/model/documentation_comment.dart
szakarias: done

lib/src/model/documentation_comment.dart
szakarias: done

lib/src/model/documentation_comment.dart
szakarias: done

lib/src/model/documentation_comment.dart
szakarias: done

lib/src/model/documentation_comment.dart
szakarias: done

lib/src/model/documentation_comment.dart
szakarias: Much much better, thanks!

lib/src/model/documentation_comment.dart
szakarias: Yup, done.

lib/src/model/documentation_comment.dart
szakarias: done

lib/src/model/documentation_comment.dart
szakarias: done

lib/src/model/documentation_comment.dart
szakarias: done

lib/src/model/documentation_comment.dart
szakarias: Only `baseUri.resolve` throws this now (after the refactoring) while parsing the user input, so the error message correctly blames the example path. I've added a clarifying comment to Step 3.

lib/src/model/documentation_comment.dart
szakarias: You're right—`Uri.resolve` guarantees we stay within the root, so this shouldn't be hit currently.

However, I'd prefer to keep it as a safeguard. If future refactors or edge cases accidentally bypass the URI logic, this ensures we never silently allow access to files outside the package. I added a comment about this in the code.

lib/src/model/documentation_comment.dart
szakarias: Thanks for the suggestions on using `pathContext.toUri` and `fromUri`. I looked into it, and I'm concerned it breaks the 'virtual root' semantics we need, specifically on Windows.
- `toUri`: On Windows, this can introduce drive letters (e.g., `file:///C:/src/...`) even for relative paths. This leaks OS-specific details into the URI, which complicates the resolution logic.

- `fromUri`: Since this returns an absolute file system path, `pathContext.join(packagePath, absoluteResult)` would discard `packagePath` and return the absolute path (e.g., just `C:\foo.dart`), effectively escaping our package sandbox. 

To guarantee that resolution stays strictly within the package structure across all platforms, I've stuck to manually managing pathSegments. This treats the path as a list of tokens rather than a system path, ensuring packagePath remains the true root. WDYT?

lib/src/model/documentation_comment.dart
```diff
@@ -366,106 +368,128 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
 
-      var extension = resolved.extension;
+      var extension =
+          pathContext.extension(resolvedPath).replaceFirst('.', '').toLowerCase();
       var language = args['lang'] ?? extension;
       var indent = args['indent'] ?? 'strip';
 
       var content = file.readAsStringSync();
+      // Normalize line endings to \n.
+      content = content.replaceAll(RegExp(r'\r\n|\r'), '\n');
```
lrhn: Total micro-optimization: Can be shortened to `RegExp(r'\r\n?')`.

Or even;
```dart
  content = content.replaceAllMapped('\r', (match) =>
     content.startsWith('\n', match.end) ? '' : '\n');
```
No RegExp to compile, but a function call per match. Can't guess if it's faster.

More relevant, if examples become numerous: Don't create a new `RegExp` every time this function runs. Get it from a static variable, so it can be cached.

lib/src/model/documentation_comment.dart
```diff
@@ -366,106 +368,128 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+  /// - Otherwise, it is resolved relative to the file containing the
+  ///   documentation comment (rooted at the package root).
   /// - Resolution follows URI reference rules, ensuring that `..` segments
   ///   stop at the package root and never escape it.
-  /// - [filepath] can include URI-encoded characters (like `%20` for spaces).
+  /// - [exampleFilePath] must be a valid URI path (using `/` as separator), and can
+  ///   include URI-encoded characters (like `%20` for spaces).
+  ///
+  /// Returns the absolute path of the resolved file, or `null` if the path
+  /// could not be resolved.
```
lrhn: (How can a path fail to be resolved? Is it if the input is not _valid_?)

lib/src/model/documentation_comment.dart
```diff
@@ -366,106 +368,128 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
-    var relativeBasePosix = '/${pathContext.split(relativeBase).join('/')}';
+    // 2. Convert to a URI, rooted at the file system root.
+    // We prepend [pathContext.separator] so that the resulting URI is absolute,
+    // which ensures that [Uri.resolve] will clamp any '..' segments and
+    // keep them within the package root.
+    // We use a synthetic URI with a path starting at '/' and segments from
+    // [relativePath] to avoid platform-specific drive letters (like 'C:')
+    // entering the path segments on Windows.
+    var segments =
+        pathContext.split(relativePath).where((s) => s != '.').toList();
```
lrhn: If using `toUri`/`fromUri` (see below on how to maybe make it work), then this is just
```dart
var baseUri = pathContext.toUri(pathContext.join(pathContext.separator, relativePath));
```
In any case, not sure you need to remove `.`s eagerly, `Uri` will normalize the path anyway.

lib/src/model/documentation_comment.dart
lrhn: That ... is true. Damn. I though `join` was just concatenating, but it's actually "resolving".

Not to be deterred, we just make the path relative again:
```dart
  resolvedPath = pathContext.join(packagePath, pathContext.relative(
     pathContext.fromUri(resolvedUri), 
     from: pathContext.separator,
  ));
```

As for leading `C:`'s, that's more worrisome. Testing in dartpad didn't reveal any Windows issues, but that's because I used the default `p.posix`, etc, which have no current path.
It's the current directory that messes things up.

So it might help to use a context without a current path:
```dart
var contextNoCurrent = switch (pathContext.style) {
   p.Style.url => p.url,
   p.Style.posix => p.posix,
   p.Style.windows => p.windows,
   var style => p.Context(style: style),
 };
```
and then use that for `toUri`/`fromUri`. Should avoid the Windows `C:` prefix.
(I think.)

lib/src/model/documentation_comment.dart
lrhn: So that's the only call that needs to be inside the `try`/`catch`?

szakarias: yes.

lib/src/model/documentation_comment.dart
szakarias: done

lib/src/model/documentation_comment.dart
szakarias: Updated the comment.

lib/src/model/documentation_comment.dart
szakarias: I agree that we shouldn't fail silently.

I've decided to opt for a 'loud failure' strategy here instead of trying to be smart about tab widths or implicit defaults. If the indentation contains any tabs, the tool will now emit a warning and disable stripping entirely for that block.

This ensures the user is explicitly informed why the output looks 'wrong' (unstripped), without forcing us to support complex mixed-indentation logic.

lib/src/model/documentation_comment.dart
szakarias: Thanks, updated to ` RegExp( r'^[ \t]*\{@example(?:[ \t]+([^}\r\n]*))?\}[ \t]*$',
    multiLine: true,
  );
`

szakarias (commented):
Key changes:

- Switched `resolveExamplePath` to `toUri`/`fromUri` with logic to strip OS-specific roots (e.g. Windows drive letters), ensuring paths stay within the package "virtual root".
- Updated regex to enforce block-level usage and disallow inline newlines.
- Implemented "loud failure" for tabs in `_stripIndentation`: warns and disables stripping to avoid incorrect formatting.
- Added strict checks to reject non-`file` schemes.
- Optimized newline normalization `_injectExamples`
- Updated tests.

test/example_directive_test.dart
szakarias: done

test/example_directive_test.dart
szakarias: done

test/example_directive_test.dart
szakarias: done

test/example_directive_test.dart
szakarias: done

test/example_directive_test.dart
szakarias: refactored this code snippet.

test/example_directive_test.dart
szakarias: Yes, I added tests for these cases in the latest commit (`test_processesExampleDirective_inline_ignored` and `test_processesExampleDirective_withIndentStrip_mixedWhitespace`).

test/example_path_resolution_test.dart
szakarias: done

test/example_path_resolution_test.dart
szakarias: done

test/example_path_resolution_test.dart
szakarias: done

test/example_path_resolution_test.dart
szakarias: done

lib/src/model/documentation_comment.dart
```diff
@@ -320,6 +325,253 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+  ///
+  /// The file content is processed as follows:
+  /// - Line endings are normalized to `\n`.
+  /// - If `indent` is `strip` (the default), common indentation is removed from
+  ///   all lines. Whitespace-only lines are ignored for calculation and
+  ///   normalized to empty lines.
+  /// - If `indent` is `keep`, the indentation is left as-is.
+  /// - If any line has tab characters in its indentation and `indent` is
+  ///   `strip`, a warning is issued and indentation stripping is skipped to
+  ///   prevent broken formatting.
```
lrhn, line 354: 👍

lib/src/model/documentation_comment.dart
```diff
@@ -320,6 +325,253 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+        // Already warned about an invalid parameter if this happens.
+        return '';
+      }
+      if (args.rest.isEmpty) {
+        warn(PackageWarning.invalidParameter,
+            message: 'Must specify a file path for the @example directive.');
+        return '';
+      }
+
+      var filepath = args.rest.first;
```
lrhn, line 390: What if there is more than one element in `rest`?

lib/src/model/documentation_comment.dart
```diff
@@ -320,6 +325,253 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+    // e.g., "file:///C:/lib/...").
+    var baseUri =
+        pathContext.toUri(pathContext.join(pathContext.separator, relativePath));
+
+    // 3. Resolve using [Uri.resolve].
+    // [exampleFilePath] is interpreted as a URI reference.
+    String resolvedPath;
+    try {
+      var resolvedUri = baseUri.resolve(exampleFilePath);
+      if (resolvedUri.scheme != 'file') {
```
lrhn: `if (!resolvedUri.isScheme('file')) {`

(Doesn't have to create a new substring like the `scheme` getter might.)

lib/src/model/documentation_comment.dart
```diff
@@ -320,6 +325,253 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+    // converts to a file URI (potentially including a drive letter,
+    // e.g., "file:///C:/lib/...").
+    var baseUri =
+        pathContext.toUri(pathContext.join(pathContext.separator, relativePath));
+
+    // 3. Resolve using [Uri.resolve].
+    // [exampleFilePath] is interpreted as a URI reference.
+    String resolvedPath;
+    try {
+      var resolvedUri = baseUri.resolve(exampleFilePath);
```
lrhn: This doesn't check that `exampleFilePath` is a path.

Consider:
```dart
try {
   var exampleFileUri = Uri.parse(exampleFilePath);
   if (exampleFileUri.hasScheme || exampleFileUri.hasAuthority) {
     warn(... "must be a path, not have a scheme or authority ...");
     return null;
   }
   var resolvedUri = baseUri.resolveUri(exampleFileUri);
```

It should do the exact same thing for paths, but rule out, fx
```dart
/// {@example file:///etc/password}
```
or other ways to exit the package root.

(I also considered `.resolveUri(Uri(path: exampleFilePath))`, but if you want to put regions into the fragment part, that won't work, the above will. Also better error message.)

lib/src/model/documentation_comment.dart
```diff
@@ -320,6 +325,253 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+      warn(PackageWarning.invalidParameter,
+          message:
+              'Invalid path for @example directive ($exampleFilePath): ${e.message}');
+      return null;
+    }
+
+    resolvedPath = pathContext.normalize(resolvedPath);
+
+    // Security check: theoretically unreachable due to Uri resolution clamping,
+    // but kept as a safeguard against future refactoring bugs.
```
lrhn, line 515: (If not ruling out a scheme or authority above, then this could trigger. Consider adding a test with `{@example file:///something}` to see that it does trigger, before adding checks for that.)

lib/src/model/documentation_comment.dart
```diff
@@ -320,6 +325,253 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+    }
+
+    if (minIndent == null || minIndent == 0) {
+      return anyWhitespaceLineChanged ? lines.join('\n') : content;
+    }
+
+    return lines.map((line) {
+      if (line.isEmpty) return line;
+      return line.substring(minIndent!);
+    }).join('\n');
```
lrhn: (Could update in-place. Saves some allocation and doesn't need the `!` on `minIndent`.
```dart
  for (var i = 0; i < lines.length; i++) {
    var line = lines[i];
    if (line.isNotEmpty) lines[i] = line.substring(minIndent);
  }
  return lines.join('\n');
  ```
)

File: lib/src/model/documentation_comment.dart [hunk 48]
```diff
@@ -320,6 +325,253 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+  }
+
+  static final _exampleRegExp = RegExp(
```
lrhn: (The `_lineEndingRegExp` is a top-level variable, this one is static. Is there a reason for the difference?)

test/example_directive_test.dart
lrhn: 👍

lrhn (approved):
Only worry now is validating the path and over-mathcing non-standard whitespace.

Unicode, the gift that keeps giving!

lib/src/model/documentation_comment.dart
szakarias: done

lib/src/model/documentation_comment.dart
szakarias: done

lib/src/model/documentation_comment.dart
szakarias: done

lib/src/model/documentation_comment.dart
szakarias: No! Thanks!

lib/src/model/documentation_comment.dart
szakarias: done

lib/src/model/documentation_comment.dart
szakarias: Good catch, added a check.

Issue #4114, opened 2025-10-01
[private-named-parameters] Dartdoc Support


munificent:
This issue tracks the work needed to support ["Private Named Parameters"](https://github.com/dart-lang/language/blob/main/accepted/future-releases/2509-private-named-parameters/feature-specification.md) in Dartdoc.

This feature does impact Dartdoc in that we want a private named parameter to be documented as its public name. Given a class like:

```dart
class C {
  int _x;

  /// Make a new [C] with [_x].
  C({required this._x});
}
```

Then the dartdoc for the constructor should look something like:

> Make a new `C` with `x`.
>
> ```dart
> C({required int x});
> ```

lrhn:
Also, all visible textual references to the parameter in the actual generated DartDoc should use the public name, so
```dart
class Baz {
  final int foo, _bar;
  /// Creates `Baz` from [foo] and [_bar].
  new(this.foo, {this._bar}) : asset(_bar != 0);
}
```
will generate a documentation of:

```html
Creates <code>Baz</code> from
<code>foo</code> and <code>bar</code>.
```

(It would be nice if DartDoc did the same thing for positional private initializing or declaring parameters, otherwise I'll still have to do `(int foo): _foo = foo;`. That needs to check if there is another parameter which has the corresponding make already, but otherwise it would be the same.)

sigurdm:
@johnniwinther this sounds like the analyzer api should provide a "external name" of parameters somehow?

johnniwinther:
Yes. To the extent that we use AST nodes for printing the constructor parameters we need this. The element should show the public name in all cases.

munificent:
> [@johnniwinther](https://github.com/johnniwinther) this sounds like the analyzer api should provide a "external name" of parameters somehow?

The analyzer API does support this now, but it's a bit confusing. If a `FieldFormalParameterElement` comes from a private named parameter, then it's `name` field will be the corresponding public name. To get the original private name, there is a new  `privateName` getter. It's only on `FieldFormalParameterElement`. (Eventually, we'll need something similar for declaring parameters in primary constructors.)

PR #4202, opened 2026-02-12
Handle private named parameters in generated docs.

munificent:
Private named parameters affect Dartdoc in two places:

* When generating docs with a comment reference that refers to a private named parameter, use the corresponding public name.

* When generating a signature for a constructor with a private named parameter, use the corresponding public name for the parameter in the signature.

So given a class like:

```dart
class C {
  int _x;

  /// Make a new [C] with [_x].
  C({required this._x});
}
```

Then the Dartdoc looks like:

> Make a new C with `x`.
>
> `C({required int x});`

Fix https://github.com/dart-lang/dartdoc/issues/4114.

### DIFF
--- a/lib/src/model/parameter.dart
+++ b/lib/src/model/parameter.dart
@@ -7,7 +7,6 @@ import 'package:dartdoc/src/element_type.dart';
 import 'package:dartdoc/src/model/comment_referable.dart';
 import 'package:dartdoc/src/model/kind.dart';
 import 'package:dartdoc/src/model/model.dart';
-import 'package:dartdoc/src/model_utils.dart';
 
 class Parameter extends ModelElement with HasNoPage {
   @override
@@ -38,9 +37,8 @@ class Parameter extends ModelElement with HasNoPage {
   String? get documentedName {
     // TODO(rnystrom): Allow private named declaring parameters in primary
     // constructors here too.
-    if (element is FieldFormalParameterElement && element.hasPrivateName) {
-      // Use the public name.
-      return element.displayName.substring(1);
+    if (element case FieldFormalParameterElement(privateName: _?)) {
+      return element.name;
     }
 
     return null;
--- a/test/parameters_test.dart
+++ b/test/parameters_test.dart
@@ -22,15 +22,23 @@ class ParameterTest extends DartdocTestBase {
   String get libraryName => 'parameters';
 
   void test_formalParameter_private_named_referenced() async {
-    var library = await bootPackageWithLibrary('''
+    var library = await bootPackageWithLibrary(
+      '''
 class C {
   int _x;
 
   /// Makes with an [_x].
   C({required this._x});
 }
-''');
+''',
+      libraryPreamble: '// @dart=3.12',
+      additionalArguments: ['--enable-experiment=private-named-parameters'],
+    );
     var cNew = library.classes.named('C').constructors.named('C.new');
+    var parameter = cNew.parameters.first;
+
+    // The parameter's documented name is the public name of the field formal parameter.
+    expect(parameter.documentedName, equals('x'));
 
     // The parameter reference uses its public name.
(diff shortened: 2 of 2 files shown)


munificent:
Sorry for the redundancy with @lrhn's PR. I forgot that he did that a while back. :)

lib/src/model/parameter.dart
```diff
@@ -33,6 +34,18 @@ class Parameter extends ModelElement with HasNoPage {
   @override
   String? get href => null;
 
+  @override
+  String? get documentedName {
+    // TODO(rnystrom): Allow private named declaring parameters in primary
+    // constructors here too.
+    if (element is FieldFormalParameterElement && element.hasPrivateName) {
+      // Use the public name.
+      return element.displayName.substring(1);
```
szakarias: If possible, I would like us to get rid of the string manipulation and just get the public name directly from the analyzer. Given the docs on `FieldFormalParameterElement` the `name` property should give us the public name. (If it doesn't I think it's a bug in the analyzer)

munificent: > If possible, I would like us to get rid of the string manipulation and just get the public name directly from the analyzer. 

Good point.

> Given the docs on `FieldFormalParameterElement` the `name` property should give us the public name.

Oh, you're right it should. The fact that this is working is actually wrong. The `name` for a `FieldFormalParameterElement` declared with a private should be the public name only if the field formal is a valid private named parameter. For that to be true, the parameter needs to be in a library on the latest language version and when the "private-named-parameters" experiment is enabled. Neither of those is currently true for the dartdoc tests.

So the reason this test passed is because the `this._x` parameter was *not* a valid private named parameter, and thus `name` used the original private name, and thus `hasPrivateName` returns `true`. If the library was correctly using private named parameters, then the `name` would be `x`, and we wouldn't get into here at all. So this code does the wrong thing when private parameters are disabled and when they are enabled. Oops!

To fix this, I need to update the code here, but also:

1. Change the dartdoc tests to enable the experiment and use the latest language version for the code being analyzed.
2. Get a new version of analyzer published that exposes the `privateName` getter we need to tell whether a field formal is a valid private named parameter or not.
3. Update dartdoc to use that new analyzer version.

I'll start a thread with analyzer folks about 2 and then update this PR.

lrhn:
I'd want to land this anyway, because it works on positional parameters too. 

I'll bet the analyzer name-fiddling only provides the public name for named parameters, and then I'll have to do:
```dart
 Foo(int banana) : _banana = banana;
```
for all eternity anyway.

lrhn:
> Sorry for the redundancy with @lrhn's PR.

NP, this looks cleaner.

(BTW, should we consider moving the `dart-doc` and `dart_style` repos into the SDK repo, since they are so tightly coupled to the analyzer code base? Even removing experiment flags has extra steps when you have to wait for the SDK to be released first.)

munificent:
> (BTW, should we consider moving the `dart-doc` and `dart_style` repos into the SDK repo, since they are so tightly coupled to the analyzer code base? Even removing experiment flags has extra steps when you have to wait for the SDK to be released first.)

I've certainly thought about this for dart_style. But I do like that:

* I can use the normal GitHub PR UI for contributions to dart_style.
* It has it's own little issue tracker dedicated to formatter issues.

munificent:
> because it works on positional parameters too.

The current state of this PR might work for positional parameters, but it's also broken. The correct fix is to use the new `privateName` API in the element model. That will tell you if a formal parameter is a private named parameter and, if so, what the original private name is. (The `name` field on the element will tell you the corresponding public name.)

But that only works for private *named* parameters. For private positional parameters, the `privateName` will always be `null` and `name` will be the original private name. I admit the API is a little confusing here, but it seemed to cause the least refactoring in the rest of the many many places in analyzer/analysis_server that were already using `name`.

If we want dartdoc to remove the `_` from private positional parameters in API docs (which I would like too), we'll have to handle those specially.

munificent:
Looks great, thank you!

Issue #4203, opened 2026-02-12
`canonicalFor` does not seem to work for non-classes?


kevmoo:
```
{"level":"WARNING","message":"  warning: library says it is {@canonicalFor http_logging.badRequestMiddleware} but http_logging.badRequestMiddleware can't be canonical there\n    from http_serving: (file:///Users/kevmoo/github/functions-framework-dart/google_cloud/lib/http_serving.dart:1:1)"}
{"level":"WARNING","message":"  warning: library says it is {@canonicalFor http_logging.currentLogger} but http_logging.currentLogger can't be canonical there\n    from http_serving: (file:///Users/kevmoo/github/functions-framework-dart/google_cloud/lib/http_serving.dart:1:1)"}
{"level":"WARNING","message":"  warning: library says it is {@canonicalFor metadata.gceMetadataHost} but metadata.gceMetadataHost can't be canonical there\n    from general: (file:///Users/kevmoo/github/functions-framework-dart/google_cloud/lib/general.dart:1:1)"}
```

I'm naming things the same way as they are other places.

Will follow-up!

PR #4204, opened 2026-02-13
Significant improvements to canonicalization logic

kevmoo:
Fixes https://github.com/dart-lang/dartdoc/issues/4203

* elements documented in their intended libraries via `@canonicalFor` overrides are now correctly placed in their respective directories.
* improved SDK interlinking to ensure elements re-exported from standard libraries (like `dart:core`) generate absolute remote links when appropriate.
* fixed a bug where property accessors (getters/setters) were incorrectly compared during canonical library lookups.
* refined scoring heuristics with definitive tie-breakers for standard SDK schemes and defining libraries.

### DIFF
--- a/tool/task.dart
+++ b/tool/task.dart
@@ -214,13 +214,29 @@ Stream<String> _fileLines(Directory dir, {required Set<String> extensions}) {
 }
 
 Future<void> runBuildbot() async {
-  await analyzeTestPackages();
-  await analyzePackage();
-  await validateFormat();
-  await validateBuild();
-  await runTryPublish();
-  await runTest();
-  await validateDartdocDocs();
+  await _section('Analyze test packages', analyzeTestPackages);
+  await _section('Analyze package', analyzePackage);
+  await _section('Validate format', validateFormat);
+  await _section('Validate build', validateBuild);
+  await _section('Run try publish', runTryPublish);
+  await _section('Run test', runTest);
+  await _section('Validate dartdoc docs', validateDartdocDocs);
+}
+
+Future<void> _section(String title, Future<void> Function() body) async {
+  if (Platform.environment['GITHUB_ACTIONS'] == 'true') {
+    print('::group::$title');
+    try {
+      await body();
+    } catch (e) {
+      print('::error::$e');
+      rethrow;
+    } finally {
+      print('::endgroup::');
+    }
+  } else {
+    await body();
+  }
 }
 
 Future<void> runClean() async {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
-## 9.0.3-wip
+## 9.0.3
+* Remove usage of `EmbedderYamlLocator` and `EmbedderSdk.new`, internal
+  analyzer APIs which were removed in analyzer 10.1.0.
 * Significant improvements to canonicalization logic:
     * elements documented in their intended libraries via `@canonicalFor` overrides are now correctly placed in their respective directories.
     * improved SDK interlinking to ensure elements re-exported from standard libraries (like `dart:core`) generate absolute remote links when appropriate.
--- a/dartdoc_options.yaml
+++ b/dartdoc_options.yaml
@@ -1,4 +1,4 @@
 dartdoc:
   linkToSource:
     root: '.'
-    uriTemplate: 'https://github.com/dart-lang/dartdoc/blob/v9.0.3-wip/%f%#L%l%'
+    uriTemplate: 'https://github.com/dart-lang/dartdoc/blob/v9.0.3/%f%#L%l%'
--- a/lib/src/version.dart
+++ b/lib/src/version.dart
@@ -1,1 +1,1 @@
-const packageVersion = '9.0.2';
+const packageVersion = '9.0.3';
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,5 +1,5 @@
 name: dartdoc
-version: 9.0.3-wip
+version: 9.0.3
 description: A non-interactive HTML documentation generator for Dart source code.
 repository: https://github.com/dart-lang/dartdoc
 


lib/src/model/canonicalization.dart
```diff
@@ -245,6 +275,11 @@ class _ScoredCandidate implements Comparable<_ScoredCandidate> {
 /// A reason that a candidate's score is changed.
 enum _Reason {
   canonicalFor('marked @canonicalFor'),
+  samePackage('is in the same package'),
```
srawlins, line 287: I like this granularization.

lib/src/model/model_element.dart
```diff
@@ -665,8 +709,20 @@ abstract class ModelElement
...
+    var documentedWhere = canonicalLibrary.package.documentedWhere;
+    if (documentedWhere == DocumentLocation.remote) {
+      assert(canonicalLibrary.package.baseHref != null);
+      return '${canonicalLibrary.package.baseHref}$filePath';
+    }
+    if (documentedWhere == DocumentLocation.local) {
+      var packageBaseHref = canonicalLibrary.package.baseHref;
+      return '$packageBaseHref$filePath';
+    }
+    // documentedWhere == DocumentLocation.missing
```
srawlins: Delete?

test/canonical_for_test.dart
srawlins, line 1: ❤️ 🙏

srawlins (commented):
LGTM, but please wait for @szakarias 's review 😁 

Thanks much!

lib/src/model/model_element.dart
kevmoo: yep

GEMINI.md
```diff
@@ -0,0 +1,29 @@
+# Local Verification for AI Agents
```
kevmoo, line 1: @srawlins – as a bonus!

tool/task.dart
```diff
@@ -214,13 +214,26 @@ Stream<String> _fileLines(Directory dir, {required Set<String> extensions}) {
...
+  await _section('Analyze test packages', analyzeTestPackages);
+  await _section('Analyze package', analyzePackage);
+  await _section('Validate format', validateFormat);
+  await _section('Validate build', validateBuild);
+  await _section('Run try publish', runTryPublish);
+  await _section('Run test', runTest);
+  await _section('Validate dartdoc docs', validateDartdocDocs);
+}
+
+Future<void> _section(String title, Future<void> Function() body) async {
```
kevmoo: Check that out!

kevmoo: Image

lib/src/model/model_element.dart
```diff
@@ -577,6 +577,38 @@ abstract class ModelElement
...
+  /// Whether this element is defined in an internal library.
+  late final bool isFromInternalLibrary = () {
+    var libraryElement = element.library;
+    if (libraryElement == null) return false;
+    var uri = libraryElement.firstFragment.source.uri;
+    if (uri.isScheme('dart')) {
+      var segments = uri.pathSegments;
+      if (segments case [var firstSegment, ...]
+          when firstSegment.startsWith('_') ||
+              firstSegment == 'nativewrappers') {
```
kevmoo, line 599: This seems a bit specific – tis an agent.
I think this was to do w/ hitting a bunch of weird test errors with dart:_XYZ types

test/options_test.dart
```diff
@@ -213,8 +213,11 @@ class Foo {
         orderedEquals(['library_1']));
     final foo =
         packageGraph.localPackages.first.libraries.first.classes.named('Foo');
-    // The name is not linked, but also does not error.
-    expect(foo.declaredFields.first.modelType.linkedName, 'Client?');
+    // The name is linked because it's from a public package.
+    expect(
+      foo.declaredFields.first.modelType.linkedName,
+      '<a href="%%__HTMLBASE_dartdoc_internal__%%package-http_http/Client-class.html">Client</a>?',
```
kevmoo, line 219: This change is WEIRD to me. Please double check.

tool/mustachio/builder.dart
```diff
@@ -61,6 +63,14 @@ Future<void> build(
     root: root,
     sourcePath: sourcePath,
   );
+
+  var formatter = DartFormatter(languageVersion: Version(3, 6, 0));
+  try {
+    aotRenderersContents = formatter.format(aotRenderersContents);
```
kevmoo, line 69: Seems weird that we didn't do this up front, no?

kevmoo: Like, why didn't we do this already?

lib/src/model/library.dart
```diff
@@ -486,24 +486,11 @@ class Library extends ModelElement with TopLevelContainer {
...
-        getter: getter,
-        setter: setter,
-      ).fullyQualifiedName;
-    }).toSet();
+    return libraryMembers
+        .expand((member) => [
+              member.originalFullyQualifiedName,
+              '$name.${member.qualifiedName}',
+            ])
+        .toSet();
```
lrhn: Drive-by-comment: Can be a literal:
```dart
 {for (var member in libraryMembers) ...[
    member.originalFullyQualifiedName, 
    '$name.${member.qualifiedName}',
  ]};
```

kevmoo: awesome! integrated!

tool/task.dart
kevmoo: I moved it to another pr. trying to avoid too much noise here.

lib/src/model/canonicalization.dart
```diff
@@ -11,11 +11,14 @@ const int _separatorChar = 0x3B;
...
-  if (library == null) return null;
-  var thisAndExported =
-      modelElement.packageGraph.libraryExports[library.element];
-  if (thisAndExported == null) {
+  var libraryElement = modelElement.element.library;
+  if (libraryElement == null) return null;
+  var libraryExports = modelElement.packageGraph.libraryExports[libraryElement];
+  var candidateList = {
+    ...?libraryExports,
+    if (modelElement.library case var library?) library,
```
szakarias, line 23: I am concerned about a potential side effect here.

By explicitly adding the defining library to the candidateList and subsequently giving it a strong score boost `(4.0 for _Reason.sameLibrary)` in line 173, isn't there a risk that this will cause SDK re-exports to be mis-canonicalized?

For example, if a local package re-exports `Future` from `dart:async`, it seems like `dart:async` would now aggressively 'win' as the canonical home. If we aren't generating SDK docs, wouldn't this result in broken or missing links that previously worked?

@srawlins, you probably know better than me whether this behavior change is intended or if we might need a safeguard for SDK elements here?

srawlins: Good catch; this line may not be needed, Kevin; does a test fail if it is removed? That'd be the simplest resolution 😁 

However, we currently have an expectation that the canonical link is to the Dart SDK. We have this test in `libraries_test.dart`:

```dart
  void test_library_containsClassWithSameNameAsDartSdk() async {
    var library = await bootPackageWithLibrary(
      'export "dart:io";',
      libraryFilePath: 'lib/library.dart',
    );

    expect(library.classes.named('FileSystemEntity').linkedName,
        '<a href="$dartSdkUrlPrefix/dart-io/FileSystemEntity-class.html">FileSystemEntity</a>');
  }
```

kevmoo: I had the robot dig in a bit

See https://github.com/dart-lang/dartdoc/pull/4204/changes/1e362eb8d8606df5acd5e1fc4b5273af25759535

kevmoo: I also asked gemini to document the code a bit more https://github.com/dart-lang/dartdoc/pull/4204/changes/25a3eee408515ff0c54069164bb26b17e6fa6dbe

PR #4206, opened 2026-02-14
Add sanity in the buildbot output

kevmoo:


szakarias (approved):
nice, LGTM

kevmoo:
This is really incomplete. We should add this more places – but maybe a good starting spot?

PR #4226, opened 2026-03-07
Fix #4207: Preserve display text for reference-style documentation links

### LINKED ISSUE #4207: Allow controlling text of documentation link.
If you write `[Foo.bar]`, you link to `Foo.bar` *and* the link text is `Foo.bar`.

There is no way to link to `Foo.bar` with a text of, fx, `Foo.current.bar`. Or anything else.

Currently if you write, as I did: ``[`Platform.current.nativePlatform`][Platform.nativePlatform]``,
the link does link to the documentation for `Platform.nativePlatform`, but the text is still `Platform.nativePlatform` too.

I don't know where the link text went. It's not in the output at all.
I'm pretty sure this is actually a bug, because DartDoc tries to *not* intercept Markdown reference-style links (`[text][reference]`) like:
```markdown
The [banana][Bananas] is fine.

[Bananas]: http://example.com/bananas "example of bananas"
```
so it should have ignored the `[Platform.nativePlatform]`. 

_It would be nice_ if it had actually worked. When DartDoc sees a markdown reference-style link, and it cannot find the target for the link (no `Platform.nativePlatform` link target defined in the document) *then* DartDoc should take over and try to interpret it as a doc element link.
It would let you make nice links to any documentable element, with full control over the link text.

Then a normal dartdoc link like `[Platform.nativePlatform]` is just short for ``[`Platform.nativePlatform`][Platform.nativePlatform]``.

NamanGoyalK:
Fixes #4207

## Description
This PR fixes an issue where Dartdoc ignores the custom text provided in a Markdown reference-style link (`[text][target]`) and instead aggressively overwrites it with the target's element name. 

With this fix, writing `[Platform.current][Platform.nativePlatform]` will correctly link to `Platform.nativePlatform` while displaying `Platform.current` as the link text.

## Implementation Details
Instead of heavily refactoring Dartdoc's core link resolver (`_makeLinkNode` / `getMatchingLinkElement`), which handles complex edge cases like callable hints, shadowing, and inherited documentation, this fix introduces a localized, safe AST interception via `package:markdown`.

* Added `_LinkSyntaxWithPreservedReferenceText` which extends `md.LinkSyntax`.
* It intercepts the node creation in the `close()` method. 
* If the `linkResolver` has overwritten the text to match the target label, it safely swaps the `md.Text` child back to the user's intended `referenceText`.
* Implemented `_parseFullReferenceLabel` to safely extract the target label while accounting for escaped characters and ignoring collapsed links (`[text][]`).

This approach guarantees that standard Markdown links remain unaffected while accurately restoring the text for Dartdoc-resolved elements.

## Testing
Added comprehensive tests in `test/reference_style_links_test.dart`:
* Verified that standard defined Markdown reference links are untouched.
* Verified that `[text][target]` correctly falls back to Dartdoc link resolution.
* Verified that the first label (`text`) is successfully preserved in the generated HTML and the fallback target label is not displayed.

## Checklist
- [x] I have read the [Contributor Guide](https://github.com/dart-lang/dartdoc/blob/main/CONTRIBUTING.md).
- [x] I have added tests to cover my changes.
- [x] All tests pass (`dart test`).

### DIFF
--- a/testing/test_package/lib/fake.dart
+++ b/testing/test_package/lib/fake.dart
@@ -1162,3 +1162,12 @@ class TypeParameterThingsExtendedQ
 /// and [fParam.fParamC].
 void set aSetterWithFunctionParameter(
     bool fParam(int fParamA, String fParamB, String fParamC)) {}
+
+
+class VisualFallbackTarget {
+  static int get bar => 0;
+}
+
+/// Visual fallback link check:
+/// [VisualFallbackTarget.current.bar][VisualFallbackTarget.bar]
+void visualFallbackLinkCheck() {}
--- a/testing/test_package/lib/fake.dart
+++ b/testing/test_package/lib/fake.dart
@@ -1163,7 +1163,6 @@ class TypeParameterThingsExtendedQ
 void set aSetterWithFunctionParameter(
     bool fParam(int fParamA, String fParamB, String fParamC)) {}
 
-
 class VisualFallbackTarget {
   static int get bar => 0;
 }
--- a/lib/src/markdown_processor.dart
+++ b/lib/src/markdown_processor.dart
@@ -261,8 +261,8 @@ class MarkdownDocument extends md.Document {
   /// Creates a document which resolves comment references as stemming from
   /// [element].
   factory MarkdownDocument.withElementLinkResolver(Warnable element) {
-    final linkResolver =
-        (String name, [String? _]) => _makeLinkNode(name, element);
+    md.Node? linkResolver(String name, [String? _]) =>
+        _makeLinkNode(name, element);
     return MarkdownDocument._(
       inlineSyntaxes: _markdownSyntaxes(linkResolver),
       blockSyntaxes: _markdownBlockSyntaxes,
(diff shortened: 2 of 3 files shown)


testing/test_package/lib/fake.dart
```diff
@@ -1162,3 +1162,12 @@ class TypeParameterThingsExtendedQ
...
     bool fParam(int fParamA, String fParamB, String fParamC)) {}
+
+
+class VisualFallbackTarget {
+  static int get bar => 0;
+}
+
+/// Visual fallback link check:
+/// [VisualFallbackTarget.current.bar][VisualFallbackTarget.bar]
+void visualFallbackLinkCheck() {}
```
kevmoo: you'll get yelled at for formatting here...

NamanGoyalK: Ah, classic! Fixed in the latest commit. Thanks for catching that! Won't forget again!

NamanGoyalK: @kevmoo Huge thanks for the blazing-fast workflow approvals! Really appreciate you helping me get that board green so quickly. I also just set up act locally, so the CI bots won't be sniping me on formatting or lints in the future. Have a great weekend!

kevmoo:
You sent a PR with TESTS. I gotta help you out.

@srawlins is likely the person who will need to review, though.

srawlins:
I'll certainly take a look, but I believe @szakarias is the primary owner of Dartdoc.

srawlins (approved):
This looks good to me. The parser work is a little dicey, but I don't know of a way around it (I think maybe that's just par-for-the-course for parser work 😅 ).

NamanGoyalK:
Thanks for the review and the approval, @srawlins! Completely agree on the parser work, it was a bit of a puzzle to find a localized way to handle it without over-engineering the core resolver.

Standing by for whenever @szakarias has a moment for the final look!

PR #4228, opened 2026-03-14
Require analyzer 12.0.0, with breaking changes

scheglov:
This change refers to not yet published `analyzer 12.0.0`, because I cannot land corresponding [analyzer CL](https://dart-review.googlesource.com/c/sdk/+/486080) (and then publish) without pulling into the Dart SDK a version of `dartdoc` (and `dart_style`, see https://github.com/dart-lang/dart_style/pull/1815) that is compatible with this breaking change version of the analyzer. So, I want to land this PR with a version of `dartdoc` that is compatible with `analyzer 12.0.0` (to be), and the pull this git commit into [the CL](https://dart-review.googlesource.com/c/sdk/+/486080).

I tested it locally with
```yaml
dependency_overrides:
  _fe_analyzer_shared:
    path: /Users/scheglov/Source/Dart/sdk.git/sdk/pkg/_fe_analyzer_shared
  analyzer:
    path: /Users/scheglov/Source/Dart/sdk.git/sdk/pkg/analyzer
  dart_style:
    path: /Users/scheglov/dart/dart_style
```
Obviously, GitHub bots will be red until `analyzer 12.0.0` is published.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 9.0.5-dev
+## 9.0.5-wip
 
 * Require analyzer 12.0.0 APIs.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 9.0.5-dev
+
+* Require analyzer 12.0.0 APIs.
+
+
 ## 9.0.4
 * Fix a bug where experiment flags passed via command-line were ignored.
 * Support specifying experimental features in `dartdoc_options.yaml`.
--- a/lib/src/model/package_graph.dart
+++ b/lib/src/model/package_graph.dart
@@ -233,11 +233,13 @@ class PackageGraph with Referable {
             }
           case EnumDeclaration():
             if (declaration.declaredFragment?.element.isPublic ?? false) {
-              for (var constant in declaration.body.constants) {
-                _populateModelNodeFor(constant);
-              }
-              for (var member in declaration.body.members) {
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 9.0.5-dev
```
kevmoo: We don't do `-wip` here?

scheglov: Done.

kevmoo (approved):
I wish we had a dance here where we didn't have to land on a sea of red. We should brainstorm a solution here.

@scheglov – you've done this before, right? I want to support you cranking on these silly bits on the weekend, but it makes me nervous.

scheglov:
It is definitely not the ideal process, but it is the best I can think of for breaking changes to AST.

My initial approach to this was to use `useBody2` variable that configures the analyzer parsing and AST behavior. When it is `true` you are allowed to use `get body2` and get body throws. And when it is `false`, `body2` throws, and `body` works. And then every place where you use body you have to check this flag, and invoke the right one. Doing it in non-breaking way does not work, because `MethodDeclaration.parent` have to return one or another.

And to support this flag now every client has to support it.
And then we would have to dance to rename back `body2` into `body`.
It seems to me that it is just less overall pain to do it all at once.

Yes, we did this once before, for initial primary constructor AST changes.
Was fairly painless, IIRC?

srawlins (approved):

PR #4231, opened 2026-04-01
Fix RangeError in doc-imports stripping by detecting index overshoot.

kevmoo:
Detect when absolute offsets from the analyzer overshoot the normalized string length
(due to intermediate non-doc comments like `// ignore` breaking the sequence, or \r\n line endings)
and skip stripping to avoid a crash. Added regression tests.

### DIFF
--- a/lib/src/model/documentation_comment.dart
+++ b/lib/src/model/documentation_comment.dart
@@ -866,7 +866,17 @@ mixin DocumentationComment implements Warnable, SourceCode {
       var buffer = StringBuffer();
       if (commentData.docImports.isEmpty) return content;
       var firstDocImport = commentData.docImports.first;
-      buffer.write(content.substring(0, firstDocImport.offset - commentOffset));
+
+      var relativeOffset = firstDocImport.offset - commentOffset;
+      if (relativeOffset >= content.length) {
+        // Safe fallback if indices drift due to line ending normalization.
+        warn(PackageWarning.deprecated,
+            message:
+                'Index drift detected when removing doc imports. This is often caused by `// ignore` blocks or `\\r\\n` line endings in your documentation. Convert file to use `\\n` and remove intermediate non-doc comments to ensure doc imports are removed properly.');
+        return content;
+      }
+
+      buffer.write(content.substring(0, relativeOffset));
       var offset = firstDocImport.end - commentOffset;
       for (var docImport in commentData.docImports.skip(1)) {
         buffer
--- a/test/documentation_comment_test.dart
+++ b/test/documentation_comment_test.dart
@@ -277,6 +277,33 @@ Three.'''));
     expect(doc, equals(''));
   }
 
+  void test_docImport_overshoot() async {
+    // Force \r\n line endings to simulate index drift if the analyzer normalizes it to \n.
+    // We use many lines to ensure the shift overshoots the string length and throws RangeError.
+    var lines = List.generate(50, (i) => '/// Line $i').join('\n');
+    await writePackageWithCommentedLibrary('''
+$lines
+/// @docImport 'dart:async' as async;
+'''.replaceAll('\n', '\r\n'));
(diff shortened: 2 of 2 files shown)


kevmoo:
Before

```

dartdoc 9.0.4 (/Users/kevmoo/github/dartdoc/bin/dartdoc.dart) failed: ParallelWaitError: RangeError (end): Invalid value: Not in inclusive range 0..1683: 1719
#0      RangeError.checkValidRange (dart:core/errors.dart:392:9)
#1      _StringBase.substring (dart:core-patch/string_patch.dart:437:27)
#2      DocumentationComment._stripDocImports (package:dartdoc/src/model/documentation_comment.dart:869:28)
#3      DocumentationComment.processComment (package:dartdoc/src/model/documentation_comment.dart:101:18)
#4      DocumentationComment.precacheLocalDocs (package:dartdoc/src/model/documentation_comment.dart:999:33)
#5      PackageGraph._precacheLocalDocs (package:dartdoc/src/model/package_graph.dart:174:27)
#6      PackageGraph.initializePackageGraph (package:dartdoc/src/model/package_graph.dart:120:11)
#7      PubPackageBuilder.buildPackageGraph (package:dartdoc/src/model/package_builder.dart:161:22)
<asynchronous suspension>
#8      Dartdoc.generateDocsBase (package:dartdoc/src/dartdoc.dart:193:24)
<asynchronous suspension>
#9      Dartdoc.generateDocs (package:dartdoc/src/dartdoc.dart:237:24)
<asynchronous suspension>
#10     Dartdoc.executeGuarded.<anonymous closure> (package:dartdoc/src/dartdoc.dart:269:9)
<asynchronous suspension>
```

srawlins:
Comments like:

```dart
/// Line 1
// ignore: something
/// Line 2
```

are perfectly legitimate, and can solve a purpose. (Or rather, they are legitimate _if_ they serve a purpose I guess.) This is also legitimate:

```dart
/// Line 1
// TODO(srawlins): something
/// Line 2
```

I think we need a way to handle these.

kevmoo:
@srawlins – I totally agree!

At least we don't crash now. And we have a stake in the ground around this issue.

lib/src/model/documentation_comment.dart
```diff
@@ -866,7 +866,17 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
       if (commentData.docImports.isEmpty) return content;
       var firstDocImport = commentData.docImports.first;
-      buffer.write(content.substring(0, firstDocImport.offset - commentOffset));
+
+      var relativeOffset = firstDocImport.offset - commentOffset;
+      if (relativeOffset >= content.length) {
+        // Safe fallback if indices drift due to line ending normalization.
+        warn(PackageWarning.deprecated,
+            message:
+                'Index drift detected when removing doc imports. This is often caused by `// ignore` blocks or `\\r\\n` line endings in your documentation. Convert file to use `\\n` and remove intermediate non-doc comments to ensure doc imports are removed properly.');
```
srawlins, line 875: I don't think we should put in the advice to remove intermediate non-doc comments.

Nor should we advise that they change all of their line endings in their codebase because dartdoc has a bug.

The only thing I can think of to write here is to upvote the issue describing how dartdoc falls on its face in these cases.

kevmoo: I mean, stopping crashing and telling the user where the error is coming from kinda stops the bleeding. 🤷

srawlins:
@kevmoo thanks for finding this and working around the crash! ❤️  I didn't know about this bug.

I've got a WIP that strips the doc-imports more carefully and can hopefully avoid the warning, so I'd like to wait a few days before landing this. I'll hopefully upload a PR today or tomorrow. 🙏

kevmoo:
Closing out! @srawlins fixed the root issue!

PR #4233, opened 2026-04-02
Fix dartdoc canonicalization for explicit getters and prioritize @canonicalFor

kevmoo:
- Update model_element.dart to check `canonicalLibraryCandidate` (and thus `{@canonicalFor}` tags) before falling back to the local public library shortcut.
- Update canonicalization.dart to provide a name and library-based fallback for comparing elements, fixing issues where explicit top-level getters failed to match.
- Add test_canonicalFor_getter in canonical_for_test.dart to verify explicit getter canonicalization.

### DIFF
--- a/lib/src/model/canonicalization.dart
+++ b/lib/src/model/canonicalization.dart
@@ -39,18 +39,28 @@ Library? canonicalLibraryCandidate(ModelElement modelElement) {
     return null;
   }
 
+  final targetElement = _normalize(topLevelElement);
+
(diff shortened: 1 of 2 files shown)


srawlins:
Is this fixing a specific bug that is tracked anywhere?

kevmoo:
> Is this fixing a specific bug that is tracked anywhere?

I don't think so. I looked. I hit it yesterday and instead of filing an issue I got it fixed. 😀

srawlins (approved):
This looks great, very simple.

PR #4234, opened 2026-04-02
Improve @canonicalFor validation with close-match suggestions

kevmoo:
Suggest matches with the same element name but different library prefixes.

Update ignoredCanonicalFor template to use {0} once to avoid duplication.

Add test case in canonical_for_test.dart.

There are cases where

```dart
/// {@canonicalFor this_doesnt_exst.foo}
```

Will just fail silently and put `foo` in the wrong place.

### DIFF
--- a/lib/src/model/library.dart
+++ b/lib/src/model/library.dart
@@ -458,7 +458,22 @@ class Library extends ModelElement with TopLevelContainer {
         if (!elementNames.contains(elementName)) elementName,
     };
     for (var notFound in notFoundInAllModelElements) {
-      warn(PackageWarning.ignoredCanonicalFor, message: notFound);
+      var parts = notFound.split('.');
+      if (parts.length == 2) {
+        var elementName = parts[1];
+        var possibleMatches = elementNames
+            .where((name) => name.endsWith('.$elementName'))
+            .toList();
+        if (possibleMatches.isNotEmpty) {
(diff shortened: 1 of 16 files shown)


srawlins:
Does this fix a specific issue?

kevmoo:
> Does this fix a specific issue?

Yes it did! If you just make up a library name it silently fails in many cases.
You don't get any warnings. It just puts the element in the wrong place!

kevmoo:
@srawlins – updated the description. See the test. It helps user fix this!

lib/src/model/library.dart
```diff
@@ -458,7 +458,17 @@ class Library extends ModelElement with TopLevelContainer {
...
     for (var notFound in notFoundInAllModelElements) {
-      warn(PackageWarning.ignoredCanonicalFor, message: notFound);
+      var message = notFound;
+      var lastDot = notFound.lastIndexOf('.');
+      if (lastDot != -1 && lastDot < notFound.length - 1) {
+        var suffix = notFound.substring(lastDot);
+        var possibleMatches =
+            elementNames.where((name) => name.endsWith(suffix));
+        if (possibleMatches.isNotEmpty) {
+          message = '$notFound (did you mean ${possibleMatches.join(', ')}?)';
```
srawlins, line 468: Huge improvement, thanks!

srawlins (approved):
Lovely!

PR #4240, opened 2026-04-13
Fix link generation for remote and re-exported symbols

### LINKED ISSUE #4075: Generated API reference points to invalid links for Flutter/Dart classes leading to 404s

### LINKED ISSUE #4239: ImageProvider and other elements from flutter/painting.dart package are pointing to wrong links in docs.
**Issue**
In our project we use a 3rd party package ([skeletonizer](https://pub.dev/packages/skeletonizer)) that exports '**package:flutter/painting.dart'** in one of its files ([skeletonizer.dart](https://github.com/Milad-Akarie/skeletonizer/blob/main/lib/skeletonizer.dart))
This seems to confuse dartdoc’s linking/canonicalization and causes the Flutter elements that are part of 'flutter/painting.dart' package to point to wrong links. More specifically, the doc links of such elements point to the skeletonizer docs instead of the official Flutter docs.

**Example**
One widget class that is affected by that is **ImageProvider**, which when the user clicks on it in dart docs, then it links to the following page:
Image

The url of the ImageProvider docs should be https://api.flutter.dev/flutter/painting/ImageProvider-class.html and not https://api.flutter.dev/flutter/skeletonizer/ImageProvider-class.html (which does not exist).

**Question**
Is this expected behavior in dartdoc’s canonicalization/linking model, or should re-exported external SDK symbols be handled differently?

kevmoo:
* Remove incorrect href overrides in TopLevelVariable, ModelFunctionTyped, and Typedef that were hardcoding local paths.
* Correct EnumField.href to use canonicalLibrary!.package.baseHref.
* Fix canonicalLibraryCandidate to consider the defining library as a candidate.
* Fix ModelElement.canonicalLibrary to use defining library for locality check.

Fixes #4075 
Fixes #4239

### DIFF
--- a/lib/src/model/canonicalization.dart
+++ b/lib/src/model/canonicalization.dart
@@ -14,14 +14,14 @@ Library? canonicalLibraryCandidate(ModelElement modelElement) {
   var libraryElement = modelElement.element.library;
   if (libraryElement == null) return null;
   var libraryExports = modelElement.packageGraph.libraryExports[libraryElement];
+  var definingLibrary =
+      modelElement.packageGraph.findButDoNotCreateLibraryFor(libraryElement);
   var candidateList = {
     ...?libraryExports,
-    // When the element is defined in a library that is not documented, the
-    // `libraryExports` map will not contain that library. However, we still want
-    // to consider the defining library as a candidate if it happens to be
-    // documented.
     if (modelElement.library case var library?) library,
+    if (definingLibrary != null) definingLibrary,
   };
+
   if (candidateList.isEmpty) {
     return null;
   }
--- a/lib/src/model/enum.dart
+++ b/lib/src/model/enum.dart
@@ -107,7 +107,7 @@ class EnumField extends Field {
     }
     assert(canonicalEnclosingContainer == enclosingElement);
     assert(canonicalLibrary != null);
-    return '${package.baseHref}${canonicalLibrary!.dirName}/'
+    return '${canonicalLibrary!.package.baseHref}${canonicalLibrary!.dirName}/'
         '${enclosingElement.fileName}';
   }
 
--- a/lib/src/model/model_element.dart
+++ b/lib/src/model/model_element.dart
@@ -500,16 +500,7 @@ abstract class ModelElement
     // This is not accurate if we are still constructing the Package.
     assert(packageGraph.allLibrariesAdded);
 
-    var definingLibraryElement = element.library;
-    var definingLibrary = definingLibraryElement == null
-        ? null
-        : packageGraph.findButDoNotCreateLibraryFor(definingLibraryElement);
-    var definingLibraryIsLocalPublic = definingLibrary != null &&
-        packageGraph.localPublicLibraries.contains(definingLibrary);
-    var possibleCanonicalLibrary = definingLibraryIsLocalPublic
-        ? library
-        : canonicalLibraryCandidate(this);
-
+    var possibleCanonicalLibrary = canonicalLibraryCandidate(this);
     if (possibleCanonicalLibrary != null) return possibleCanonicalLibrary;
 
     if (this case Inheritable(isInherited: true)) {
--- a/lib/src/model/model_element.dart
+++ b/lib/src/model/model_element.dart
@@ -500,8 +500,12 @@ abstract class ModelElement
     // This is not accurate if we are still constructing the Package.
     assert(packageGraph.allLibrariesAdded);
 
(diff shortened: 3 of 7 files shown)


kevmoo:
*More explanation from Gemini*

Here is the "color" on what was going on. It was actually a perfect storm of three different broken assumptions layered on top of each other!

### The Scenario That Was Broken
Imagine you have Package **Two** that depends on a remote package **One** (like the SDK or a published package). 
Package **Two** re-exports a symbol (like a top-level variable or a typedef) defined in Package **One**.

When you build documentation for Package **Two** with `--link-to-remote` enabled, you expect the link for that re-exported symbol to point to Package **One**'s remote documentation (e.g., `https://pub.dev/...` or `https://api.dart.dev/...`).

Instead, `dartdoc` was generating a **local link** pointing into Package **Two**'s output directory (e.g., `%%__HTMLBASE_dartdoc_internal__%%two/topLevelVariable.html`). Since Package **Two** doesn't actually generate documentation for that symbol (it just re-exports it), the link was broken!

---

### The 3 Broken Assumptions

#### 1. The "We know better than our parent" Assumption
`TopLevelVariable`, `ModelFunctionTyped`, and `Typedef` were overriding the `href` getter and hardcoding a local path: `${package.baseHref}$filePath`. 
*   **The broken assumption:** They assumed that they would always be documented locally. By hardcoding this, they bypassed the smart logic in `ModelElement.href` that checks if the canonical element is remote and uses the remote package's `baseHref`.

#### 2. The "If it's here, it belongs here" Assumption
Once we removed the overrides and let `ModelElement.href` do its job, it failed too! In `ModelElement.canonicalLibrary`, it was doing this check:
```dart
var definingLibraryIsLocalPublic = packageGraph.localPublicLibraries.contains(library);
```
*   **The broken assumption:** Here, `library` refers to `this.library` (the library where we *found* the reference, i.e., Package **Two**). Because Package **Two** is local and public, `dartdoc` assumed: *"Hey, we are documenting this library locally, so let's just say this library is the canonical home for this symbol!"* 
*   It didn't check if the *defining* library (`element.library`) was local. I changed it to check the defining library instead.

#### 3. The "You must be on the guest list" Assumption
Once we forced it to look for a better canonical library candidate (because the defining library was remote), `canonicalLibraryCandidate` failed to find Package **One**!
*   **The broken assumption:** It assumed that the only candidates for being a canonical library were libraries that explicitly *exported* the symbol (found in the `libraryExports` map) plus the current library. 
*   But because Package **One** is remote, we didn't build an export map for it! So the defining library itself was not even on the list of candidates to be scored! I fixed this by explicitly adding the defining library to the candidate list.

---

So, by removing the hardcoded paths, fixing the locality check, and inviting the defining library to the candidate party, we finally got it to generate the correct remote URL!

kevmoo:
CC @sigurdm RE https://github.com/dart-lang/dartdoc/pull/4076

srawlins (approved):

PR #4244, opened 2026-05-19
Fixes crash when seeing a `this;` in a body.

lrhn:
Adds test that crashes without fix.


lrhn:
Found this when trying to add primary constructors to `dart:core`.

srawlins (approved):
Thanks!

PR #4245, opened 2026-05-20
Add baseline tests for Primary Constructors field induction and metadata propagation.

szakarias:
Adding tests verifying that `dartdoc` correctly consumes this information from the analyzer without requiring structural changes to the model.

**This PR Covers:**
* Field induction from `var`/`final` parameters.
* Bodyless declarations (`;` syntax) across all container types.
* Doc comment and annotation propagation to induced fields.
* Enum `const` constructor identification.

**Deferred to future PRs:**
* Comment references
* Parsing the `this` block body 
* Secondary constructor keywords (`new`/`factory`).

### DIFF
--- a/test/dartdoc_test_base.dart
+++ b/test/dartdoc_test_base.dart
@@ -49,7 +49,6 @@ abstract class DartdocTestBase {
 
   List<String> get experiments => [
         'enhanced-parts',
-        'wildcard-variables',
         'primary-constructors',
       ];
 
--- a/test/dartdoc_test_base.dart
+++ b/test/dartdoc_test_base.dart
@@ -47,7 +47,11 @@ abstract class DartdocTestBase {
   /// These are written out to a pubspec file using `path` keys.
   Map<String, String> get pubDependencyPaths => const {};
 
-  List<String> get experiments => ['enhanced-parts', 'wildcard-variables'];
+  List<String> get experiments => [
+        'enhanced-parts',
+        'wildcard-variables',
+        'primary-constructors',
+      ];
 
   bool get skipUnreachableSdkLibraries => true;
 
--- a/test/primary_constructor_test.dart
+++ b/test/primary_constructor_test.dart
@@ -26,7 +26,7 @@ class PrimaryConstructorsTest extends DartdocTestBase {
   void test_fieldInduction_class() async {
     var library = await bootPackageWithLibrary('''
 class C(var int x, final int y, int z);
-class Derived(super.x) extends C;
+class Derived(super.x, super.y, super.z) extends C;
 ''');
 
     var c = library.classes.named('C');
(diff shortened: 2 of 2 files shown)


test/primary_constructor_test.dart
```diff
@@ -0,0 +1,136 @@
...
+  String get libraryName => 'primary_constructors';
+
+  // ---------------------------------------------------------------------------
+  // Field Induction & Syntax
+  // ---------------------------------------------------------------------------
+
+  void test_fieldInduction_class() async {
+    var library = await bootPackageWithLibrary('''
+class C(var int x, final int y, int z);
+class Derived(super.x) extends C;
```
lrhn: Drive-by-comment: 
This code is invalid. The superclass constructor requires three positional arguments, `Derived` only passes one.

szakarias: fixed

test/primary_constructor_test.dart
```diff
@@ -0,0 +1,181 @@
...
+  void test_fieldInduction_class() async {
+    var library = await bootPackageWithLibrary('''
+class C(var int x, final int y, int z);
+class Derived(super.x, super.y, super.z) extends C;
+''');
+
+    var c = library.classes.named('C');
+    expect(c.instanceFields.any((f) => f.name == 'x'), isTrue);
+    expect(c.instanceFields.any((f) => f.name == 'y'), isTrue);
+    expect(c.instanceFields.any((f) => f.name == 'z'), isFalse);
```
sigurdm: Maybe slightly more elegant (if there are no more fields than those three)
```suggestion
    expect(c.instanceFields.map((f) => f.name), {'x', 'y', 'z'});
```

sigurdm (approved):

test/primary_constructor_test.dart
srawlins: Agreed, also I think `matcher`'s `equals` has support for Iterables having the same number of elements and the same elements.

test/dartdoc_test_base.dart
```diff
@@ -47,7 +47,11 @@ abstract class DartdocTestBase {
   /// These are written out to a pubspec file using `path` keys.
   Map<String, String> get pubDependencyPaths => const {};
 
-  List<String> get experiments => ['enhanced-parts', 'wildcard-variables'];
+  List<String> get experiments => [
+        'enhanced-parts',
+        'wildcard-variables',
```
bwilkerson: nit: You could probably remove `wildcard_variables` from the list as they were enabled by default in 3.7.

szakarias: done

test/primary_constructor_test.dart
szakarias: Done

PR #4247, opened 2026-05-27
Add tests for abbreviated constructor declarations

szakarias:
This PR adds tests for the new abbreviated constructor syntax, continuing our coverage for the Primary Constructors feature.

### DIFF
--- a/test/primary_constructor_test.dart
+++ b/test/primary_constructor_test.dart
@@ -295,4 +295,45 @@ class C(int x) {
     var result = referenceLookup(constructor, 'x').referable as ModelElement;
     expect(result.element, equals(xParam.element));
   }
+
+  // ---------------------------------------------------------------------------
+  // Abbreviated Constructors
+  // ---------------------------------------------------------------------------
+
+  void test_secondaryConstructors_new() async {
+    var library = await bootPackageWithLibrary('''
+class MyClass {
+  const new();
+  new name();
+  new redir(): this.name();
+}
+''');
+
+    var c = library.classes.named('MyClass');
+    expect(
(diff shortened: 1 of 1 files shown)


test/primary_constructor_test.dart
```diff
@@ -295,4 +295,45 @@ class C(int x) {
     var result = referenceLookup(constructor, 'x').referable as ModelElement;
     expect(result.element, equals(xParam.element));
   }
+
+  // ---------------------------------------------------------------------------
```
sigurdm, line 299: You know you can do
```
group('Abbreviated Constructors', () {
});
```
for grouping the tests?

test/primary_constructor_test.dart
```diff
@@ -295,4 +295,45 @@ class C(int x) {
...
+  const new();
+  new name();
+  new redir(): this.name();
+}
+''');
+
+    var c = library.classes.named('MyClass');
+    expect(
+      c.constructors.map((cons) => cons.name),
+      unorderedEquals([
```
sigurdm, line 315: Don't we actually expect these to come in declaration order? In that case we could skip the `unorderedEquals` part

sigurdm (approved):

test/primary_constructor_test.dart
szakarias: I am not sure we are guaranteed the order. I'll leave it for now.

test/primary_constructor_test.dart
szakarias: Yeahm I thought about it, but I'm using the `@reflectiveTest` pattern (like most of the other model element tests) and because of that `group()` can't be used around the individual methods.

PR #4251, opened 2026-06-03
Stop using modifiers on parameters.

lrhn:
The lines fail to compile when the code is included in the 3.13 SDK.

(Should the SDK version in pubspec.yaml be updated regularly?)


szakarias (approved):
LGTM

PR #4253, opened 2026-06-09
Support region extraction in {@example} directive

szakarias:


lib/src/model/documentation_comment.dart
```diff
@@ -339,14 +339,26 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
   ///
   /// The `path` is resolved relative to the current file's path, using the
   /// package's root as the root path;
   /// see [resolveExamplePath] for details.
   ///
   /// The file content is processed as follows:
   /// - Line endings are normalized to `\n`.
+  /// - If a `#<region>` is specified in the path, only the content within that
+  ///   region is extracted. A region is defined by `// #region <region>` and
+  ///   `// #endregion`. Region markers are stripped.
```
sigurdm: Does the region marker include the // ?

Should we just strip the entire line the region marker is on instead of stripping the region marker?

lib/src/model/documentation_comment.dart
```diff
@@ -339,14 +339,26 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+  /// - Lines containing a `#hide` marker are entirely omitted from the
+  ///   extracted output. This is useful for hiding setup, or assertions, or
+  ///   other code that is necessary for the example to compile but
+  ///   irrelevant to the documentation (e.g., `exit(0); // #hide`).
+  /// - Region extraction uses a best effort approach: if there are unmatched
+  ///   closing tags or unclosed regions, the code is extracted based on the
+  ///   stack ordering and emit a warning for the structural error.
+  /// - If no region is specified, the entire file is included and no markers
+  ///   are stripped.
```
sigurdm, line 361: Oh, so #hide doesn't work when no region is specified?
I had expected the opposite. But I guess it makes some sense...

lib/src/model/documentation_comment.dart
```diff
@@ -339,14 +339,26 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+  /// - If a `#<region>` is specified in the path, only the content within that
+  ///   region is extracted. A region is defined by `// #region <region>` and
+  ///   `// #endregion`. Region markers are stripped.
+  /// - Lines containing a `#hide` marker are entirely omitted from the
+  ///   extracted output. This is useful for hiding setup, or assertions, or
+  ///   other code that is necessary for the example to compile but
+  ///   irrelevant to the documentation (e.g., `exit(0); // #hide`).
+  /// - Region extraction uses a best effort approach: if there are unmatched
+  ///   closing tags or unclosed regions, the code is extracted based on the
+  ///   stack ordering and emit a warning for the structural error.
```
sigurdm, line 359: Nobody sees dartdoc warnings. Would it be bad to just fail hard?

lib/src/model/documentation_comment.dart
```diff
@@ -418,14 +438,40 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+          warn: warn,
+        );
+
+        if (extractedLines == null) {
+          warn(PackageWarning.missingExampleRegion,
+              message: '$filepath#$targetRegion');
+          return '';
+        }
+
+        lines = extractedLines;
```
sigurdm, line 457: nit suggestion
```suggestion
        lines = _extractRegion(
          lines,
          targetRegion,
          filepath: filepath,
          warn: warn,
        );

        if (lines == null) {
          warn(PackageWarning.missingExampleRegion,
              message: '$filepath#$targetRegion');
          return '';
        }
```

lib/src/model/documentation_comment.dart
```diff
@@ -418,14 +438,40 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+          lines,
+          targetRegion,
+          filepath: filepath,
+          warn: warn,
+        );
+
+        if (extractedLines == null) {
+          warn(PackageWarning.missingExampleRegion,
+              message: '$filepath#$targetRegion');
+          return '';
```
sigurdm, line 455: Should we return some text saying that the example failed to load?

lib/src/model/documentation_comment.dart
```diff
@@ -1146,4 +1192,96 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+  /// Returns `null` if the target region was never found.
+  static List<String>? _extractRegion(
+    List<String> lines,
+    String targetRegion, {
+    required String filepath,
+    required void Function(PackageWarning, {String? message}) warn,
+  }) {
+    // Defines standard comment markers.
+    // Uses a negative lookbehind `(?<!:)` on the `//` marker to prevent
+    // false positives in URLs (like http://) while still allowing
```
sigurdm: Are we sure we want to support a fixed set of comment types.

I think we either just detect `#region`/`#endregion`/`#hide` part, no matter where it occurs, or we require it to be prefixed with a dart single line comment somewhere on the line. Then other languages could embed that in their own comment style.

```sql
-- comment in sql // #region abc
select * from soap_operas;
-- comment in sql // #endregion abc
```

Would a url ever look like `http:// #hide`? Sounds overly cautious to me.

Also commentTokens is only ever used at the beginning of a regexp - so I guess you could get away with
`[^:]` instead of using negative lookahead...

lib/src/model/documentation_comment.dart
```diff
@@ -418,14 +438,40 @@ mixin DocumentationComment implements Warnable, SourceCode {
       // Normalize line endings to \n.
       content = content.replaceAll(_lineEndingRegExp, '\n');
 
+      var lines = content.split('\n');
+
+      // Extract the region and filter markers before touching the indentation
+      if (targetRegion != null && targetRegion.isNotEmpty) {
+        var extractedLines = _extractRegion(
```
sigurdm, line 445: I guess this is fine - but one could find all regions and cache them, so if there are several regions in one file it is only processed once...

lib/src/model/documentation_comment.dart
sigurdm: Did we not talk about repeating the name at the point of #endregion? such that nested/interleaved regions become a bit more clear. I might be misremembering,

sigurdm (approved):
LGTM, a few design questions, but I could live with the current design

lib/src/model/documentation_comment.dart
szakarias: The marker is inside a commented line. The entire line is stripped (as you can see in the tests). I will update the doc here to make it more clear, for instance: 'Lines containing region markers are stripped.'

lib/src/model/documentation_comment.dart
szakarias: I think it makes sense that it is the raw file when you include the entire file.

lib/src/model/documentation_comment.dart
szakarias: Crashing the entire documentation build for a mismatched region tag feels a bit too aggressive to me. It's also more consistent with our existing error handling to use a warning, just as we do when an author specifies an example file that doesn't exist.

lib/src/model/documentation_comment.dart
szakarias: `lines` is declared earlier and inferred as non-nullable, so to apply this suggestion, it would require explicitly typing `lines` as nullable. I think I have a vague preference for the current approach where `lines` remains guaranteed non-null, and we use the temporary `extractedLines` variable to safely test for null before reassigning.

lib/src/model/documentation_comment.dart
szakarias: I think it's better to stick with returning an empty string to remain consistent with the rest of the file. For example, if the user specifies an example file that doesn't exist, we currently emit a warning and return '' rather than injecting an error message into the generated documentation.

lib/src/model/documentation_comment.dart
szakarias: We do allow adding the name for clarity after endregion but we don't require it for parsing.

lib/src/model/documentation_comment.dart
szakarias: Good point! However, I'm not sure it's worth the added code complexity right now. I think this processing is generally quite fast, so I'd prefer to keep it simple and introduce caching later if it turns out to be a performance bottleneck.

lib/src/model/documentation_comment.dart
sigurdm: Ah makes sense

lib/src/model/documentation_comment.dart
szakarias: I agree with you on this one, thanks. Will update the impl to detect for #region without the comment.

lib/src/model/documentation_comment.dart
szakarias: Given that we change to allow region markers outside comments too, I will update this doc to reflect that.

lib/src/model/documentation_comment.dart
```diff
@@ -1209,28 +1216,22 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
-    // marker styles like `exit(0);// #hide`.
-    const commentTokens = r'(?:(?<!:)//|#|/\*|--|<!--)';
-
-    // Matches `#region <name>` anywhere in the line, preceded by a comment,
-    // and captures the region name.
+    // Matches `#region` followed by spaces/tabs (no newlines), and captures
+    // the name (any characters up to the next whitespace or newline).
     final regionStartPattern = RegExp(
-      commentTokens + r'.*#region\s+(\S+)',
+      r'#region[ \t]+([^\s]+)',
```
sigurdm: Should we also require a line ending?

lib/src/model/documentation_comment.dart
sigurdm: Do we validate it if it is present? Otherwise it might obscure more than it helps.

lib/src/model/documentation_comment.dart
sigurdm: That would make false positives less likely.

szakarias: This would not work in all languages for instance HTML

sigurdm: Hmm, good point!

Ok, then never mind.

PR #4259, opened 2026-06-16
Stop using `NullabilitySuffix.star`.

lrhn:
There is no such nullability suffix in Dart 3.0+ SDKs, since pre-2.12 language versions are not supported.

Don't refer to the constant, so that tools can eventually remove it.

The `nullabilitySuffixWithin` extension is now more complicated than it needs to be,
it can be reduced to `=> this.nullabilitySuffix == .question ? '?' : ''`, which isn't really a computation worth caching,
so I removed the extension and made the field a getter.

Small clean-ups in the file.

### DIFF
--- a/lib/src/element_type.dart
+++ b/lib/src/element_type.dart
@@ -16,8 +16,9 @@ import 'package:dartdoc/src/runtime_stats.dart';
 import 'package:dartdoc/src/type_utils.dart';
 import 'package:meta/meta.dart';
 
-/// Base class representing a type in Dartdoc.  It wraps a [DartType], and
-/// may link to a [ModelElement].
+/// Base class representing a type in Dartdoc.
+///
+/// It wraps a [DartType], and may link to a [ModelElement].
 abstract class ElementType with Referable {
   final DartType type;
   @override
@@ -25,10 +26,7 @@ abstract class ElementType with Referable {
   @override
   final Library? library;
 
-  final String nullabilitySuffix;
-
-  ElementType._(this.type, this.library, this.packageGraph)
-      : nullabilitySuffix = type.nullabilitySuffixWithin;
+  ElementType._(this.type, this.library, this.packageGraph);
 
   factory ElementType.for_(
       DartType type, Library? library, PackageGraph packageGraph) {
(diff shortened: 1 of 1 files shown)


lib/src/element_type.dart
```diff
@@ -107,7 +111,7 @@ class UndefinedElementType extends ElementType {
     // We can not simply throw here because not all SDK libraries resolve
     // all types.
     if (type is InvalidType) return 'dynamic';
-    assert(const {'Never'}.contains(type.documentableElement?.name),
+    assert(type.documentableElement?.name == 'Never',
         'Unrecognized type for UndefinedElementType: $type');
     return type.documentableElement!.name!;
```
lrhn, line 116: Could just be `return 'Never';` if the assertion is correct.

lib/src/element_type.dart
```diff
@@ -361,19 +365,3 @@ mixin Rendered implements ElementType {
...
   ElementTypeRenderer<ElementType> get _renderer;
 }
-
-extension on DartType {
-  /// The dartdoc nullability suffix for this type in [library].
-  String get nullabilitySuffixWithin {
-    if (this is! VoidType && !isBottom) {
-      /// If a legacy type appears inside the public interface of a Null
-      /// safety library, we pretend it is nullable for the purpose of
-      /// documentation (since star-types are not supposed to be public).
```
lrhn, line 371: No longer refers to library, so documentation is outdated.

The `nullabilitySuffix` cannot be `.star`, and for `VoidType` or a bottom type, the `nullabilitySuffix` cannot be `.question`, so this can be reduced to `nullabilitySuffix == .question ? '?' : ''`.

lib/src/element_type.dart
```diff
@@ -61,10 +63,12 @@ abstract class ElementType with Referable {
   String get nameWithGenericsPlain;
 
   @override
-  String get displayName => throw UnimplementedError();
```
lrhn, line 64: (I'd usually use `UnsupportedError` if the method is never intended to be implemented. I can see the code uses `UnimplementedError` in a number of places where it makes sense, for example if another language constructor is added and there is nothing handling that.
Then it's not intended to not be supported, it was just not known at writing time that there'd be something to implement.
I don't know if this place matches that usage.)

szakarias (approved):

PR #4262, opened 2026-06-18
Require analyzer 14.1.0, migrate from deprecated updateAnalysisOptions4.

scheglov:
Using
```yaml
dependency_overrides:
  analyzer: 14.0.0
```
because package test does not support this version of analyzer yet (but is not broken, [updating](https://github.com/dart-lang/test/pull/2676) it),  and I'd like to roll dartdoc into Dart SDK, and remove these deprecated non-APIs.


szakarias (commented):
Is it intentional that you updated the dependency_overrides instead of just a regular dependency update?

natebosch (approved):

Issue #4243, opened 2026-05-11
Add CSS rules to style custom example containers


dkwingsmt:
## Background

The Flutter team is decoupling the material and cupertino libraries into independent packages. As part of this transition, their API documentation will use standard pub.dev generation processes. This decoupling work is scheduled to complete around June 2026.

While we want to use standard generation processes where possible, some visual elements are necessary to maintain the usability of the documentation.

## The need for a styled example container

Flutter API documentation has historically enclosed code examples inside a blue container. We evaluated this design and concluded that this container is important for readability. This makes the container a blocking requirement for our decoupling plans.

We explored alternative implementations, such as using [blockquote highlights](https://github.com/orgs/community/discussions/16925). However, the visual distinction is insufficient, and the `@example` directive does not parse correctly inside a blockquote.

To resolve this, we propose adding support for a custom `<example>` HTML tag.

## Proposed implementation

Because markdown supports raw HTML tags, this approach does not require complex changes to the dartdoc parser. It only requires adding CSS rules to dartdoc to style the `<example>` tag as a styled container.

We understand that adding styles specifically for Flutter can increase the maintenance burden of dartdoc. We propose this CSS-only approach because it is self-contained and minimizes changes to the core codebase.

Below is the CSS we have tested and verified:

```css
  example {
    padding: 10px;
    overflow: auto;
    display: block;
    unicode-bidi: isolate;
    margin: 1em 0;
  }

  example p {
    margin: 8px 40px 12px 8px;
  }

  example pre {
    margin: 0;
  }

  example code.hljs {
    max-height: 500px;
  }

  .light-theme example {
    background-color: rgb(215, 235, 252);
  }

  .dark-theme example {
    background-color: rgb(30, 40, 51);
  }
```

Below is how API doc will look like in practice:

```dart
///
/// <example>
///
/// This example shows a toggleable [CupertinoSwitch]. When the thumb slides to
/// the other side of the track, the switch is toggled between on/off.
///
/// {@example /examples/api/switch/cupertino_switch.0.dart}
///
/// </example>
///
```

<img width="461" alt="Image" src="https://github.com/user-attachments/assets/5f2a3aa6-0d42-4501-8dc9-9258ae9957d7" />

We would appreciate your feedback on this approach.

szakarias:
Fixed by https://github.com/dart-lang/dartdoc/pull/4264

PR #4264, opened 2026-07-01
Add support and styling for custom <callout-box> HTML tags

szakarias:
image


web/styles/_content.scss
```diff
@@ -142,3 +142,23 @@ th,
...
+example p {
+  margin: 8px 40px 12px 8px;
+}
+
+example pre {
+  margin: 0;
+}
+
+example code.hljs {
+  max-height: 500px;
```
jonasfj, line 164: `overflow-y: scroll` or maybe `auto` is better to not always have a scroll bar.

jonasfj (approved):

web/styles/_content.scss
jonasfj: This is probably not necessary, really... because auto is probably the default behavior, try it out -- if it works then everything is awesome :D

lib/src/render/documentation_renderer.dart
```diff
@@ -206,6 +206,7 @@ final _allowedElements = <String>{
   'SPAN',
   'TIME',
   'WBR',
+  'EXAMPLE',
```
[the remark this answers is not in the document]
jonasfj: Hmm, this is potentailly a good argument. `<example-block>` or `<example-section>` or just `<dart-example>` or `<doc-example>`.

Using `<dart-` for custom HTML tags might not be bad idea. There is a risk this isn't the last time we'll add something like this..

We could also do `<flutter-example>`, since it really is flutter style for flutter documentation.

sigurdm (approved):
Would be cool with a screenshot in the PR description.

szakarias:
Fixes https://github.com/dart-lang/dartdoc/issues/4243

PR #4267, opened 2026-07-03
Add interactive copy-to-clipboard button to code snippets

szakarias:
image

image

image

image


sigurdm (approved):

web/copy_button.dart
```diff
@@ -0,0 +1,66 @@
...
+
+import 'dart:js_interop';
+
+import 'package:web/web.dart';
+
+const _copyIconHtml = '''
+  <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round">
+    <rect x="9" y="9" width="13" height="13" rx="2" ry="2"></rect>
+    <path d="M5 15H4a2 2 0 0 1-2-2V4a2 2 0 0 1 2-2h9a2 2 0 0 1 2 2v1"></path>
+  </svg>
```
jonasfj: I don't dislike this, but we do currently load Material Symbols:
https://github.com/szakarias/dartdoc/blob/283c22da4351b4b945b94f64dc59f8661689a4d5/lib/templates/_head.html#L23

Isn't there appropriate icons from there we can just use?

web/copy_button.dart
```diff
@@ -0,0 +1,66 @@
...
+            window.navigator.clipboard.writeText(text);
+          } catch (_) {
+            return;
+          }
+
+          button.innerHTML = _checkIconHtml.toJS;
+          button.classList.add('copied');
+
+          if (timeoutId != null) {
+            window.clearTimeout(timeoutId!);
```
jonasfj, line 37: Why clear old timeout? if I click copy quickly between two differnet boxes one of them won't clear up 🤣

web/copy_button.dart
```diff
@@ -0,0 +1,66 @@
...
+''';
+
+const _checkIconHtml = '''
+  <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round">
+    <polyline points="20 6 9 17 4 12"></polyline>
+  </svg>
+''';
+
+void init() {
+  var blocks = document.querySelectorAll('pre');
```
jonasfj: consider `document.querySelectorAll('pre:has(> code)')`

or `document.querySelectorAll('pre:has(> code:first-child)')` to get places where code is the first child of pre

web/copy_button.dart
jonasfj: I didn't test this, but I think this could be just:


```dart
void _copyCodeToClipboard(Event e) async {
  var button = e.currentTarget as HTMLElement?;
  var code = button?.parentElement?.querySelector('code');
  var text = code?.textContent ?? '';
  window.navigator.clipboard.writeText(text);

  if (button != null) {
    button
      ..querySelector('span')?.textContent = 'download_done'
      ..classList.add('copied');
    await Future<void>.delayed(const Duration(seconds: 2));
    button
      ..querySelector('span')?.textContent = 'content_copy'
      ..classList.remove('copied');
  }
}

void init() {
  var blocks = document.querySelectorAll('pre:has(>code)');
  for (var i = 0; i < blocks.length; i++) {
    var pre = blocks.item(i) as HTMLElement?;
    if (pre == null) {
      continue;
    }
    pre.prepend(
      document.createElement('button')
        ..appendChild(document.createElement('span')
          ..textContent = 'content_copy'
          ..classList.add('material-symbols-outlined'))
        ..innerHTML = _copyIconHtml.toJS
        ..className = 'copy-button'
        ..addEventListener('click', _copyCodeToClipboard.toJS),
    );
  }
}
```

web/copy_button.dart
```diff
@@ -0,0 +1,66 @@
...
+          var codeBlock = pre.querySelector('code') as HTMLElement?;
+          var text = codeBlock?.textContent ?? '';
+          
+          try {
+            window.navigator.clipboard.writeText(text);
+          } catch (_) {
+            return;
+          }
+
+          button.innerHTML = _checkIconHtml.toJS;
```
jonasfj: avoid `innerHTML` we should forbid it, while yes it's safe when we only write known strings to it.

In a ideal world we should just forbid writing to it, it's too easy to make XSS with it.

web/copy_button.dart
```diff
@@ -0,0 +1,66 @@
...
+    button.addEventListener(
+        'click',
+        (Event _) {
+          var codeBlock = pre.querySelector('code') as HTMLElement?;
+          var text = codeBlock?.textContent ?? '';
+          
+          try {
+            window.navigator.clipboard.writeText(text);
+          } catch (_) {
+            return;
```
jonasfj: Why not let it bubble into the console? so users can see it in the debug console if they look... this just disappears we'll never find the bug?

It won't popup anywhere users can see it, unless they open the developer console in chrome

web/copy_button.dart
szakarias: "Thanks! I updated the code to use the Material font, `span.textContent`, and the `pre:has(>code)`. I kept the old `timeoutId`, though, to avoid multi-click race conditions.

web/copy_button.dart
szakarias: done.

web/copy_button.dart
szakarias: done.

web/copy_button.dart
szakarias: done

web/copy_button.dart
szakarias: `timeoutId` is declared inside the for loop, so the timers are completely isolated. Clicking quickly between two different boxes works perfectly. I need to clear it to handle when the user clicks the same box multiple times quickly, so the older timers don't wake up and reset the icon prematurely.

jonasfj (approved):

PR #4268, opened 2026-07-03
Fix RangeError when stripping doc-imports from files with CRLF line endings

sigurdm:
Normalizing source ranges for doc-imports previously computed offsets based on raw file offsets without accounting for CRLF line ending normalization in element.documentationComment. As a result, files with CRLF line endings caused index drift and RangeError during doc-import stripping.

This change updates comment offset normalization to correctly convert raw file offsets to content offsets, taking into account line ending normalization and gaps between comment tokens.

### DIFF
--- a/lib/src/model/model_node.dart
+++ b/lib/src/model/model_node.dart
@@ -207,19 +207,35 @@ extension on Comment {
   /// A mapping of all comment references to their various data.
   CommentData get data {
     var tokenList = tokens.toList();
+    var tokenContentStarts = List<int>.filled(tokenList.length, 0);
+    var normalizedLengths = List<int>.filled(tokenList.length, 0);
+    var currentContentOffset = 0;
+    for (var i = 0; i < tokenList.length; i++) {
+      tokenContentStarts[i] = currentContentOffset;
+      var token = tokenList[i];
+      var hasTrailingNewline =
+          i + 1 < tokenList.length && tokenList[i + 1].offset > token.end;
+      var normalizedLexemeLength = token.lexeme.replaceAll('\r', '').length;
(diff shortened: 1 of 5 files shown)


jonasfj:
Seems like this is a known issue, perhaps we should ask if fixing it this way is desirable?

Maybe this mapping of ranges should be done when we extract the comment contents?

Just off the top of my head,

szakarias (approved):

szakarias:
Thanks for the quick fix! The long-term plan is to move doc comment processing into the analyzer but let's solve the bleeding  and merge this now to unblock code_assets and the rest of the ecosystem.

lib/src/model/model_node.dart
```diff
@@ -205,26 +206,38 @@ final String _escapedStarSlash = _escape.convert(_starSlash);
...
+    var tokenList = tokens.toList();
+    var tokenContentStarts = List<int>.filled(tokenList.length, 0);
+    var normalizedLengths = List<int>.filled(tokenList.length, 0);
+    var currentContentOffset = 0;
+    for (var i = 0; i < tokenList.length; i++) {
+      tokenContentStarts[i] = currentContentOffset;
+      var token = tokenList[i];
+      var hasTrailingNewline =
+          i + 1 < tokenList.length && tokenList[i + 1].offset > token.end;
+      var normalizedLexemeLength = token.lexeme.replaceAll('\r', '').length;
```
lrhn, line 218: (Totally pedantic: If this replaced `\r\n` with `\n`, it would not affect any (very hypothetical) text that uses plain `\r` as line terminators. Which Unicode allows, even if nobody will ever do that again. Right?!)

PR #4285, opened 2026-09-18
Adapt the {@example} code fence to the example content

szakarias:
The `{@example}` directive wrapped the example in a fence of exactly three backticks, so an example which itself contains a fenced code block closed the block early. 

The fence is now one backtick longer than the longest run beginning a line of the example, so the code block encloses the whole example. Examples with no such run are unaffected: the fence stays at three backticks.

Also rejects a backtick in `lang=`, which would otherwise make the opening fence invalid.


lib/src/model/documentation_comment.dart
```diff
@@ -626,6 +641,24 @@ mixin DocumentationComment implements Warnable, SourceCode {
...
+  ///
+  /// A code block ends at a line beginning with as many backticks as opened it,
+  /// so the fence is one backtick longer than the longest run beginning a line
+  /// of [content], and at least three.
+  static String _codeFenceFor(String content) {
+    var longestRun = 0;
+    for (var line in LineSplitter.split(content)) {
+      var trimmed = line.trimLeft();
+      var run = 0;
+      while (run < trimmed.length && trimmed.codeUnitAt(run) == 0x60) {
```
sigurdm, line 654:
```suggestion
      while (run < trimmed.length && trimmed.codeUnitAt(run) == 0x60) { // 0x60 = `.
```

sigurdm (approved):

lib/src/model/documentation_comment.dart
szakarias: done

